PR #2630, opened 2022-01-19
Display Settings from BrowserView

jed-fox-neeva:
DEPENDS ON #2635

~NOTE: This does not support viewing settings from non-`BrowserView` builds. We should make `BrowserView` the default before shipping this.~

### Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines


jedireza:
Closing due to freshness.

PR #2796, opened 2022-02-04
Add Height Preference for Middle Position in OverlaySheet

EdwardJTL:
## Description
- Add new preference key for middle position
- Update how top space height is calculated to support optional middle position preference
- Added new overlay style to support sheet with no `ScrollView` wrapper and fixed positions
- Apply `topDrag` with `highPriorityGesture` to resolve conflict with the buttons in the overflow menu
- Updated the use of `ScrollView` with `OverflowMenu` to work with the new UX

## Note
- After this change, i noticed that it is a bit slower to open the sheet overlays, esp on the first time. This happened after I started reading additional values when calculating the top spacer height in the middle position. I am not sure how to resolve this.

and thanks @evancrow for giving me advice a couple times with this change :) 

### Security


EdwardJTL:
> Performance seems ok to me, but maybe I'm not testing the right thing.
> 
> A few things I did notice however:
> 
> 1- The bottom padding appears to be gray (the border color) and is cutting off the content. The content should appear to full-bleed into the bottom safe area.
> 
> 2- It looks like there is a scrollview. Ideally we should not need a scrollview for this menu unless the content does not fit on the screen (talking portrait mode of course).
> 
> 3- According to [the design mock](https://www.figma.com/file/oMHQ4MKMCFhtidxPptElXZ/iOS-Design-System-(Mobile-Browser)?node-id=6194%3A78420), there shouldn't be a handle bar at the top.

1. ✅Did this with another preference key to lay out the child as if they can ignore the bottom safe edge
2. ⚠️overlaysheet comes with a scrollview that gets enabled when the sheet is in the top position. Maybe we could introduce an additional style to the sheet that can be dragged up or down but does not wrap content with  scrollview?
3. ✅ did this with another preference key to hide the top handle. although now i wonder if user will know how to drag it back down...

darinwf:
> 2. ⚠️overlaysheet comes with a scrollview that gets enabled when the sheet is in the top position. Maybe we could introduce an additional style to the sheet that can be dragged up or down but does not wrap content with  scrollview?

That sounds like a reasonable approach. I think disabling this ScrollView is important so the user can drag down the menu just as easily as they dragged it up. Without this change, the ScrollView will get in the way of that and users would then really need the handle bar for a larger target to drag from. (Or course users can also tap on the backdrop to dismiss, but they might not realize that.)

darinwf:
The other goal should be to have the full-height be just what is needed to completely contain the overflow menu. Currently it opens all the way leaving a gray area at the bottom. I think the ideal setup here is for the menu to open to where it would be if we were just using the old fixed height approach. Does that make sense?

darinwf:
Also, if you want to land this in stages that's okay too.

darinwf:
This is looking really great! Only issue I'm seeing now is that the height seems to vary and is always more than necessary, video:

https://user-images.githubusercontent.com/79867437/153691698-8951e2bd-604c-4b88-bc7c-8f1943de600c.MP4

EdwardJTL:
> This is looking really great! Only issue I'm seeing now is that the height seems to vary and is always more than necessary, video:
> 
>  RPReplay_Final1644630627.MP4

the mistake was that i wrapped the `sheetContent` in a `VStack{ Spacer() }` to fill the screen when its dragged too high. but the `contentHeight` was read from that VStack, which just kept expanding and allowed the height to continue to increase. I shuffled around where view modifiers are applied, which fixed this issue; i tried a couple of other places where sheet is used and they still look fine.

darinwf:
Test drove and it works great! Nice job

darinwf (approved):

PR #2827, opened 2022-02-08
Remove slack script from build

chung-lai:
### Security


darinwf (commented):
Is the idea that we will just need to define NEEVA_REPO? Perhaps this script should detect if that variable is not defined and warn that it should be created? I think we can also just make this error out if not set and if `send-slack-message` cannot be found since we want these slack messages.

chung-lai:
> Is the idea that we will just need to define NEEVA_REPO? Perhaps this script should detect if that variable is not defined and warn that it should be created? I think we can also just make this error out if not set and if `send-slack-message` cannot be found since we want these slack messages.

Ah, thought `NEEVA_REPO` is always set after setting up the server, i think only some people have set it :D Ok, will trigger a warning. Since we are using `build_script.sh` for pushing build these days, I will do the check in the build script as well.

chung-lai:
Just want to get this in to unblock the slack integration. Will have a follow up PR to refactor the check into another script so we don't have to repeat the same code everywhere :D

darinwf (approved):

PR #2830, opened 2022-02-09
Use Resolved Query from Fast Tap in Cheatsheet

EdwardJTL:
## Description
We want to default to using the resolved query when available in Cheatsheet search. No UI changes.

- Add additional string to QueryForNavigation via new Struct
  - propagate change to call stack
- Add additional string to SessionData
  - Propagate change to create and restore tab methods

### Security


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -805,7 +811,7 @@ class BrowserViewController: UIViewController, ModalPresenter {
                     self.recordNavigationInTab(tab, navigation: nav, visitType: visitType)
                 }
             } else if let nav = tab.loadRequest(URLRequest(url: url)) {
-                tab.queryForNavigation.currentSearchQuery = searchQueryModel.value
+                tab.queryForNavigation.currentQuery = .init(searchQuery: searchQueryModel.value, suggestedQuery: suggestedQuery)
```
darinwf, line 814: nit: It might be helpful to change `searchQuery` to a name that makes it more clear that this is the user typed value or prefix query. Maybe `typedQuery` would work.

It could also be reasonable to leave off the `Query` suffixes and just go with `typed:` and `suggested:` since it is the context of a struct named `Query`.

Client/Frontend/Browser/ContentScripts/SessionRestoreHelper.swift
```diff
@@ -32,11 +32,16 @@ class SessionRestoreHelper: TabContentScript {
 
                     if let navigationList = tab.webView?.backForwardList.all {
                         for (index, item) in navigationList.enumerated() {
-                            guard tab.sessionData?.queries.indices.contains(index) ?? false,
-                                let query = tab.sessionData?.queries[index]
+                            guard let sessionData = tab.sessionData,
+                                sessionData.queries.indices.contains(index),
```
darinwf, line 36: nit: maybe here we can also use a name like `typedQueries` instead of just `queries`.

Client/Frontend/Browser/ContentScripts/SessionRestoreHelper.swift
```diff
@@ -32,11 +32,16 @@ class SessionRestoreHelper: TabContentScript {
...
+                                let query = sessionData.queries[index]
                             else { break }
 
-                            tab.queryForNavigation.queryForNavigations[item] = query
+                            var suggestedQuery: String? = nil
+                            if sessionData.suggestedQueries.indices.contains(index) {
+                                suggestedQuery = tab.sessionData?.suggestedQueries[index]
```
darinwf, line 42: nit: can this be `sessionData.suggestedQueries` as well?

Client/Frontend/Browser/SessionData.swift
```diff
@@ -52,13 +52,15 @@ class SessionData: NSObject, NSCoding {
...
     /// If a query is nil, there is not a query for that specific navigation.
     let queries: [String?]
+    let suggestedQueries: [String?]
     let lastUsedTime: Timestamp
 
     var jsonDictionary: [String: Any] {
         return [
             "currentPage": String(self.currentPage),
             "urls": urls.map { $0.absoluteString },
             "queries": queries,
```
darinwf, line 62: If we go with renaming `queries` to `typedQueries`, we'd need to preserve using the json key `"queries"` to ensure compat with existing saved tabs.

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -312,7 +312,7 @@ class Tab: NSObject, ObservableObject {
                 {
```
darinwf, line 312: Just a note for the future, we should really move this cheatsheet specific code out of the `Tab` class. It doesn't belong here :-)

darinwf (approved):

darinwf:
LGTM w/ nits. Please fix and then good to submit. Thanks!

Client/Frontend/Browser/ContentScripts/SessionRestoreHelper.swift
EdwardJTL: good catch :D

darinwf (approved):
LGTM!

PR #2737, opened 2022-01-28
BrowserHost IntroView

evancrow:
Fixed: #2728
Fixed: #2725

cc @macy-n @chung-lai  

### Security
- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

PR #2832, opened 2022-02-09
Disable flaky test `testSignInWithStaleLoginCookie()`

chassu:
### Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines


evancrow:
I think my recent PR #2737 might have fixed this test actually

darinwf (changes requested):
We should use the `skipTest` method of disabling a test. That way the test is still compiled. Also, sounds like Evan thinks we may not need to disable this test?

evancrow:
Maybe try leaving the test unskipped in this PR to see if it works...

chassu:
I'm seeing this test getting flaky on main again... adding `skiptest()`

darinwf (approved):

PR #2836, opened 2022-02-09
inline tab group flip flag

chassu:
### Security

- Flip the flag and remove old tab group related code


Client/Frontend/Browser/Card/Card.swift
```diff
@@ -245,24 +241,12 @@ struct Card<Details>: View where Details: CardDetails {
 
                 if !details.thumbnailDrawsHeader {
                     HStack(spacing: 0) {
-                        if !FeatureFlag[.tabGroupsNewDesign] && isChildTab(details: details)
-                            && (cardTransitionModel.state == .visibleForTrayShow)
```
darinwf, line 249: I think this means we get to remove the `@EnvironmentObject var cardTransitionModel: CardTransitionModel` line above :-D

Client/Frontend/Browser/Card/Card.swift
```diff
@@ -207,10 +207,6 @@ struct Card<Details>: View where Details: CardDetails {
         GeometryReader { geom in
             VStack(alignment: .center, spacing: 0) {
                 let button = Button(action: {
-                    if let tabGroupCardDetail = tabGroupCardDetail {
-                        tabGroupCardModel.detailedTabGroup = tabGroupCardDetail
```
darinwf, line 211: Does this mean you can get rid of the `@EnvironmentObject var tabGroupCardModel: TabGroupCardModel` above?

darinwf (approved):
LGTM, please look to see if there are any other `@EnvironmentObject` or `@ObservedObject` lines that are no longer needed. These are free performance wins!

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -48,20 +48,17 @@ class TabCardModel: CardModel, TabEventHandler {
...
-        register(self, forTabEvents: .didClose, .didChangeURL)
-        onDataUpdated()
-        self.anyCancellable = manager.objectWillChange.sink { [weak self] (_) in
+        // Process updates to the TabManager state asynchronously, avoiding duplicates. This
+        // way we wait until after `tabs` has been updated before we do any work.
+        self.subscription = manager.$tabs.receive(on: DispatchQueue.main).removeDuplicates().sink {
+            [weak self] (_) in
             if manager.didRestoreAllTabs {
                 self?.onDataUpdated()
                 self?.objectWillChange.send()
```
evancrow, line 58: If this is listening for changes, why do you have to call objectWillChange?

evancrow (approved):
@darinwf Your changes LGTM, with one question

Client/Frontend/Browser/Card/CardModels.swift
darinwf: I think that is to signal downstream observers of the TabCardModel... but come to think of it, all of the properties of TabCardModel are `@Published`, so this is probably an unnecessary `objectWillChange.send()`. Hmm...

darinwf: Yes, that `objectWillChange.send()` should be totally redundant with all of the `@Published` properties that get modified by `onDataUpdated`. No need to keep it around.

chassu:
Text fixes LGTM.

PR #2841, opened 2022-02-10
Fix issue with remove from group not updating the card grid

### LINKED ISSUE #2840: Inline tab groups "Remove from group" feature appears to delete the tab...
The tab is removed from the group immediately, but then will show up some time later next to the tab group. I haven't been able to figure out the pattern.

https://user-images.githubusercontent.com/79867437/153367231-d4e2d349-0caa-494e-8cec-c4f460e2596c.MP4

@chassu FYI... i wonder if some of my changes to suppress updating caused this. we can investigate tomorrow...

darinwf:
Fixes #2840


darinwf:
I was being too aggressive by only listening to `$tabs`. That's where we want to get to, but it'll take more steps.

evancrow (approved):
LGTM

darinwf:
Just for posterity, there were two issues:

1- Observing `tabs` was insufficient to pickup all changes to `TabManager` that should result in `onDataUpdated`, so watching for `objectWillChange` instead was necessary.

2- The check for `didRestoreAllTabs` is important to avoid calling `onDataUpdated` 100s of times at app startup, once per restored tab. That check needs to happen before the `receive(on: DispatchQueue.main)` otherwise it wouldn't actually serve its purpose.

PR #2845, opened 2022-02-10
Remove unused environment object in card.swift

chassu:
@darinwf You've also mentioned removing `@EnvironmentObject var cardTransitionModel: CardTransitionModel`, but I'm seeing it being used in:

```
        .accesibilityFocus(
            shouldFocus: details.isSelected, trigger: cardTransitionModel.state == .hidden
        )
```
Do we still get to remove it? Or we should keep it?


darinwf:
> @darinwf You've also mentioned removing `@EnvironmentObject var cardTransitionModel: CardTransitionModel`, but I'm seeing it being used in:
> 
> ```
>         .accesibilityFocus(
>             shouldFocus: details.isSelected, trigger: cardTransitionModel.state == .hidden
>         )
> ```
> 
> Do we still get to remove it? Or we should keep it?

Oh looks like it is still needed :-/

darinwf (approved):

PR #2846, opened 2022-02-10
Add flow common ios source files to project

EdwardJTL:
## Description
Added source files from FlowCommoniOS to the project. https://github.com/createwithflow/FlowCommoniOS
Since we don't use cocoapods, the files were added directly.

## Notes
I made a readme for the copyright notice.


FlowCommon/README.md
```diff
@@ -0,0 +1,25 @@
+ Code Adapted from FlowCommon:
```
darinwf, line 1: nit: Please move into the ThirdParty/ top-level directory, thanks!

FlowCommon/README.md
```diff
@@ -0,0 +1,25 @@
...
+ 
+ Copyright © 2016-2019 JABT Labs Inc.
```
darinwf, line 3: Maybe this should be a LICENSE file instead? That would probably be a bit more of a canonical way of handling this.

darinwf: See for example ThirdParty/Deferred/LICENSE

ThirdParty/FlowCommon/CAKeyframeAnimation+Extension.swift
```diff
@@ -0,0 +1,15 @@
+// Code Adapted from FlowCommoniOS. See README.md for license
```
darinwf, line 1: These should all just say `See LICENSE for license` I guess. Or, alternatively... why not preserve the original license headers for these files?

I know it is more work for you upfront, but perhaps you could try doing one PR that just lands the code unmodified from the upstream source, and then do a subsequent PR that applies your changes? That way in our revision history we can see the edits we have made on top of the original source. This will be helpful if we want to rebase the patch onto a newer version of the third-party library, etc. It also just really helps document what we changed.

EdwardJTL:
redo-ing this one

EdwardJTL:
merge #2851 first

darinwf (approved):

PR #2847, opened 2022-02-10
add loggings for inline tab group

chassu:
Still WIP. I'm thinking about what kind of data the logging can generate. Open to any suggestions.


darinwf (approved):
LGTM. Would be great to get a review from Macy / Chung who know the logging system better.

chung-lai (approved):
LGTM, thanks for adding the logging!

Client/Application/SceneDelegate.swift
```diff
@@ -100,9 +100,18 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
         bvc.downloadQueue.resumeAll()
 
+        var attributes = EnvironmentHelper.shared.getAttributes()
```
chung-lai, line 103: I think we had a conversation before about whether `normalTabsOpened` should include inline tab groups' tabs. Should we have `normalTabsOpened`, `numTabGroupsTotal` and `numofTabGroupTabs`? That way, we can see some distribution of # of tab groups' tabs.

Or we already log that when we open tab switcher?

chassu:
Some new added loggings are depending on https://github.com/neevaco/neeva-ios/pull/2891, will wait for that to be merged then I'll merge this.

PR #2848, opened 2022-02-10
Add new NeevaScope loading animation

EdwardJTL:
## Description
Add new loading screen with the animation.
The animation looks mostly correct and centered.

The path code and animation code was generated from Flow. #2846 adds the flow source code. The diff will look cleaner once that one is merged.


Client/Frontend/Overlay/Content/Cheatsheet/NeevaScopeLoading/NeevaScopeLoadingKeyFrameView.swift
```diff
@@ -0,0 +1,224 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
```
darinwf, line 1: You mentioned over Slack that these files were generated using Flow export. Could you include a README.md file in this directory that explains the steps needed to recreate these files? Just want to make sure that we enable other people to also repeat your steps if needed. Thanks!

PR #2849, opened 2022-02-11
Digest Polish

evancrow:
- Remove Space names from description
- Limit digest to two items per space
- Tap Space name to open the Space
- Add see more button


chung-lai:
Thanks! Does this PR include change to these?

* Show total count? e.g. “Daily Digest (4)”
* Drop “Your” on both grid and list, and title casing “Daily Digest"

@hwiatneeva, also wonder if we want different bg color for "see more" link

hwiatneeva:
> Thanks! Does this PR include change to these?
> 
> * Show total count? e.g. “Daily Digest (4)”
> * Drop “Your” on both grid and list, and title casing “Daily Digest"
> 
> @hwiatneeva, also wonder if we want different bg color for "see more" link

Questions & some suggestions re: see more @chung-lai 
1. Is hyper-linking on the Space name an option? (I forgot the reasoning for the separate 'See More')
2. Another idea:  [Space name]      [Go to Space]  might be in the same row but separate invisible columns, and make 'Go to Space' as hyperlink, and right-aligned? 
3. If the see more in a separate row is the only feasible option, make the back ground the same as the item background, so it's subtle, and blended in to its Space section (vs. separate row that would be ambiguous of where it belongs when there's another Space title is shown below).
4. Make the label 'Title Casing' probably for iOS convention? And try 'Go to Space' to be a bit more explicit than 'See More', which seems that it would expand items inline. 

Cheers.

chung-lai:
> > Thanks! Does this PR include change to these?
> > 
> > * Show total count? e.g. “Daily Digest (4)”
> > * Drop “Your” on both grid and list, and title casing “Daily Digest"
> > 
> > @hwiatneeva, also wonder if we want different bg color for "see more" link
> 
> Questions & some suggestions re: see more @chung-lai
> 
> 1. Is hyper-linking on the Space name an option? (I forgot the reasoning for the separate 'See More')
> 2. Another idea:  [Space name]      [Go to Space]  might be in the same row but separate invisible columns, and make 'Go to Space' as hyperlink, and right-aligned?
> 3. If the see more in a separate row is the only feasible option, make the back ground the same as the item background, so it's subtle, and blended in to its Space section (vs. separate row that would be ambiguous of where it belongs when there's another Space title is shown below).
> 4. Make the label 'Title Casing' probably for iOS convention? And try 'Go to Space' to be a bit more explicit than 'See More', which seems that it would expand items inline.
> 
> Cheers.

Thanks for the comment! I remember @yozuysal mentioned it might be tricky to add a hyper-linking option to the space name section. Yusuf, could you confirm if that's the case or I made the wrong assumption :(

evancrow:
@chung-lai I created a new space entity item that we can use for hyperlinking titles! So currently if you tap the the space name in the digest it opens the space.

chung-lai:
> @chung-lai I created a new space entity item that we can use for hyperlinking titles! So currently if you tap the the space name in the digest it opens the space.

Ah, i guess I misunderstood then. Awesome, as long as the space title will take user to the space, we don't need the "See More" or "go to space" link at the end as @hwiatneeva suggested. Thanks!

evancrow:
oh okay, I'll take it out! Should the title color change though? To a link blue or something like that?

chung-lai:
> oh okay, I'll take it out! Should the title color change though? To a link blue or something like that?

Link blue sounds good to me but will consult @hwiatneeva :)

chung-lai (approved):
LGTM, thanks!

evancrow:
Here's where it's at now:

Simulator Screen Shot - iPhone 13 Pro - 2022-02-14 at 19 42 46

chung-lai:
That looks good to me! we can probably land this and we can always do any small polish before the release. @hwiatneeva, please keep the feedback coming and we can prioritize that as it comes.

PR #2852, opened 2022-02-11
Keyboard SafeArea Fixes

evancrow:
Fix some cases where the keyboard would hide content. Be careful when using `ignoreSafeArea`!


darinwf (changes requested):
Actually, I tested this and now I'm seeing the bottom bar show up above the keyboard on linkedin.com:

Screen Shot 2022-02-10 at 9 33 24 PM

darinwf:
It is not consistent.. the bottom bar doesn't always appear above the keyboard. really strange.

evancrow:
@darinwf That was caused by switching form one textfield to another, causing the view to update when it was not necessary and somehow that caused the keyboard height to get set to 0.

darinwf (approved):
Tested it out on linkedin.com, and it works great. Thanks!

PR #2853, opened 2022-02-11
Display gasPrices both in gwei and in dollars

yozuysal:
- Gets an estimate for the gas amount
- In the spirit of small models and Views, moves all transaction fee logic out of Web3Model
- Start processing TransactionData as a Codable instead of a json dictionary
- Use this new struct to move structured transaction data around
- Display gas amount * gasPrice in $ as an estimate
- Build a TransactionFeeView that rechecks this estimate and displays the up to date amount with an animation


yozuysal:
https://user-images.githubusercontent.com/83613310/153539752-417326ee-84a0-462e-b196-6b5ae755edaa.mp4

yozuysal:
ping!

Client/Frontend/CryptoWallet/TransactionFee.swift
```diff
@@ -0,0 +1,71 @@
...
+                options: options,
+                transaction: transaction
+            ) { price, estimate in
+                model.gasPrice = price
+                model.gasEstimate = estimate
+            }
+            model.timer = Timer.scheduledTimer(withTimeInterval: 20, repeats: true) { _ in
+                model.updatingGas = true
+                model.updatedOnce = true
+                DispatchQueue.main.asyncAfter(deadline: .now() + 6) {
```
darinwf, line 59: So this is just a fixed 6 second delay before actually updating the price? Ideally you'd want the actual gasPrice query to trigger this sooner possibly or 6 seconds is desired so users notice that we are refreshing?

darinwf (approved):

PR #2856, opened 2022-02-11
Protect against crash due to onDataUpdated running asynchronously

darinwf:
@evancrow - would be great if you could give this a test run since you were seeing this crash


evancrow (approved):
LGTM - Tests seem to pass which was where I was seeing the crash

PR #2857, opened 2022-02-11
Add Basic Logging to Cheatsheet

EdwardJTL:
## Description
Added Logging Cases
- Click Neeva Button open cheatsheet
- "Try Cheatsheet" Tooltip Popover Impression
- Acknowledge education screen on SRP
- Acknowledge education screen on non-SRP pages
- Open link from Cheatsheet
- Empty Cheatsheet

Also added logging attributes to log current query and page url

## Notes
- We only have error log if user is not signed in and tap "Sign In" on the error view
- empty cheatsheet may have false positives
  - this happens when a link is opened from cheatsheet. the cheatsheet data gets cleared before the view has really disappeared. this triggers the `onAppear` in the empty view and logs


macy-n (approved):

macy-n:
LGTM, thanks!

PR #2860, opened 2022-02-11
Switch back to updated TabCardModel synchronously

darinwf:
Since the CardDetails system uses IDs to refer to Tabs, its a problem
if TabCardModel and TabManager are ever out of sync in terms of what
tabs they know about.

So instead of updating TabCardModel asynchronously, we need to update
it synchronously. To achieve that, it's important for TabManager.tabs
to not be a published variable. Published variables notify observers
before they are modified, not after. TabCardModel wants to read from
the TabManager upon notification, so those notifications need to happen
after TabManager.tabs is updated.

This is a good thing as well because it sets the stage to allow us to
optimize a bit more. For bulk operations on TabManager, we can defer
notifications until all of the operations complete. This PR doesn't
really take advantage of that. Just adds some TODO comments about it.

With this change we can undo the guard added to protect buildRows from
accessing a TabCardDetails referencing a Tab that the TabManager no
longer knows about.


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -51,9 +51,9 @@ class TabCardModel: CardModel {
         self.groupManager = groupManager
         // Process updates to the TabManager state asynchronously. This way we wait
         // until after `tabs` has been updated before we do any work.
-        self.subscription = manager.objectWillChange.filter({ [weak self] (_) in
+        self.subscription = manager.tabsUpdatedPublisher.filter({ [weak self] in
             self?.manager.didRestoreAllTabs ?? false
-        }).receive(on: DispatchQueue.main).sink { [weak self] (_) in
+        }).sink { [weak self] in
             self?.onDataUpdated()
```
yozuysal, line 57: if onDataUpdated is going to change a published value, we do want to keep the receive(on: DispatchQueue.main) above in case TabManager somehow updates things in a different thread, no?

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -77,7 +77,9 @@ class TabManager: NSObject, ObservableObject {
         }
     }
 
-    @Published fileprivate(set) var tabs = [Tab]()
+    private(set) var tabs = [Tab]()
+    var tabsUpdatedPublisher = PassthroughSubject<Void, Never>()
```
EdwardJTL, line 81: does this need to be a `private(set)`

Client/Frontend/Browser/Card/CardModels.swift
darinwf: I only added this DispatchQueue just the other day :-)

It used to run synchronously before.

yozuysal: ok, I guess we dont have any instances of tabs updating out of main thread.. :)

yozuysal (approved):

Issue #2862, opened 2022-02-11
Space Headers Have Extra Padding Under Title


evancrow:
Seeing some extra padding under each header:
![image](https://user-images.githubusercontent.com/37356984/153664598-2674a7c8-35a4-4458-8ba4-1ea739097d98.png)

PR #2865, opened 2022-02-12
Fix Extra Padding in Space Header

evancrow:
Fixed: #2862 

Before:

After:

This bug was caused by blank snippets being rendered, so the fix was to make sure the snippet existed, and had content before showing it.


darinwf (approved):

evancrow:
This does not effect tests, going to merge

PR #2867, opened 2022-02-12
use latest logging category flag value for logging

chung-lai:
Enable newly registered user for logging instead of waiting for next app restart to start tracking. This change starts making me think what the best way to observe change in individual flag.


Client/Logging/LogConfig.swift
```diff
@@ -343,15 +347,27 @@ public struct LogConfig {
...
         if enabledLoggingCategories == nil {
             enabledLoggingCategories = Set<InteractionCategory>()
-            NeevaFeatureFlags[.loggingCategories].components(separatedBy: ",").forEach { token in
+            flagsObserver = Defaults.publisher(NeevaFeatureFlags.stringFlagsKey)
+                .combineLatest(
+                    Defaults.publisher(NeevaFeatureFlags.stringFlagOverridesKey)
+                ).sink { _ in
+                    updateLoggingCategory()
+                }
+            updateLoggingCategory()
```
darinwf, line 356: My combine-fu is lacking, but I think maybe it will generate an event when initially hooked up? Maybe this is calling `updateLoggingCategory` twice? Would be good to check.

Client/Logging/LogConfig.swift
```diff
@@ -343,15 +347,27 @@ public struct LogConfig {
 
         if enabledLoggingCategories == nil {
             enabledLoggingCategories = Set<InteractionCategory>()
-            NeevaFeatureFlags[.loggingCategories].components(separatedBy: ",").forEach { token in
+            flagsObserver = Defaults.publisher(NeevaFeatureFlags.stringFlagsKey)
```
darinwf, line 350: We should probably add an API to NeevaFeatureFlags that returns a Publisher you can use to subscribe to the latest value, maybe it can just be called `latestValuePublisher` or something like that.

darinwf (approved):

Client/Logging/LogConfig.swift
chung-lai: Yea, When I did a single publisher `sink`, it calls with initial value but this `CombineLatest` one doesn't. I should double check the documentation of combine to see why it's the case

PR #2872, opened 2022-02-14
Enable default browser interstitial

chung-lai:
Get ready to enabling default browser interstitial once we make the decision. Leaving the default browser promo card as we will iterate on when to show that again to users.


Client/Frontend/Settings/Debug/ExperimentSettingsView.swift
```diff
@@ -22,11 +22,12 @@ struct ExperimentSettingsView: View {
                         .foregroundColor(Color.label)
                 }
 
-                OptionalPrefilledStringField<NeevaExperiment.DefaultBrowserV2>(
-                    "DefaultBrowserPromptView",
-                    value: $valueText,
-                    experiment: .defaultBrowserPromptV2
-                )
+                /* Experiment Example */
+//                OptionalPrefilledStringField<NeevaExperiment.DefaultBrowserV2>(
```
darinwf, line 26: nit: I think swift-format might change this...

darinwf (approved):

PR #2878, opened 2022-02-15
Fix Cheatsheet Promo Popover Issue

### LINKED ISSUE #2875: Neeva Scope promo can show at the wrong times (when opening a space from notifications)
Neeva Scope promo can show at the wrong times (when opening a space from notifications)

I opened the browser via a notification (the one we recently generated for the Superbowl space), and I saw this:

Then I scrolled down a bit, tapped on the superbowl space and saw this:

Then I turned off the screen and then back on and I saw this:

The browser thought it should show this promo because the current URL is a recipe site with a cheatsheet:
https://www.simplyrecipes.com/recipes/moms_roast_turkey/

But we need to be smart about only showing the promo when the WebViewContainer is in the foreground.

EdwardJTL:
## Description
fix #2875 
also fix issue where the popover can still be visible when rotating from inline toolbar to bottom tool bar

- Use presentAsPopover modifier
  - this modifier appears to be more stable and well behaved
  - add call to dismiss presented VC in `viewWillDisappear`

## Note
now `WithPopover` wrapper is only used on some NeevaMenu buttons for Tours. We can possible remove all of them?


EdwardJTL:
@darinwf, when the app is not already running, if we open a spaces link, the popover will flash. I think this is because the app first attempts to load the webview for the last viewed tab. I'm not sure how to fix this.

darinwf:
> ## Note
> now `WithPopover` wrapper is only used on some NeevaMenu buttons for Tours. We can possible remove all of them?

Yes, tours are disabled and deprecated. I think we can remove them.

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetTooltipPopoverView.swift
```diff
@@ -11,7 +11,10 @@ struct CheatsheetTooltipPopoverView: View {
     @Environment(\.colorScheme) var colorScheme
 
     var isDarkMode: Bool { colorScheme == .dark }
-    static let backgroundColorMode: WithPopoverColorMode = .dyanmicBackground(.brand.blue, .brand.variant.polar)
+
+    static var backgroundColor: UIColor = UIColor { (trait: UITraitCollection) -> UIColor in
```
darinwf, line 15: nit: could leave off the `: UIColor` bit

Client/Frontend/Widgets/Popover.swift
```diff
@@ -87,9 +87,13 @@ struct Popover<Content: View>: UIViewControllerRepresentable {
             }
         }
 
+        override func viewWillDisappear(_ animated: Bool) {
+            presentee = nil
+            super.viewWillDisappear(animated)
+        }
+
         override func viewDidDisappear(_ animated: Bool) {
             // Dismiss the popover when removing this view controller.
```
darinwf, line 96: nit: We can remove this override then, right? And the comment should move to `viewWillDisappear`

darinwf (approved):

darinwf:
> @darinwf, when the app is not already running, if we open a spaces link, the popover will flash. I think this is because the app first attempts to load the webview for the last viewed tab. I'm not sure how to fix this.

Startup sequencing is a bit messy. We can take this issue as a separate matter. Would be great if you could file an issue for it and attach a screencapture. Thanks!

EdwardJTL:
> > @darinwf, when the app is not already running, if we open a spaces link, the popover will flash. I think this is because the app first attempts to load the webview for the last viewed tab. I'm not sure how to fix this.
> 
> Startup sequencing is a bit messy. We can take this issue as a separate matter. Would be great if you could file an issue for it and attach a screencapture. Thanks!

#2884

PR #2879, opened 2022-02-15
[ios] remove recipe sites that we do not want to show in cheatsheet

macy-n:
Remove recipes sites that we got request to not show recipe details. On the blacklisted site, we will not show recipe details.

image

FYI @kelvinso


EdwardJTL (approved):
LGTM

darinwf:
@macy-n - LGTM... any thoughts on how to unify this with the server side record so we aren't duplicating this information and can more easily keep it in sync?

PR #2882, opened 2022-02-15
Change wallet transaction UI to new designs

yozuysal:
- Separates WalletSequenceContent into many small Views
- This allows readability and also simplifies the logic quite a bit
- Starts showing malicious site views as a separate top level view
- Changes all text, layout and padding according to new designs


yozuysal:
Screen Shot 2022-02-14 at 8 16 34 PM
Screen Shot 2022-02-14 at 8 22 16 PM

yozuysal:
This is not 100% done (a few icons and behavior changes to be added as a followup), but I still think it is worth landing like this and testing for a day.

darinwf (approved):

PR #2887, opened 2022-02-15
inline tab group BuildRows logic refactoring

chassu:
Remove the function 'modifyAllDetailsAvoidingSingleTabs' and simply process the tabs/tab groups in order. If the row we're working on can fit the next incoming tab/tab group, then put them in the same row. 

Based on a previous design sync, here are the rules for displaying inline tab groups:
- Tab groups must have at least two tabs
- If the tab group has fewer tabs than the number of remaining columns in the current row, display it in-line
- If the tab group has more tabs than the number of remaining columns in the current row, but fewer tabs than the number of columns in the tab switcher, start a new row containing the tab group
- If the tab group has more tabs than the number of columns in the tab switcher, then make the tabs in the group scroll horizontally and show the expand/collapse button

For more information, please refer to https://www.figma.com/file/t1jW7cahkClVHAIlnHJuv8/Tab-Manager-(Spaces-%26-Tab-Groups)?node-id=2291%3A8659

### DIFF
--- a/Client/Frontend/Browser/Card/CardModels.swift
+++ b/Client/Frontend/Browser/Card/CardModels.swift
@@ -60,6 +60,12 @@ class TabCardModel: CardModel {
 
     struct Row: Identifiable {
         var id: Set<String> { Set(cells.map(\.id)) }
+        var numTabsInRow: Int {
+            cells.reduce(0, {x, y in
+                x + y.numTabs
+            })
+        }
+
         enum Cell: Identifiable {
             case tab(TabCardDetails)
             case tabGroupInline(TabGroupCardDetails)
@@ -86,6 +92,17 @@ class TabCardModel: CardModel {
                     return details.allDetails[range].contains { $0.isSelected }
                 }
             }
+
+            var numTabs: Int {
+                switch self {
+                case .tab(_):
+                    return 1
+                case .tabGroupInline(let details):
+                    return details.allDetails.count
+                case .tabGroupGridRow(let details, _):
+                    return details.allDetails.count
+                }
+            }
         }
         var cells: [Cell]
     }
@@ -109,11 +126,12 @@ class TabCardModel: CardModel {
                 && tab.isIncognito == incognito
         }.reduce(into: []) { partialResult, details in
             let tabGroup = tabGroupModel.allDetails.first(where: { $0.id == details.rootID })
-            if partialResult.isEmpty || partialResult.last?.cells.count == maxCols
+            if partialResult.isEmpty || partialResult.last!.numTabsInRow >= maxCols
                 || tabGroup != nil
             {
                 if let tabGroup = tabGroup {
                     if tabGroup.isExpanded {
+                        // tabGroupGridRow always occupies a row by itself.
                         for index in stride(from: 0, to: tabGroup.allDetails.count, by: maxCols) {
                             var max = index + maxCols
                             if max > tabGroup.allDetails.count {
@@ -124,7 +142,9 @@ class TabCardModel: CardModel {
                                 Row(cells: [Row.Cell.tabGroupGridRow(tabGroup, range)]))
                         }
                     } else {
-                        if (tabGroup.allDetails.count + (partialResult.last?.cells.count ?? 0)) <= maxCols {
+                        // If there's enough remaining columns, fit the tab group in the same row with individual tabs.
+                        // Otherwise, build a horizontal scroll view in the next row.
+                        if (tabGroup.allDetails.count + (partialResult.last?.numTabsInRow ?? 0)) <= maxCols && !partialResult.isEmpty {
                             partialResult[partialResult.endIndex - 1].cells.append(.tabGroupInline(tabGroup))
                         } else {
                             partialResult.append(Row(cells: [Row.Cell.tabGroupInline(tabGroup)]))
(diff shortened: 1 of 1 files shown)


chassu:
Before:

Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-15 at 02 41 24

After:

simulator_screenshot_81F4E6AD-A7E1-4182-981A-00AE06EE8D74

chassu:
I'm aware there's some padding issues 😔

darinwf:
Help me understand the plan here. What is the goal of removing modifyAllDetailsAvoidingSingleTabs?

darinwf:
@chassu and I discussed a variation on this approach today. He's going to update the PR.

chassu:
Got the logic as close to what I wanted... but l can think of a few cases that will still leave spaces in the switcher. For example, if we have [[tab, tab], tab, [tab, tab]], and let's assume there are 4 columns:

The first two tabs get inserted, and then the individual tab gets processed, which then gets inserted to the same row because there are enough remaining columns:

[[tab, tab], tab, empty_space]

The remaining column can't fit the last tab group, so that leaves us with:

[[tab, tab], tab, empty_space]
[[tab, tab]]

 Technically, the first row can fit two tab groups, and then leave the individual tab to the second row. 
 
 [[tab, tab], [tab, tab]]
[tab]

I think this happen because of the nature of my heuristics. I'm not sure if it's necessary to also cover that case... it's getting to a point where it's solving an optimization problem to use minimum space 😆 

Here's an example of how it works now:

https://user-images.githubusercontent.com/87332917/154603812-16f59944-6198-49a6-961c-c6f42a00522f.mp4

chassu:
AFAIU, the behaviors here are all following the rules based on the design:

https://user-images.githubusercontent.com/87332917/154604920-69d0ff02-1c98-47d1-809d-c03f3a6eb405.mp4

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -96,52 +116,128 @@ class TabCardModel: CardModel {
...
+        var allDetailsFiltered = allDetails.filter { tabCard in
             let tab = tabCard.manager.get(for: tabCard.id)!
             return
                 (tabGroupModel.representativeTabs.contains(tab)
                 || allDetailsWithExclusionList.contains { $0.id == tabCard.id })
                 && tab.isIncognito == incognito
-        }.reduce(into: []) { partialResult, details in
+        }
+
+        // An array to keep track of whether a CardDetail has been processed. This allows us to skip
```
darinwf, line 136: Love the comments!

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -86,8 +94,20 @@ class TabCardModel: CardModel {
...
+                case .tabGroupGridRow(_, let range):
+                    return range.count
+                }
+            }
         }
         var cells: [Cell]
+        var multipleCellTypes: Bool = false
     }
 
     func buildRows(incognito: Bool, tabGroupModel: TabGroupCardModel, maxCols: Int) -> [Row] {
```
darinwf, line 113: I think we need to figure out how to write unit tests for `buildRows`. This is just begging for unit testing.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -60,6 +60,14 @@ class TabCardModel: CardModel {
 
     struct Row: Identifiable {
         var id: Set<String> { Set(cells.map(\.id)) }
+        var numTabsInRow: Int {
+            cells.reduce(
+                0,
+                { x, y in
```
darinwf, line 66: nit: I think this would be slightly clearer if you named the parameters to the callback `total, cell` instead. That way you use descriptive names.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -96,52 +116,128 @@ class TabCardModel: CardModel {
...
+            // Row.multipleCelltypes is needed for the UI to create spacing between different types of cells.
+            // This value won't be evaluated in the UI if the row contains only individual tabs. So, it's
+            // set to true as long as some tab/tab group is promoted.
+            if didPromote {
+                row.multipleCellTypes = true
+            }
+        }
+
+        for (index, details) in allDetailsFiltered.enumerated() {
+            if processed[index] == true { continue }
```
darinwf, line 177: You can also just write `if processed[index] { continue }` since the value of `processed[index]` is already `Bool`

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -96,52 +116,128 @@ class TabCardModel: CardModel {
...
-                if tabGroup != nil {
+                // Insert a new row (following expanded tab group) for individual tabs
+                if tabGroup != nil && tabGroup!.isExpanded {
                     partialResult.append(Row(cells: []))
                 }
             } else {
                 partialResult[partialResult.endIndex - 1].cells.append(.tab(details))
             }
-        }.filter { !$0.cells.isEmpty }
+            processed[index] = true
```
darinwf, line 228: Is it actually important to stamp `processed[index] = true` here? This `index` is never going to be visited again anyways right?

Client/Frontend/Browser/Card/CardModels.swift
chassu: Wrote some tests in CardTests.Swift

Client/Frontend/Browser/Card/CardModels.swift
chassu: Fixed

Client/Frontend/Browser/Card/CardModels.swift
chassu: Fixed

Client/Frontend/Browser/Card/CardModels.swift
chassu: Fixed

darinwf (approved):
LGTM, thanks for the tests!

Issue #2877, opened 2022-02-14
Disable swipe-to-close for now


darinwf:
From UX review:

- Concerned that the current swipe-to-close treatment is too fiddly and that it is easy for people to accidentally close tabs; also easy for users to mistake the feature as drag-n-drop to group or reorder and then be surprised that it closes the tab.
- Swipe-to-close not working with collapsed groups is a concern as well.
- Pinch to close is overloading a zoom gesture, which might also surprise users or not be very discoverable
- Drag to trash area makes sense, but what we might do is have an “editing mode” for tabs that users activate through a new action in the overflow menu. In editing mode, users can drag tabs to a trash area, or drag to reorder / group / ungroup tabs.
- Would like to experiment with disabling swipe-to-close (i.e., put it behind a feature flag), and see if users complain about the absence of swipe-to-close. This will help us understand what matters to our user base and help us prioritize this feature.

evancrow:
Another idea discussed in #2888 is a two-finger swipe for swipe to close, I will take a look into this

darinwf:
This was done.

PR #2888, opened 2022-02-15
Put swipe-to-close tabs behind a feature flag

darinwf:
Disabling as this feature doesn't work with collapsed tab groups and
is a bit finicky. More details in the linked bug about this change.

Issue #2877


yozuysal (approved):

darinwf:
> :(

I share your sentiment, as I'm a regular user of this feature, but I don't know a better way at the moment... open to ideas!

evancrow:
> > :(
> 
> I share your sentiment, as I'm a regular user of this feature, but I don't know a better way at the moment... open to ideas!

I'll start looking into edit mode possibilities, but maybe for swipe to close we should require two fingers? (i.e. a two-finger swipe left/right)

darinwf:
> I'll start looking into edit mode possibilities, but maybe for swipe to close we should require two fingers? (i.e. a two-finger swipe left/right)

That's a super interesting idea! Let's try it out behind the flag.

PR #2890, opened 2022-02-15
Fix default browser interstitial disappearing upon sign in

chung-lai:
With the various login flow, certain flow is waiting for an auth request to redirect to Neeva home. Once the redirection happened, the web view dismiss all overlays which dismisses the interstitial.

This also fix the logging for certain sign in flow which the cookie is not ready until later.


evancrow:
Unfortunate my fix didn't work but that's for taking a look!

evancrow (approved):

PR #2893, opened 2022-02-15
Cleanup Docs/ folder removing stale content

darinwf:


darinwf:
I'm thinking it would be nice to try building / developing the app using a non-corp mac and with a different Apple ID to see what works and what doesn't work. Can then re-add a section to BUILDING.md about what is needed.

yozuysal (approved):

PR #2897, opened 2022-02-16
add option to close tab in context menu

chassu:
Providing another way to close the tab. Will be necessary for tab pinning since the close button will be changed.


darinwf (approved):
LGTM

Did you get a chance to run this by Ria?

PR #2898, opened 2022-02-16
Update maybeRewriteURL in NavigationPath

EdwardJTL:
## Description
Update the routine to better support rewriting search urls
- Add list of [all currently supported google domains](https://www.google.com/supported_domains)
- Update matching (from observation in manual testing)
  - yahoo: path is of the format `search;_ylt=A2KIbMtxhw1i.6EAoTly.9w4;_ylu=Y29sbwNiZjEEcG9zAzEEdnRpZANDb2NvMzAwXzEEc2VjA3Nj`
  - baidu: 
    - path is of the format `/from=844b/s`
    - keyword is `word`
  - so: 
    - add domain `m.so.com`
    - keyword is `q`
  - sogou: 
    - add domain `wap.sogou.com`
    - path is of the format `/web/searchList.jsp`
    - keyword is `keyword`
- Add new enum case to refactor the code
  - use a String backed hashmap instead of switch case matching. We can go back to using a switch statement if we find it more desirable.


Client/Application/NavigationRouter.swift
```diff
@@ -265,40 +265,282 @@ enum NavigationPath {
...
+        enum SearchPathType {
+            case google
+            case search
+            case none
+            case searchTouch
+            case s
+            case web
+
+            static var map: [String: Self] = {
+                let googleDomains = [
```
darinwf, line 277: nit: Hmm, I think it might help readability to move this googleDomains definition out to a separate `.swift` file or to at least move this to the top of the file.

Alternatively, it could just make sense to move this entire function out to a separate file. Then this definition could just live at the top of that file.

Yeah that's what I would probably do :)

darinwf (commented):
Probably makes sense to write some unit tests (adding to ClientTests target) for this function, providing some coverage for the interesting cases.

darinwf:
Thanks for being so thorough in the analysis and updating this function to be way more robust!

Client/Application/NavigationRouter.swift
EdwardJTL: yea i agree we should move this outside. the difference is that right now this enum is file private. we will lose access control if we do that.

darinwf: By "move this entire function out to a separate file", I meant to move `maybeRewriteURL`.

EdwardJTL: gotcha :D will do

EdwardJTL:
> Probably makes sense to write some unit tests (adding to ClientTests target) for this function, providing some coverage for the interesting cases.

i added hardcoded url tests.

I tried doing XCUITests where we would go to their website and search a query so we can read their url.
with google, i was able to select the query box so that the flashing cursor shows up and the keyboard becomes in focus, but still i can't access the textfield to type in a query...

Client.xcodeproj/project.pbxproj
```diff
@@ -120,6 +120,8 @@
 		3215C67727B6191B00D593BF /* FlowSound.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3215C66A27B6191B00D593BF /* FlowSound.swift */; };
 		3215C67827B6191B00D593BF /* NSShadow+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3215C66B27B6191B00D593BF /* NSShadow+Extension.swift */; };
 		3215C67A27B6194D00D593BF /* CharacterSet+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3215C67927B6194D00D593BF /* CharacterSet+Extension.swift */; };
+		324948F227BDE84B00082FCF /* NavigationRouter+maybeRewriteURL.swift in Sources */ = {isa = PBXBuildFile; fileRef = 324948F127BDE84B00082FCF /* NavigationRouter+maybeRewriteURL.swift */; };
```
evancrow, line 123: small nit: Can you rename this file to be NavigationRouter+MaybeRewriteURL (capitalize the M)

Client/Application/NavigationRouter+maybeRewriteURL.swift
```diff
@@ -0,0 +1,320 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+import Shared
+
+extension NavigationPath {
+
```
evancrow, line 9: nit: remove extra line

Client/Application/NavigationRouter+maybeRewriteURL.swift
```diff
@@ -0,0 +1,320 @@
...
+                return true
+            case .searchTouch:
+                return path == "/search/touch/"
+            case .s:
+                // baidu path may start with /from
+                return path.contains("/s")
+            case .web:
+                // sogou path contains /searchList
+                return path.starts(with: "/web")
+            @unknown default:
```
evancrow, line 251: Can remove the `@unknown` here

Client/Application/NavigationRouter+maybeRewriteURL.swift
```diff
@@ -0,0 +1,320 @@
...
+                uniqueKeysWithValues: zip(allDomains, labelsPartA + labelsPartB)
+            )
+        }()
+
+        private func isPathAMatch(path: String) -> Bool {
+            switch self {
+            case .google, .search:
+                // yahoo path starts with /search
+                return path.starts(with: "/search")
+            case .none:
```
evancrow, line 241: I think if you remove `@unknown` (see earlier comment), you can remove this case

Client/Application/NavigationRouter+maybeRewriteURL.swift
```diff
@@ -0,0 +1,320 @@
...
+            //          - some : "UTF-8"
+            //      ▿ 3 : hl=en
+            //        - name : "hl"
+            //        ▿ value : Optional<String>
+            //          - some : "en"
+            //      ▿ 4 : client=safari
+            //        - name : "client"
+            //        ▿ value : Optional<String>
+            //          - some : "safari"
+            //
```
evancrow, line 282: Remove extra `//`

Client/Application/NavigationRouter+maybeRewriteURL.swift
```diff
@@ -0,0 +1,320 @@
...
+                let pathType = Self.map[host],
+                pathType.isPathAMatch(path: components.path),
+                let queryItems = components.percentEncodedQueryItems
+            else { return nil }
+            switch pathType {
+            case .google, .search:
+                // yahoo uses p for the search query name instead of q
+                let queryName = (host == "search.yahoo.com") ? "p" : "q"
+                return queryItems.first(where: { $0.name == queryName })?.value
+            case .none:
```
evancrow, line 293: Think the same thing here with remove `@unknown` and `.none`

evancrow (commented):
Cool! Just a few nits...

Client/Application/NavigationRouter+maybeRewriteURL.swift
EdwardJTL: the none case is for duckduckgo

darinwf:
> I tried doing XCUITests where we would go to their website and search a query so we can read their url. with google, i was able to select the query box so that the flashing cursor shows up and the keyboard becomes in focus, but still i can't access the textfield to type in a query...

Unit tests (ClientTests) is the right thing for this situation. XCUITests is really best for cases that you can't test using unit tests. Reason: XCUITests tend to be slow and can be susceptible to flakiness.

Client/Application/NavigationRouter+MaybeRewriteURL.swift
```diff
@@ -0,0 +1,314 @@
...
+            "www.google.com.vn",
+            "www.google.vu",
+            "www.google.ws",
+            "www.google.rs",
+            "www.google.co.za",
+            "www.google.co.zm",
+            "www.google.co.zw",
+            "www.google.cat",
+        ]
+        private static let searchDomains = [
```
darinwf, line 209: These other domains probably have localized variants too, but we can worry about that in another PR :)

darinwf (approved):

Client/Application/NavigationRouter+MaybeRewriteURL.swift
EdwardJTL: yes i found that if i access yahoo from a browser, it takes me to ca.search.yahoo.com. i think the google domains are more visible because spotlight will use those links.

Client/Application/NavigationRouter+maybeRewriteURL.swift
evancrow: Two questions:
1) why is this named none instead of duckduckgo
2) if there is a specific reason, can it be renamed to something else since `none` is a default Swift option that we should try to not override.

evancrow (approved):
Besides my last question LGTM

Client/Application/NavigationRouter+maybeRewriteURL.swift
EdwardJTL: its named none because there is no path in duck duck go queries; we name all the other cases based on what the path says rather than which search engines use it. afaik `none` is only a protected keyword for Optionals. I could change this to `empty`?

PR #2899, opened 2022-02-17
Make the wallet fullscreen

yozuysal:
This finishes almost all of the main UI changes and only leaves us side menus to be finalized

- Wallet Screen is now fullscreen using a fullscreen modal in OverlayManager
- Overhaul SessionInfoView according to new design
- Wallet is now the last menu item in overflow with branded wordmark
- Fixed disconnect/switch from the dashboard working properly
- SendForm changed to the new spec
- Other cosmetic fixes


Client/Frontend/CryptoWallet/Assets/CollectionViews.swift
```diff
@@ -196,21 +199,28 @@ struct CollectionView: View {
...
-                    .scaledToFit()
-                    .frame(height: 18)
-                    .padding(.vertical, 20)
             }
             .offset(y: -28)
         }
     }
 }
+
+/*
```
evancrow, line 224: Is there a reason for this?

evancrow (approved):
LGTM

Client/Frontend/CryptoWallet/Assets/CollectionViews.swift
yozuysal: :( I wish. I totally blame lintswift for making me park views inside comments because history is lost.

PR #2901, opened 2022-02-17
Add some more logging to getTabFor

darinwf:
Trying to investigate why it seems like this function is not finding my
existing tabs. It's been hard to pin down a repro case. Hoping I can
just check the logs next time it happens.


evancrow (approved):
LGTM

PR #2902, opened 2022-02-17
Add Browsing Activities to Spotlight Search

EdwardJTL:
## Description
This change allows users to search for websites visited in the neeva browser in iOS spotlight search (left screen to the home screen or drag down on the home screen)

This revives parts of the old firefox code that was removed five years ago. See issues [here](https://bugzilla-dev.allizom.org/show_bug.cgi?id=1390200) and [here](https://bugzilla-dev.allizom.org/show_bug.cgi?id=1390294). We want to test this and keep an eye on if this leads to any crashes.

instead of using the `webpageURL` property on the user activity, we keep the absolute string in the user info dict; this allows our app to be opened even when its not set as the default browser

## To-Dos
we are also missing favicons atm; they have to be supplied via data in memory or fileURL on the local file system.

## Behaviour
When a url or the page metadata is loaded, we invalidate the old activity in the tab and create a new user activity which is marked as active. We add rich data to the activity from the page metadata if available. 

This means for each url, there is one invalid and one current activity when the page is fully loaded, but it does not seem to affect indexing (no duplicate entries observed).

We also have the option to manually remove activities from indexing by
1) deleting all, or
2) deleting by a persistent identifier


EdwardJTL:
implementing a workaround that will still launch the neeva app even if its not set as the default browser...

Client/Helpers/UserActivityHandler.swift
```diff
@@ -36,8 +36,17 @@ class UserActivityHandler {
...
+        userActivity.isEligibleForSearch = true
+        userActivity.requiredUserInfoKeys = ["url"]
+        userActivity.userInfo = ["url": url.absoluteString]
+        // we can set userActivity.keywords = [String] to specify additional queries with which this item can be found in spotlight
+        let attributes = CSSearchableItemAttributeSet(contentType: .url)
+        attributes.title = tab.pageMetadata?.title
+        attributes.contentDescription = tab.pageMetadata?.description
+        userActivity.contentAttributeSet = attributes
+        userActivity.needsSave = true
         userActivity.becomeCurrent()
```
darinwf, line 50: Can you summarize how this makes it so that spotlight will generate those cards? Maybe a comment block in this code would be helpful.

EdwardJTL: see comments

macy-n:
Tested drove and looks great! Can wait for Darin to confirm before merging.

PR #2905, opened 2022-02-17
Increase Spacing in Cheatsheet Info View between Text and Image

EdwardJTL:
## Description
Slightly larger spacing

i know this doesnt look exactly like the mock, its hard to do with the infrastructure we have for overlay sheet and popover on iphone and ipad :(


EdwardJTL:
one idea is to try using three ZStacks to layout the views and manually calculate the positions and sizes while keeping the text and button at the top and bottom

macy-n (approved):

PR #2907, opened 2022-02-17
Add Spaces to Spotlight Search

EdwardJTL:
## Description
- Add spaces to spotlight search with CSSearchableItem
  - All operations are done in `onUpdateSpaces`
    - index items are created or updated from the set of new spaces
    - the difference between old and new set is used to remove any items
- Add opening space functionality from scene delegate

## Future Work
- We can create our own UTType for richer results and potentially allow sharing "Space" objects
- The infinity spin loop problem of opening an non-existent space is more visible now.
  - Create a space
  - exit app. find space in spotlight
  - go back to app, delete space
  - open spotlight again. the search result is still there
  - click result and get stuck :( 
 we had a conversation in slack about it last week but no AI came out of it


Client/Application/SceneDelegate.swift
```diff
@@ -169,6 +170,11 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     }
 
     func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
+        if userActivity.activityType == CSSearchableItemActionType,
+           let itemIdentifier = userActivity.userInfo?[CSSearchableItemActivityIdentifier] as? String {
+            // this itemIdentifier is the spaces id
+            self.bvc.browserModel.openSpace(spaceId: itemIdentifier, bvc: self.bvc, completion: {})
```
darinwf, line 176: Should we do any validation of `itemIdentifier` here? What happens if the space does not exist?

I know you are taking care to remove items from the index corresponding to deleted spaces, but there could be a race condition. The spotlight index might update lazily or something like that.

Shared/API/SpaceStore.swift
```diff
@@ -387,6 +393,36 @@ public class SpaceStore: ObservableObject {
             }
         }
 
+        // add spaces to CS
+        Self.searchableIndex.indexSearchableItems(
```
darinwf, line 397: This is already a long function. I think it would help to add this code to a private helper function.

Client/Application/SceneDelegate.swift
EdwardJTL: yea it's very easy for spotlight to give a card to a space that should be removed (for example, it doesn't update the result page when you go back to spotlight search)...
`browserModel.openSpace` does a check for the id and fetches if it's not there. This will put the app in an infinite loading state if the space is indeed not valid. we have been discussing on slack on how we should treat invalid space ids.

EdwardJTL: https://github.com/neevaco/neeva-ios/issues/2914

macy-n (approved):

Issue #2896, opened 2022-02-16
No restore option after closing all tabs


chung-lai:
We are exploring option not to show confirmation alert when user clicks "Close All Tabs" option; however, there is no restore option after closing all tabs. Is this an intended behavior?

cc @evancrow, @darinwf, @EdwardJTL

evancrow:
Yes, we intentionally suppress the Toast because users have to confirm

darinwf:
I notice that when using the "Close All Tabs" feature in the Tab Group context menu, I get both a confirmation prompt and a toast. However, when using the "Close All Tabs" feature from the switcher overflow menu, I only get a confirmation prompt. Perhaps these should behave consistently?

I think it is reasonable to not show the toast following a confirmation prompt.

PR #2908, opened 2022-02-18
Add Option to Bypass Confirm Close All Tabs

evancrow:
Fixed: #2896

Adds a setting that will skip the confirm close all tabs dialog, and shows a Toast to restore all the tabs instead.


darinwf:
Actually, can you please add some test coverage for this setting?

Issue #2912, opened 2022-02-18
Presented Overlay should be dismissed when opening the app with a link


EdwardJTL:
The overlay stays open. It doesn't make sense, especially for cheatsheet.

PR #2921, opened 2022-02-22
Hide Overlay When Opening URL

evancrow:
Fixed: #2912


evancrow:
This openURLInNewTab path is called whenever a URL is opened from outside the app, or from a Neeva app element (i.e. help center)

EdwardJTL:
if we open a link to a page that is already in a tab, we switch tab but the overlay is still open

evancrow:
Hm, maybe it makes more sense to add this to `switchToOrCreateTab` then

EdwardJTL (approved):
nit: theres white space on line 954

PR #2923, opened 2022-02-22
Switch to XCFrameworks

j-f1:
This allows building from M1 Macs, which use the same arm64 architecture for both the simulator and real iOS devices. Additionally, this removes the need for the `carthage copy-frameworks` build step.

After merging, I recommend deleting the `Carthage/Build` directory and re-running `carthage_command.sh` to build the new frameworks.

Also: runs on macOS :)

Screenshot_2022-02-22 08 55 04


evancrow (approved):
This LGTM! I assume this still works fine on Intel macs as well right?

j-f1:
I no longer have an Intel Mac to test on — can you give it a try

evancrow:
Works fine. For future reference, if you right click on an application in Finder, then Info and select "Run with Rosette" it will run like it's on an Intel Mac.

evancrow:
Once you merge, I'll alert everyone to rebuild carthage

j-f1:
> Works fine. For future reference, if you right click on an application in Finder, then Info and select "Run with Rosette" it will run like it's on an Intel Mac.

That doesn’t seem to work well for Xcode as it still seems to think it’s on Apple Silicon even with that checked.

Also, I no longer have merge access — can you merge for me?

evancrow:
Sure, so good to merge now?

j-f1:
Seems like it works for both of us so I think so!

Issue #2616, opened 2022-01-18
Fix “popover” sheets in portrait


jed-fox-neeva:
Steps to reproduce:

1. Rotate to landscape
2. Tap add to space in toolbar
3. Rotate to portrait

<img src=https://user-images.githubusercontent.com/76001228/150009178-113af35b-62a7-4faa-a073-6d11088385e0.jpeg width=390>

PR #2925, opened 2022-02-22
Fix Popover Layout

evancrow:
Fixed: #2616 

Instead of setting a width with a constant divisor, it makes more sense to use padding which changes based on available space. It also fits the height of the popover to content, so a sheet is only as tall as it needs to be.


Client/Frontend/Overlay/Popover/PopoverView.swift
```diff
@@ -78,18 +79,38 @@ struct PopoverView<Content: View>: View {
...
+                }
+                .padding(.horizontal, paddingForSizeClass(horizontalSizeClass))
+                .padding(.vertical, paddingForSizeClass(verticalSizeClass))
             }
             .accessibilityAction(.escape, onDismiss)
         }
     }
+
+    func paddingForSizeClass(_ sizeClass: UserInterfaceSizeClass?) -> CGFloat {
+        let defaultValue: CGFloat = 12
```
EdwardJTL, line 101: could we condense this into
`if let class = sizeClass, case .regular { return 50 } else { return 12}`

Client/Frontend/Overlay/Popover/PopoverView.swift
```diff
@@ -78,18 +87,29 @@ struct PopoverView<Content: View>: View {
...
                     }
                     .clipShape(RoundedRectangle(cornerRadius: 16))
                     .background(
                         Color(style.backgroundColor)
                             .cornerRadius(16)
                     )
-                    .padding()
-                }.frame(width: geo.size.width / 1.5)
+                    // 60 is button height + VStack padding
+                    .frame(maxHeight: geo.size.height - paddingForSizeClass(verticalSizeClass) - 60)
```
EdwardJTL, line 98: could this call be replaced with verticalPadding?

evancrow: Missed that one!

EdwardJTL (approved):

PR #2927, opened 2022-02-23
Add sink from selectedTabSubscription to CardDetails

chassu:
This makes the CardDetails update possible, but we won't see the performance improvement right away. CardGrid body is still being evaluated two times when switching between incognito and normal mode. After this, I will look into removing the delegate function and use the same selectedTabPublisher.


Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -179,6 +180,15 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
         isPinnedSubscription = tab.$isPinned.sink { [weak self] _ in
             self?.objectWillChange.send()
         }
+
+        selectedTabSubscription = manager.selectedTabPublisher
+            .sink { tab in
+                if tab?.id == self.id && manager.selectedTab?.tabUUID != self.id {
+                    self.objectWillChange.send()
+                } else if manager.selectedTab?.tabUUID != self.id && tab?.id == self.id {
```
yozuysal, line 188: these conditions look the same to me :)

yozuysal (commented):
is the reason for changing the timing to be able to construct the condition for "was this not selected before and is selected now or vice versa"?

If yes, I feel like this will be pretty fragile. There is nothing in the publisher that hints of this property and why it is important. So down the line, it might be moved back again, causing potential confusion.

Thankfully, Combine to our rescue! .withPrevious() will combine the previous value published and send a tuple (if you search for this, you will find a few uses of it in our codebase as well). So you can just use .withPrevious() without the TabManager side changes.

EdwardJTL:
is there a reason we have a selected tab property and a publisher? could we use one current value publisher?

Client/Frontend/Browser/Card/CardDetails.swift
chassu: 😂

yozuysal (approved):

chassu:
Could you explain where the fragileness comes from? The goal of this change is hoping that we can further optimize the system such that when a select tab happens, the only thing that's causing the UI to change is CardDetails' `objectWillChange`. Now, we're getting that update kind of like a side effect from the delegate function BVC implements, and also due to a sink from selectedTabPublisher to Web3Model, if I'm not mistaken. (Web3Model is an EnvironmentObject of CardGrid)

> is the reason for changing the timing to be able to construct the condition for "was this not selected before and is selected now or vice versa"?
> 
> If yes, I feel like this will be pretty fragile. There is nothing in the publisher that hints of this property and why it is important. So down the line, it might be moved back again, causing potential confusion.
> 
> Thankfully, Combine to our rescue! .withPrevious() will combine the previous value published and send a tuple (if you search for this, you will find a few uses of it in our codebase as well). So you can just use .withPrevious() without the TabManager side changes.

chassu:
^Adding on to the above, this PR won't actually achieve any improvement yet, since it's only adding logic to the system. 

Do you think this is a reasonable direction to head toward?

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -179,6 +180,16 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
         isPinnedSubscription = tab.$isPinned.sink { [weak self] _ in
             self?.objectWillChange.send()
         }
+
+        selectedTabSubscription = manager.selectedTabPublisher
+            .withPrevious()
```
darinwf, line 185: I was reading comments on `withPrevious`, and I think we should insert a `.prepend(nil)` before it. That way the very first change to `TabManager.selectedTab` will still be reported to the sink. I think that's actually pretty important so you don't miss the first event.

darinwf (changes requested):

darinwf:
Although we haven't completed the other work necessary to make this PR result in a performance improvement, I still think it is technically the right change. I think you should proceed to land and then we can work on the performance improvements (eliminating other extraneous updates) separately.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -132,6 +132,7 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
 
     public let id: String
     private var isPinnedSubscription: AnyCancellable?
+    private var selectedTabSubscription: AnyCancellable?
```
darinwf, line 135: nit: since both of these subscriptions are just used to keep the sink alive, but we never try to cancel an individual one, we should just use a `Set<AnyCancellable>` here to store the subscriptions, i.e.:

```
private var subscriptions: Set<AnyCancellable> = []
```

Then use the `.store(in: &subscriptions)` method.

darinwf (approved):

PR #2928, opened 2022-02-23
Fix Edit URL UI

evancrow:
Before:

After:


evancrow:
Works! Just uploaded a color fix...

PR #2929, opened 2022-02-23
CircleCI Changes

evancrow:
Fixes for CircleCI caused by a breakage from #2923 

FYI @j-f1


j-f1:
If that works, you should be able to add `-v2` to the cache keys to retain the benefits of the cache while throwing out any old caches.

evancrow:
Working now @j-f1!

chassu (approved):
Can you explain the v2 thing?

evancrow:
The V2 just changes the ID of the cache. That way the simulated mac has to rebuild Carthage using the new XCFramework.

Issue #2918, opened 2022-02-21
NeevaScope Intro Cutoff on Compact Devices


evancrow:
On iPhone SE:
![image](https://user-images.githubusercontent.com/37356984/154884353-1639685d-0574-41b5-80e2-7c7537a87b66.jpeg)

FYI @EdwardJTL

EdwardJTL:
I'll see if I can add the middle preferred height to this; not sure if the animation will look good. 

Otherwise I'll try to wrap things in zstacks and manually calculate image size.

EdwardJTL:
fixed in #2933

PR #2933, opened 2022-02-24
Adjust Cheatsheet Intro Screen Size for SE

EdwardJTL:
## Description

Add a `overlayMinHeightToFillScrollView` environment object to read the size containing the scrollview so that the item can take up the full size with a `Spacer()`.

## Notes
- The size read on compact device is not correct; it is slightly larger than what the scrollview can display on the screen. I am not sure why. @evancrow do you have any ideas on why this is the case? maybe it has something to do with the home button?
- The geometry reader doesn't work in `PopoverView` :( It just shrinks the view to no height


EdwardJTL:
#2918

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuView.swift
```diff
@@ -87,44 +87,26 @@ struct QueryButton: View {
...
             }
+
             Group {
                 Spacer(minLength: 10)
                 // hide the image on small iPhones in landscape
                 if horizontalSizeClass == .regular || verticalSizeClass == .regular {
                     Image("cheatsheet", bundle: .main)
                         .resizable()
                         .scaledToFit()
-                        .frame(minHeight: 115, maxHeight: 500)
```
evancrow, line 127: Seems like these values are kinda arbitrary... any way they can be calculated? (or at least add a comment)

Client/Frontend/Overlay/Sheet/OverlaySheetView.swift
```diff
@@ -226,9 +226,12 @@ struct OverlaySheetView<Content: View, HeaderContent: View>: View, KeyboardReada
...
                 } else {
-                    ScrollView(model.position == .top ? [.vertical] : [], showsIndicators: false) {
-                        sheetContent
-                            .padding(.bottom, OverlaySheetUX.bottomPadding)
+                    GeometryReader { proxy in
+                        ScrollView(model.position == .top ? [.vertical] : [], showsIndicators: false) {
+                            sheetContent
+                                .padding(.bottom, OverlaySheetUX.bottomPadding)
+                                .environment(\.overlayMinHeightToFillScrollView, proxy.size.height)
+                        }
```
evancrow, line 234: Instead of wrapping a GeometryReader which might cause some of the flakiness you were describing, you could use `onHeightOfViewChanged` modifier.

evancrow (commented):
Maybe the home button is causing some of the issues you were describing. Since there isn't any `SafeArea` it might be messing with the calculations.

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuView.swift
EdwardJTL: added a routine to calculate it. 
i still can't add a reader to popover. The popover just shrinks to the height of the content in the scrollview if the content is not tall, so i don't have a chance to read the value.

Client/Frontend/Overlay/Sheet/OverlaySheetView.swift
EdwardJTL: yea this one works. also interesting how this one works but writing to a PerferenceKey when reading the height doesn't work :/

Client/Extensions/ViewExtensions.swift
```diff
@@ -114,6 +114,30 @@ extension View {
...
+                GeometryReader { geom in
+                    Color.clear
+                        .useEffect(deps: geom.size.width) { width in
+                            updater(width)
+                        }
+                }
+            )
+    }
+
+    func onSizeOfViewChanged(perform updater: @escaping (CGSize) -> Void) -> some View {
```
evancrow, line 129: nice, thanks for adding these

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuView.swift
```diff
@@ -89,37 +89,54 @@ struct CheatsheetInfoView: View {
...
+        return (height < 0) ? 250 : height
+    }
+
     var body: some View {
         VStack(alignment: .center, spacing: 0) {
             VStack(alignment: .leading) {
                 header
                 text
             }
+            .onSizeOfViewChanged { size in
```
evancrow, line 113: nit: can remove the new line here and do `}.onSizeOfViewChanged { size in`

evancrow (approved):
LGTM just a minor nit

Issue #2930, opened 2022-02-23
release script failing


chassu:
I'm getting this error when running build-release.sh

```
*** Error: Error uploading '/tmp/Neeva.ipa'.
*** Error: CFBundleIdentifier Collision. There is more than one bundle with the CFBundleIdentifier value 'com.dailymotion.SDWebImage' under the iOS application 'Client.app'. With error code STATE_ERROR.VALIDATION_ERROR.90685 for id b4c4caf2-08de-485b-8d40-f4ebe8b6a0c6 Asset validation failed (-19208)
```

I'm guessing it's related to rebuilding Carthage after https://github.com/neevaco/neeva-ios/pull/2923. @j-f1 Do you have any idea?

j-f1:
This is probably because multiple things are using the `SDWebImage` package. Making sure it’s only embedded by one target (perhaps `Shared`?) should fix it

PR #2935, opened 2022-02-24
Fix Release Script

evancrow:
*possibly* Fixed: #2930


chung-lai (commented):
It's hard to read the project file. Could you please explain what was changed? Do we remove the SDWebImage framework from `Client` and only being linked in `Shared`?

evancrow:
I actually removed `SDWebImage` from `Shared` and kept it in `Client`. I guess `Client` is the only target using it because the build failed when I removed it from there, but not `Shared`.

chung-lai:
> I actually removed `SDWebImage` from `Shared` and kept it in `Client`. I guess `Client` is the only target using it because the build failed when I removed it from there, but not `Shared`.

That makes sense, thanks!

chung-lai (approved):

PR #2936, opened 2022-02-24
Pinned tabs/tab groups on top

chassu:
From tab groups plan doc:

"Pinned tabs are moved to the pinned section of the card grid, at the top of the card grid before any non-pinned tabs. Pinned tabs are ordered within the pinned section according to when they were pinned, with newly pinned tabs coming after already pinned tabs. (Users may re-order pinned tabs within the set of pinned tabs.)

Pinned tabs within a tab group are moved to the head of the tab group, and the tab group is moved into the pinned section of the card grid."

This PR handles the above except for reordering pinned tabs within the set of pinned tabs. After this PR and #2887  are merged, I will focus on making sure non of the non-pinned tabs come before pinned tabs due to the logic that promotes tabs to fill holes.


darinwf (approved):
This looks great, but I think you should add some test automation for exercising tab pinning in various scenarios. I'm okay if you want to take that as a separate PR since this is already a large PR, but that should be a pre-requisite to enabling the feature in production. Thanks!

Issue #2914, opened 2022-02-18
Show Popover When OpenSpace Fails


evancrow:
If `openSpace` fails for any reason (private, doesn't exist), we should replace the loading spinner with a popover saying an error occurred.

PR #2939, opened 2022-02-25
Add Error Popover Page

evancrow:
Fixed: #2914

Here's how it looks:

Now we have another issue of making the errors make sense...


evancrow:
Added an optional description field:

evancrow:
Discussed with @yozuysal, going to make this a Toast instead. This requires the user to tap something and read an error they won't understand or be helpful.

EdwardJTL:
could you update the screenshots on the PR when you get a chance?

evancrow:
Here's a recording @EdwardJTL:

https://user-images.githubusercontent.com/37356984/155815175-d00b8877-1b8d-4a1f-a2a2-c11fd62de2ad.mp4

Client.xcodeproj/project.pbxproj
```diff
@@ -7352,7 +7352,6 @@
 			buildSettings = {
 				CODE_SIGN_ENTITLEMENTS = Client/Entitlements/Neeva.entitlements;
 				DEFINES_MODULE = YES;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "";
```
EdwardJTL, line 7355: where did these changes come from? we should move them to a separate PR if they need to be checked in.

evancrow: I was testing running from an M- Mac, must have made it in unintentionall. Thanks for catching.

EdwardJTL (approved):

Issue #1014, opened 2021-07-17
Fully rebrand Private → Incognito


jed-fox-neeva:
- [ ] Make sure that no strings reference private browsing or private mode
- [ ] Update the names of everything in code to match

PR #2942, opened 2022-02-25
Finish Private to Incognito Rebrand

evancrow:
Fixed: #1014

LMK if I missed anything, there were some instances I couldn't change as they would have reset some user preferences.


chassu (approved):
LGTM

PR #2943, opened 2022-02-25
Resolve Warnings

evancrow:
Fixed a few warnings


Client/Frontend/CryptoWallet/Web3Model.swift
```diff
@@ -48,9 +48,8 @@ class Web3Model: ObservableObject, ResponseRelay {
 
     @Published var currentSequence: SequenceInfo? = nil {
         didSet {
-            guard let sequence = currentSequence, let wallet = wallet else { return }
```
yozuysal, line 51: I think the right fix here would be to do let _ = and keep the guard.

evancrow: Do you have a reason? Guard statements are meant for unwrapping optional to use, but if you just want to protect against a nil variable this is a cleaner way to do it.

EdwardJTL: we could do 
`guard currentSquence != nil, wallet != nil else { return } `
i think this just comes down to readability and preferences since we are not using the unwrapped values

evancrow: Works for me

yozuysal: let _ = is meant for unwrapping optionals to use in general, guard is meant for early returns that should warn against proceeding (compared to if , guard enforces that nothing should happen in that block unless the condition is true)

here if you use an if, 6 months later, someone can add in another if below it checking for something else. With a guard we are hinting that there shouldnt be anything put here if there is no wallet and currentSequence.

yozuysal (approved):

Issue #2944, opened 2022-02-25
Tab Switch Stuck on Space Detail View when Launching App with Link


EdwardJTL:
If the tab switcher is on space detail list when the app was last in foreground, opening a link with the app will then cause the tab switcher to be stuck on that page. See video.

https://user-images.githubusercontent.com/44557675/155807193-919b6255-0c98-4fe2-88cc-732b0f19e15b.MP4

PR #2945, opened 2022-02-25
Close Space When Opening Link

evancrow:
Fixed: #2944


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -935,20 +935,29 @@ class BrowserViewController: UIViewController, ModalPresenter {
...
-        if browserModel.showGrid {
-            browserModel.hideWithNoAnimation()
-        }
-
         if let introViewModel = introViewModel {
             introViewModel.dismiss {
                 self.introViewModel = nil
             }
         }
+
```
yozuysal: can we keep the order of these the same, just in case introViewModel related bits were after for a reason?

EdwardJTL:
tested and it works! ty for the quick fix 🔥

EdwardJTL (approved):

PR #2946, opened 2022-02-25
Add Space Entities to Core Spotlight

EdwardJTL:
## Description
Add routine to index entities (pages, urls, etc.) from spaces to core spotlight as well
- add new default variables for it
- add method to clean the index
- update scene delegate for CSSearchableItem's to support new flow

Add logging for indexing spaces to spotlight

## Notes
- `Still need to verify that thumbnail works for all sorts of PreviewEntity types~ done
- There is no fallback thumbnail for these items because everything is implemented in the shared module
- I recognize that the call sites for adding and removing index is not the most efficient; they are implemented in parallel to our current routine but scheduled on its own queue. I imagine those will get called a lot. Next task is to add logging, which will give us some insights on how costly these operations are and how much usage we get out of them.


Shared/API/SpaceStore+CoreSpotlight.swift
```diff
@@ -70,10 +119,133 @@ public extension SpaceStore {
...
+            with: imageURL,
+            options: options,
+            progress: nil,
+            completed: onCompletedPageFavicon
+        )
+    }
+
+}
+
+private extension SpaceEntityData {
```
yozuysal, line 195: Looks like some of these are copies from CardDetails? I wonder if there is a way to not having to maintain both these and the originals moving forward (as is, it will be easy to break things by modifying one file and not the other without knowing.)

yozuysal (commented):
yeah, I do wonder if space entities is really useful for spotlight... for members who use spaces, they will have public spaces they follow which makes sense to get the description and space title in spotlight... But they wont necessarily care about all space items. Maybe this is a design discussion topic?

Shared/API/SpaceStore+CoreSpotlight.swift
EdwardJTL: I believe these should just be declared on SpaceEntityData itself and used in `SpaceEntityThumbnail`, which is where this is loosely copied from. I am not sure if we want to include that change in this PR or make a separate one for it (which can go in first).

EdwardJTL:
> yeah, I do wonder if space entities is really useful for spotlight... for members who use spaces, they will have public spaces they follow which makes sense to get the description and space title in spotlight... But they wont necessarily care about all space items. Maybe this is a design discussion topic?

I think adding these into spotlight will increase the visibility of our app to users, especially with spaces in which content may be added by others (such as shared spaces and news alert). I am not seeing any downside to adding items to spotlight that may never match a user search. I added the item to the design sync minutes.

Shared/API/SpaceStore+CoreSpotlight.swift
```diff
@@ -70,10 +119,133 @@ public extension SpaceStore {
...
+    ) {
+        let manager = SDWebImageManager.shared
+        let options: SDWebImageOptions = .lowPriority
+
+        let onCompletedPageFavicon: SDInternalCompletionBlock = {
+            (img, _, _, _, _, _) -> Void in
+            completionHandler(img)
+        }
+
+        manager.loadImage(
```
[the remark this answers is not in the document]
EdwardJTL: I added a resizer. However, most of the spaces entities i tested already have the thumbnail data parsed into a string, in which case the SDWebImage is not used. they can go up to very large sizes (2K x 1K) though.

EdwardJTL:
> yeah, I do wonder if space entities is really useful for spotlight... for members who use spaces, they will have public spaces they follow which makes sense to get the description and space title in spotlight... But they wont necessarily care about all space items. Maybe this is a design discussion topic?

We discussed this feature in the mobile growth meeting. We feel that we should add this feature.

macy-n (approved):

PR #2954, opened 2022-02-28
Move most of wallet code to its own target as a framework

yozuysal:
Moves all the files over and rejiggers dependencies to work correctly
Adds a few new interfaces to pull some of the classes/structs partially

This is to be able to use most of the code in an onboarding appclip later


chung-lai (commented):
We went over the project related changes offline and LGTM.

Let me know if there is particular part of changes that you want me to go over given most of the code changes here are code extraction.

yozuysal:
Yeah, not much change beyond the moving things around and making things public with public initializers. I added a OnboardingModel, pulling some logic from Web3Model, but that is also just copying code over.

yozuysal:
That said, maybe it is better to land #2948 first and land this after the branch is cut. This way, the branch will have all the updated UI and I can work on any weirdness this causes only in main

chung-lai:
> That said, maybe it is better to land #2948 first and land this after the branch is cut. This way, the branch will have all the updated UI and I can work on any weirdness this causes only in main

Yes and also the build issue that we are dealing with. Would love to see this change land after branch

chung-lai (approved):

PR #2967, opened 2022-03-02
support drag-and-drop in tab switcher

chassu:
Poking around right now...

https://user-images.githubusercontent.com/87332917/156290759-8296ce76-4496-40fd-abc4-ab11a491f749.MP4

I have a few questions for now:

1. Why is the `performDrop` needed in TabCardDetails? I haven't understood why this part is needed

```
                    DispatchQueue.main.async {
                        self.manager.get(for: self.id)?.loadRequest(URLRequest(url: url))
                    }
```

2. What do the strings in the argument `.onDrop(of: ["public.url", "public.text"], delegate: details)` represent? Does it work with any custom strings we give as long as we are consistent on the sender and receiver?

3. I have to tweak TabCardDetails to let it have access to CardModels so that it can do the swapping on AllDetails when drag and drop event happens. This is making me thinking that it might be actually important (or better) to have TabCardDetails access CardModels instead of TabManager.

4. I also see another problem - Are we allowing drag-and-drop to happen between child tabs and individual tabs? What are we expecting to see when a user drags a child tab outside of a tab group? Are we implicitly making 'remove out of group' happening as an intermediate step? Does the other tab that gets swapped become part of the tab group after?

### DIFF
--- a/Client/Frontend/Browser/Card/Card.swift
+++ b/Client/Frontend/Browser/Card/Card.swift
@@ -161,6 +161,7 @@ struct FittedCard<Details>: View where Details: CardDetails {
 struct Card<Details>: View where Details: CardDetails {
     @ObservedObject var details: Details
     var dragToClose = false
+    var dragNDrop = FeatureFlag[.dragAndDropTabs]
     /// Whether — if this card is selected — the blue border should be drawn
     var showsSelection = true
     var animate = false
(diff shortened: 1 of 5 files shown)


chassu:
the thumbnail doesn't persist when dragging, not sure why :/

yozuysal:
Just to quickly answer the questions about old implementation without looking at the review, if we are implementing drag and drop, feel free to remove any onDrop and performDrop impls existing right now. They were for the first version of CardStrip which is not used anymore. So we can get a new design if they are causing complexity.

darinwf:
@chassu - nice work getting this going! think we have some 1:1 time tomorrow... let's use that to walk through the approach here.

chassu:
Inserting actually doesn't look that bad simply with the preview

https://user-images.githubusercontent.com/87332917/156475969-60f85f58-2dea-4940-b04f-525594c476d4.mp4

chassu:
Made some fix to make preview work on iOS 15 only

https://user-images.githubusercontent.com/87332917/157976437-aca97927-ac99-4c85-bd47-d1c513514113.mp4

PR #2968, opened 2022-03-02
TabManager Cleanup

evancrow:
- Separated different TabManager sections into extensions so it's easier to read
- Remove duplicate functions
- Renamed some variables

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


Client/Frontend/Browser/Tabs/TabManager/TabManager+HelperFunctions.swift
```diff
@@ -0,0 +1,32 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Shared
+import Storage
+
+// Helper functions for test cases
+extension TabManager {
```
darinwf, line 9: nit: actually this file can be called `TabManager+TestSupport.swift` since that's all these functions are used for.

darinwf (approved):

PR #2972, opened 2022-03-03
Add Additional Logging to Cheatsheet (Funnel, Fallback)

EdwardJTL:
## Description

Add logging needed to support the data analysis we want to do

- Add logging to when cheatsheet content is shown
- Add logging to when the info view is shown
- Move logging empty view to listen to network request status to avoid false positive from SwiftUI
- Add logging to capture which query source is used

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually tested by observing proxyman snoops and triggering flows in simulator.
I have a url that will give an empty cheatsheet; macy also has it.


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -112,4 +80,43 @@ public class CheatsheetMenuViewModel: ObservableObject {
             .assign(to: \.searchRichResultsError, on: self)
             .store(in: &subscriptions)
     }
+
+    class func isCheatsheetEmpty(
```
EdwardJTL, line 84: this is temporary. i will revisit this once we reorganize the cheatsheet code from `Tab` class.

macy-n (approved):

PR #2977, opened 2022-03-03
isolate web3 model showing wallet details

chassu:
This is a smaller change in attempt to prevent CardGrid() from getting drawn caused by Web3Model when a selectTab event happens.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
This shouldn't affect the wallet view at all and I've tested the wallet app runs successfully.


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -54,6 +54,7 @@ class BrowserViewController: UIViewController, ModalPresenter {
     private(set) lazy var web3Model: Web3Model = {
         return Web3Model(server: self.server, presenter: self, tabManager: self.tabManager)
     }()
```
yozuysal, line 56: I would actually make web3Model own this just to keep things tidy.

Construct it during init() for web3Model and then in the above do bvc.web3Model. walletDetailsModel while injecting the environmentObject

yozuysal (approved):

yozuysal:
lgtm, but note that the test failure is a real one!

PR #2980, opened 2022-03-04
Fix incorrectly set version fields

darinwf:
Adds tool to read version. When all of the version fields are not set
to the same value, our scripts will break.


Scripts/get-version.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
chung-lai, line 1: How does this script get used?

darinwf: I just added it as a convenience if you wanted to check the current version. Before I had been just running update-version.sh and aborting it, but that's a bit clunky.

chung-lai (approved):

PR #2982, opened 2022-03-04
Test Fixes

evancrow:
Fix issues with test after transitioning to the Overflow Menu from the Neeva Menu

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


evancrow:
Changed how the text was entered on the simulator since Paste & Go stopped working too. Now we just set the text of the TextField directly, a lot faster now though.

evancrow:
Paste & Go seems to be fine on real devices though

darinwf (approved):
LG, and thanks for working on this, but what's the issue w/ iPad? Why do those tests need to be disabled? Is there a path for re-enabling them?

evancrow:
I disabled tests for a couple reasons:

- `UIPasteboardExtensionsTests` and `URLEntry` were disabled for being flakey on CircleCI. I created issues for those if someone wants to spend time resolving the flakiness
- KeyboardShortcutTests were disabled on iPhone because iPhone doesn't support KeyboardShortcuts and it doesn't make sense to run them there

PR #2986, opened 2022-03-04
make pinned tabs and non-pinned tabs stay on separate rows

chassu:
Make sure that no pinned tabs and non-pinned tabs stay on the same row. Doing this as a starter to break tabs into sections and help support drag-n-drop for each sections.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually tested for now

Before:

After:


chassu:
I'm trying to add XCUI tests for this PR... but I haven't figured out a way to access a CardDetail in the CardGrid. I'm imagining if I can access CardDetails I can do something like:
```
CardDetail.press(forDuration: 1)
waitForExistence(app.buttons["Pin Tabs"], timeout: 3)
```

Also, is it possible to test a feature that is behind a flag?

darinwf:
Hmm... but this doesn't seem very ideal for the interim period before we have Ria's new design w/ sections. I don't quite follow why we would make this change.

chassu:
This also makes sure that none of the non-pinned tabs come before pinned tabs. With the logic from `PromoteCellsAfterIndex `, there is a chance that some non-pinned tabs come for a pinned tab. 

For example:
[pinned tab]
[pinned child tab, pinned child tab]
[non-pinned tab]

With the hole-filling logic it currently converts it to:
[pinned tab, non-pinned tab]
[pinned child tab, pinned child tab]

chassu:
From the tab groups doc: "Pinned tabs are moved to the pinned section of the card grid, at the top of the card grid before any non-pinned tabs."

chassu:
Also I think we've discussed this before the new pinned section UX was out, and you pointed out that we shouldn't want a non-pinned tab to come before a pinned tab.

darinwf:
Oh, 100%... pinned tabs should always be listed before non-pinned tabs. Agree we should fix that issue.

I just don't think we should add an unnecessary line-break between pinned and non-pinned tabs until we introduce the section concept.

chassu:
Here's an example of a corner case that I talked about earlier where a line break will still happen.

https://user-images.githubusercontent.com/87332917/157133272-6b2f4373-b677-413d-a73b-597d2ee54be4.mp4

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -150,12 +149,18 @@ class TabCardModel: CardModel {
 
         // This functions performs lookahead and checks if there are tab/tab groups that can be inserted
         // in the current row before a new row gets inserted.
-        func PromoteCellsAfterIndex(_ row: inout Row, _ index: Int) {
+        func PromoteCellsAfterIndex(currDetail: TabCardDetails, _ row: inout Row, _ index: Int) {
```
darinwf, line 152: nit: I'd probably change the second and third parameters to be named (i.e., drop the `_`). Once you have a bunch of parameters, you start to really benefit from named parameters. The function name doesn't alone make it very clear what the parameters should be so probably wise to let them be named.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -150,12 +149,18 @@ class TabCardModel: CardModel {
...
             var didPromote = false
             var id = index
 
             while id < allDetailsFiltered.count && row.numTabsInRow < maxCols {
                 let details = allDetailsFiltered[id]
+
+                // don't promote non-pinned tabs if we're still in pinned section
+                if let tabGroup = tabGroupModel.allDetails.first(where: { $0.id == currDetail.rootID }), tabGroup.isPinned && !details.isPinned
```
darinwf, line 160: nit: looks like you should run swift-format.sh

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -150,12 +149,18 @@ class TabCardModel: CardModel {
...
+                { return }
+                if currDetail.isPinned && !details.isPinned { return }
```
darinwf, line 162: nit: good practice to put return statements on a new line so it is easy to stick a breakpoint on a return statement :)

darinwf (commented):
This would be a very good scenario to add a unit test for.

chassu:
The unit test you're looking for is probably already in #3015. But that test should only pass after this PR is merged.

darinwf (approved):

PR #2990, opened 2022-03-04
Fix Cheatsheet Loading Animation

### LINKED ISSUE #2959: The NeevaScope animation often does not run for me
Tested using 1.29.3

I rarely see it. Almost never to be precise. It made me wonder if there might be a delay on the animation and that the network response is just coming in fast enough that the animation doesn't have a chance to run. But @EdwardJTL tells me that is probably not the case and that this is likely due to some other issue.

EdwardJTL:
I found out that in some cases, `FlowTimeline.playing` can be true when read in `viewDidAppear`. I am not sure why. But we can fix this by asking the timeline to reset first.

Also fixed a typo in the timeline constructor; the animation should look closer to the original figma design now.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested repeatedly on device. @darinwf if you could try it on your device too it would be great.

## Issues addressed
closes #2959


darinwf (approved):
Nice!

darinwf:
Maybe the View is not getting torn down when hidden. We probably still have the OverlayView in the View hierarchy, etc.

Issue #2964, opened 2022-03-01
TabCardModel should always be consistent with TabManager [was: TabCardDetails should not lookup Tab by ID]


darinwf:
See https://github.com/neevaco/neeva-ios/issues/2951 for why this is not workable. We need to move away from CardDetails having a manager that we use to query the Tab by ID. The concrete subclasses of CardDetails should just have a reference to the underlying model they represent (e.g., Tab or Space).

FYI @yozuysal

darinwf:
New approach:

Instead of such a radical change, we can just simply make `TabCardModel.allDetails` not be `@Published` and call `objectWillChange.send()` after `allDetails` has been modified. This way `TabCardModel` and `TabManager` will always be in sync when we are updating the UI. Therefore `buildRows` should not need to worry about a `TabCardDetails` having a nil associated `Tab`.

PR #2991, opened 2022-03-05
Ensure allDetails is always consistent with TabManager.tabs

darinwf:
We do this by making `allDetails` not be `@Published`. That way
we can control when the `TabCardModel` signals any listeners. An
explicit `objectWillChange.send()` at the end of `onDataUpdated`
achieves this.

Also updates `TabGroupCardModel` in a similar fashion and cleans
up some code.

Reverts bandaid fix to handle a nil Tab from a `TabCardDetails`,
so we can see if this fix didn't work.

Fixes: #2964

### How was this tested?
- [x] I tested this manually


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -706,68 +697,44 @@ class SiteCardModel: CardModel {
...
-        allDetails = manager.getAll()
-            .map {
-                TabGroupCardDetails(
-                    tabGroup: $0,
-                    tabGroupManager: manager)
-            }
-        objectWillChange.send()
-    }
-
-    func setupDetailsListener() {
```
yozuysal, line 760: I cant remember why we ever needed the async post here... Good riddance! But I wonder whether we had band-aided a bug with this...

yozuysal (approved):

Client/Frontend/Browser/Card/CardModels.swift
darinwf: I actually preserved it with the `.receive(on: DispatchQueue.main)` above. I don't know why it is needed either.

PR #2992, opened 2022-03-05
Refactor CheatsheetMenuViewModel and cheatsheet data fetching

### LINKED ISSUE #2971: Optimize Cheatsheet Model Logic
Opening the cheatsheet once creates multiple `CheatsheetMenuViewModel`, and they seem to never get de-initialized.
Ideally we should only need at most one per tab (or maybe only one globally shared amongst all tabs)

This ties back to a conversation we had before about cleaning up cheatsheet fetching logic and removing them from `Tab` class @darinwf

EdwardJTL:
- Remove cheatsheet code from `Tab` class
- Relocate loading logic to `CheatsheetMenuViewModel`

Details on implementation:
- each `Tab` now holds a lazily created menu view model property, which is reused to present the cheatsheet
- this prevents us from creating multiple objects each time we open the sheet, which is something I observed prior to this change
- we still refresh the content each time the sheet is opened. we can choose to disable this and only load on first open
- `CheatsheetMenuViewModel` only has one published property now

Future work:
- This paves the way to using only one view model in the entire app life cycle. This is not done right now because it would involve handling switching query source mid request.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
I opened cheatsheets on tabs and verified they work. Also checked memory graph to make sure there is a correct number of them.

## Issues addressed
closes #2971


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -47,76 +82,139 @@ public class CheatsheetMenuViewModel: ObservableObject {
...
-        .store(in: &subscriptions)
+    // MARK: - Load Methods
+    func reload() {
+        fetchCheatsheetInfo()
+    }
 
-        tabManager.selectedTab?.$cheatsheetData.assign(to: \.cheatsheetInfo, on: self).store(
-            in: &subscriptions)
+    func fetchCheatsheetInfo() {
+        guard !cheatsheetDataLoading else { return }
```
EdwardJTL, line 98: this logic is new

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -47,76 +82,139 @@ public class CheatsheetMenuViewModel: ObservableObject {
...
+        setupCheatsheetLoaderLogger()
 
-        tabManager.selectedTab?.$url.assign(to: \.currentPageURL, on: self).store(
-            in: &subscriptions)
+        clearCheatsheetData()
 
-        tabManager.selectedTab?.$cheatsheetDataError
-            .assign(to: \.cheatsheetDataError, on: self)
-            .store(in: &subscriptions)
+        currentPageURL = tab?.webView?.url
```
EdwardJTL, line 104: we now only update the url when we start a loading request, instead of sinking on the value repeatedly

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -47,76 +82,139 @@ public class CheatsheetMenuViewModel: ObservableObject {
...
-            if let recipe = cheatsheetInfo.recipe {
-                // recipeView
-                if let ingredients = recipe.ingredients,
-                    let instructions = recipe.instructions,
-                    ingredients.count > 0,
-                    instructions.count > 0
+        CheatsheetQueryController.getCheatsheetInfo(url: url.absoluteString) { [self] result in
+            switch result {
+            case .success(let cheatsheetInfo):
+                self.cheatsheetInfo = cheatsheetInfo.first
```
EdwardJTL, line 118: using `.first` instead of `[0]`

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetOverlayContent.swift
```diff
@@ -14,7 +14,7 @@ struct CheatsheetOverlayContent: View {
 
     init(menuAction: @escaping (NeevaMenuAction) -> Void, tabManager: TabManager) {
         self.menuAction = menuAction
-        self.model = CheatsheetMenuViewModel(tabManager: tabManager)
+        self.model = tabManager.selectedTab?.cheatsheetModel ?? CheatsheetMenuViewModel(tab: nil)
```
EdwardJTL, line 17: if `selectedTab` is `nil`, then we will have a cheatsheet that always shows no data. The behaviour shouldn't be any different though; in the old code, the constructor puts subscriptions on the selectedTab's published properties, which would be nil if the tab is nil anyways.

EdwardJTL (commented):
Here are all the changes in the logic

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -97,13 +97,9 @@ class Tab: NSObject, ObservableObject {
...
     // MARK: - Cheatsheet Properties
     /// Cheatsheet info for current url
-    @Published private(set) var cheatsheetData: CheatsheetQueryController.CheatsheetInfo?
-    @Published private(set) var searchRichResults: [SearchController.RichResult]?
-    @Published private(set) var cheatsheetDataLoading: Bool = false
-    @Published private(set) var cheatsheetDataError: Error?
-    @Published private(set) var searchRichResultsError: Error?
-    @Published private(set) var currentCheatsheetQuery: String?
-    private var cheatsheetLoggerSubscription: AnyCancellable?
+    lazy private(set) var cheatsheetModel: CheatsheetMenuViewModel = {
```
darinwf, line 100: Note: Eventually it would be ideal for Tab to not depend on anything cheatsheet related including this.

darinwf (approved):
Thanks for cleaning up the code!

PR #2993, opened 2022-03-05
add toast after a tab is pinned/unpinned

chassu:
Adding a toast to indicate that the user has successfully pinned/unpinned a tab.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Pinned and unpinned tab manually and see if the correct toast show up accordingly.

https://user-images.githubusercontent.com/87332917/156866962-d26a5f23-d71c-42de-9b22-eed3b9b4a2b6.mp4


darinwf:
Is this the suggested text from @Ria-yu ?

chassu:
Hmm, I couldn't find relevant texts in figma and paper doc for now. I temporarily put some texts there. Can comfirm with @Ria-yu once she has time to look at it.

chassu:
https://user-images.githubusercontent.com/87332917/157520194-4befd72e-e470-4d91-b8bd-23ddb707ca04.mp4

Client/Frontend/Toast/ToastDefaults.swift
```diff
@@ -52,6 +52,36 @@ class ToastDefaults: NSObject {
         }
     }
 
+    func showToastForPinningTabs(tabTitle: String, pinning: Bool, tabManager: TabManager) {
+        toastProgressViewModel = ToastProgressViewModel()
```
evancrow, line 56: This should only be used for asynchronous tasks, like downloading a file. I would suggest just passing the tab pinned string to `showToast(with text: String, toastViewManager: ToastViewManager)`. Will also reduce duplicate code.

Client/Frontend/Toast/ToastDefaults.swift
```diff
@@ -52,6 +52,36 @@ class ToastDefaults: NSObject {
...
+        toastProgressViewModel?.status = .success
+
+        var toastText = LocalizedStringKey("")
+
+        if pinning {
```
evancrow, line 61: This can be simplified to `let toastText =  pinning ? "Tab Pinned" : "Tab unpinned"`

Client/Frontend/Toast/ToastDefaults.swift
```diff
@@ -52,6 +52,27 @@ class ToastDefaults: NSObject {
         }
     }
 
+    func showToastForPinningTabs(tabTitle: String, pinning: Bool, tabManager: TabManager) {
+        let toastText = pinning ? LocalizedStringKey("Tab Pinned") : LocalizedStringKey("Tab Unpinned")
+
+        let toastContent = ToastViewContent(
```
evancrow, line 58: Remove this (see next comment)

Client/Frontend/Toast/ToastDefaults.swift
```diff
@@ -52,6 +52,27 @@ class ToastDefaults: NSObject {
...
+            ))
+
+        guard
+            let toastViewManager = SceneDelegate.getCurrentSceneDelegate(
+                with: tabManager.scene)?.toastViewManager
+        else {
+            return
+        }
+
+        let toastView = toastViewManager.makeToast(text: toastText, checkmark: true)
```
evancrow, line 71: Instead of creating a new toast here, pass everything too `showToast()`

Client/Frontend/Toast/ToastView.swift
```diff
@@ -77,6 +81,17 @@ struct ToastView: View {
...
+                            Circle()
+                                .foregroundColor(.white)
+
+                            Symbol(decorative: .checkmark, size: 10)
+                                    .foregroundColor(Color(ToastViewUX.ToastDefaultColor))
+                        }
+                        .frame(width: 18, height: 18)
+                    }
+
                     if let toastProgressViewModel = toastProgressViewModel {
```
evancrow, line 95: Should make this an `else if`, so we don't show two checkmarks

Client/Frontend/Toast/ToastDefaults.swift
```diff
@@ -53,13 +53,7 @@ class ToastDefaults: NSObject {
     }
 
     func showToastForPinningTabs(tabTitle: String, pinning: Bool, tabManager: TabManager) {
```
evancrow, line 55: Two more final nits: 

1. This function would make more sense with `Tab` not `Tabs` plural
2. Doesn't look like tabTitle is being used, can remove that

evancrow (approved):
LGTM!

PR #2996, opened 2022-03-05
Bug fixes to Wallet

### LINKED ISSUE #2953: Crash on startup w/ wallet enabled
Saw this startup crash last week when I had zero-bars on my iPhone:

```
Thread 0 name:
Thread 0 Crashed:
0   libswiftCore.dylib            	0x0000000185033514 _assertionFailure(_:_:file:line:flags:) + 304 (AssertCommon.swift:132)
1   libswiftCore.dylib            	0x0000000185099ea4 swift_unexpectedError + 560 (ErrorType.swift:188)
2   Client                        	0x00000001010fea0c specialized WalletAccessor.init() + 1104 (WalletAccessor.swift:31)
3   Client                        	0x000000010119bda8 init + 8 (Web3Model.swift:0)
4   Client                        	0x000000010119bda8 specialized Web3Model.init(server:presenter:tabManager:) + 2468
5   Client                        	0x0000000101180480 init + 20 (<compiler-generated>:122)
6   Client                        	0x0000000101180480 init + 20 (Web3Model.swift:0)
7   Client                        	0x0000000101180480 init + 20 (<compiler-generated>:122)
8   Client                        	0x0000000101180480 init + 20 (BrowserViewController.swift:55)
9   Client                        	0x0000000101180480 BrowserViewController.web3Model.getter + 260
10  Client                        	0x00000001012476d4 BrowserViewController.configureWalletServer() + 140 (BrowserViewController+ServerDelegate.swift:39)
11  Client                        	0x000000010119c588 specialized BrowserViewController.init(profile:scene:) + 1452 (BrowserViewController.swift:196)
12  Client                        	0x0000000100e1df84 init + 16 (BrowserViewController.swift:186)
13  Client                        	0x0000000100e1df84 init + 16 (SceneDelegate.swift:0)
14  Client                        	0x0000000100e1df84 SceneDelegate.setupRootViewController(_:) + 240
15  Client                        	0x0000000100e21f38 specialized SceneDelegate.scene(_:willConnectTo:options:) + 528 (SceneDelegate.swift:38)
16  Client                        	0x0000000100e1de64 scene + 16 (<compiler-generated>:0)
17  Client                        	0x0000000100e1de64 @objc SceneDelegate.scene(_:willConnectTo:options:) + 96
18  UIKitCore                     	0x0000000182db0a80 +[UIScene _sceneForFBSScene:create:withSession:connectionOptions:] + 1348 (UIScene.m:1261)
19  UIKitCore                     	0x00000001830b2264 -[UIApplication _connectUISceneFromFBSScene:transitionContext:] + 1272 (UIApplication.m:3900)
20  UIKitCore                     	0x0000000182db3260 -[UIApplication workspace:didCreateScene:withTransitionContext:completion:] + 332 (UIApplication.m:3933)
21  UIKitCore                     	0x0000000182cf2218 -[UIApplicationSceneClientAgent scene:didInitializeWithEvent:completion:] + 388 (UIApplicationSceneClientAgent.m:45)
22  FrontBoardServices            	0x00000001922a18d4 -[FBSScene _callOutQueue_agent_didCreateWithTransitionContext:completion:] + 440 (FBSScene.m:439)
23  FrontBoardServices            	0x00000001922c7824 __94-[FBSWorkspaceScenesClient createWithSceneID:groupID:parameters:transitionContext:completion:]_block_invoke.215 + 128 (FBSWorkspaceScenesClient.m:384)
24  FrontBoardServices            	0x000000019228206c -[FBSWorkspace _calloutQueue_executeCalloutFromSource:withBlock:] + 240 (FBSWorkspace.m:352)
25  FrontBoardServices            	0x00000001922836ac __94-[FBSWorkspaceScenesClient createWithSceneID:groupID:parameters:transitionContext:completion:]_block_invoke + 372 (FBSWorkspaceScenesClient.m:383)
26  libdispatch.dylib             	0x00000001802bb670 _dispatch_client_callout + 20 (object.m:560)
27  libdispatch.dylib             	0x00000001802bf128 _dispatch_block_invoke_direct + 264 (queue.c:489)
28  FrontBoardServices            	0x00000001922839c4 __FBSSERIALQUEUE_IS_CALLING_OUT_TO_A_BLOCK__ + 48 (FBSSerialQueue.m:157)
29  FrontBoardServices            	0x0000000192282d8c -[FBSSerialQueue _targetQueue_performNextIfPossible] + 220 (FBSSerialQueue.m:181)
30  FrontBoardServices            	0x0000000192287414 -[FBSSerialQueue _performNextFromRunLoopSource] + 28 (FBSSerialQueue.m:194)
31  CoreFoundation                	0x000000018066b0d0 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 28 (CFRunLoop.c:1972)
32  CoreFoundation                	0x000000018067bd90 __CFRunLoopDoSource0 + 208 (CFRunLoop.c:2016)
33  CoreFoundation                	0x00000001805b6104 __CFRunLoopDoSources0 + 376 (CFRunLoop.c:2061)
34  CoreFoundation                	0x00000001805bb8a4 __CFRunLoopRun + 820 (CFRunLoop.c:2951)
35  CoreFoundation                	0x00000001805cf468 CFRunLoopRunSpecific + 600 (CFRunLoop.c:3268)
36  GraphicsServices              	0x000000019c15a38c GSEventRunModal + 164 (GSEvent.c:2200)
37  UIKitCore                     	0x0000000182f71088 -[UIApplication _run] + 1100 (UIApplication.m:3493)
38  UIKitCore                     	0x0000000182cef958 UIApplicationMain + 2092 (UIApplication.m:5046)
39  Client                        	0x0000000100deccf0 main + 260 (main.swift:19)
40  dyld                          	0x0000000101a29aa4 start + 520 (dyldMain.cpp:879)
```

yozuysal:
- Avoid crashes when there is no network connection, and make sure to recreate web3 when we have network
- Align the url being seen through the subscription calling updateSession and what we use to compute trustSignal
- The above fixes an issue with navigation to a verified site from malicious site showing a malicious site view again
- Avoid calling updateSession if we dont have a wallet, this avoid setting the session to nil continously for non-wallet users
- Fix remove wallet not showing up confirmation

### How was this tested?
- [X] I tested this manually
- [ ] I added automated tests

### Manual test cases
Go to clayfriends.app
Observe the malicious site view. Click "Navigate to clayfriends.io"
Make sure no extra malicious site views pop up

Disconnect network and restart app (This should create a nil web3 object)
Reconnect network and go to a dApp and make sure it connects correctly

https://user-images.githubusercontent.com/83613310/156872482-2c3fb107-ffb4-4c94-94f4-3173c6c7be31.mp4

## Issues addressed
Fixes #2953


Client/Frontend/NeevaWallet/Web3Model.swift
```diff
@@ -126,8 +126,8 @@ class Web3Model: ObservableObject, ResponseRelay {
 
             self.selectedTab = tab
             self.updateCurrentSession()
-            self.urlSubscription = tab.$url.sink { _ in
-                self.updateCurrentSession()
+            self.urlSubscription = tab.webView?.url.publisher.sink { url in
```
darinwf, line 129: Note: this will generate an empty (or maybe nil?) `URL` value if the renderer process is killed. Are you sure that's what you want? An alternative that might be better is to listen to `tab.$url` instead.

WalletCore/WalletAccessor.swift
```diff
@@ -152,23 +159,27 @@ public struct WalletAccessor {
...
+            web3.eth.estimateGasPromise(tx, transactionOptions: opts)
                 .done(on: DispatchQueue.main) { estimate in
                     completion(gasPrice, estimate)
                 }.cauterize()
         }
     }
 
     public func send(on chain: EthNode, transactionData: TransactionData) throws -> String {
-        let contract = web3(on: chain).contract(
+        guard let web3 =  web3(on: chain) else { return "" }
```
darinwf, line 180: Hmm, would it make sense to throw an error instead? Silently dropping the `send(..)` call on the floor might lead to other bugs / flakiness down the line? Maybe we will have a harder time understanding why `send(..)` sometimes does nothing if there is some future race condition with `web3` being good to go? Dunno...

Client/Frontend/NeevaWallet/Web3Model.swift
yozuysal: the bug was caused by tab.$url not being in sync with selectedTab?.url actually. We depend on selectedTab.url for trustedSignal and that lags tab.$url being updated by a bit at times. Which means, we sometimes get the url updating to the new verified page, but the trustSignal is still staying unTrusted.

Being nil here wouldnt cause an undefined state even if we stay there for a while

WalletCore/WalletAccessor.swift
yozuysal: Hmm, so the context is, we should never be null here, and a web3 should be assigned by the time we are here. Maybe I should log the error instead? I dont think there is anything useful to show the user at this point though.

Client/Frontend/NeevaWallet/Web3Model.swift
darinwf: Hmm... I'm a bit confused here. Do you mean selectedTab.$url and selectedTab.url? Is this about how modifications to a `@Published` property generate a `objectWillChange` before the new value is set? Do we need to add in an asynchronous delay?

e.g.,
```
self.urlSubscription = tab.$url.receive(on: DispatchQueue.main).sink { _ in
 ...
}
```

Or alternatively instead of ignoring the passed value of URL, use that and avoid checking `selectedTab?.url` within `updateCurrentSession()`.

darinwf: I think it's really important that we move away from directly monitoring / accessing properties of the WebView. Tab should be our abstraction for that. There are many quirks with WebView that we need to handle (e.g., think about how session restore works).

yozuysal: yeah it was about the objectWillChange issue again. I am passing the url here to updateCurrentSession, but trustSignal is currently a computed variable... Maybe I should change that to be updated with session actually... I will also test out receive(on:). Either way, sg re avoding webview.$url. Looks like we have enough options to try out here..

darinwf: Cool, there's one annoying issue still with `Tab.url` in that it is nil until the Tab is restored. I want to fix that. Hopefully it won't cause a problem for wallet.

yozuysal: made trustSignal and alternate Domain update based on the same url as well. This nicely aligns all used urls for verified signals. I will later pull all this logic out to its own model in a followup PR... PTAL!

Client/Frontend/NeevaWallet/Web3Model.swift
```diff
@@ -159,11 +166,23 @@ class Web3Model: ObservableObject, ResponseRelay {
     }
 
     func updateCurrentSession() {
+        updateCurrentSession(with: selectedTab?.url)
+    }
+
+    func updateCurrentSession(with url: URL?) {
```
darinwf, line 172: nit: should this variant be `private`?

darinwf (approved):

Issue #2526, opened 2022-01-07
Add attribute to neeva.com URLs from "Neeva for Safari" extension


darinwf:
Add attribute to neeva.com URLs from "Neeva for Safari" extension

We can vary the `src=nvobar` parameter to help identify traffic from this extension. This parameter is added when Neeva is the default search engine and the user enters a query in the URL bar on other devices. We can use the same parameter, and then that will enable us to track server side how often Safari users are benefitting from the "Neeva for Safari" extension.

@evancrow FYI

PR #2998, opened 2022-03-06
Add extension-specific query parameter (#2526)

jonathangarelick:
_This PR adds a query parameter to address bar queries that are handled by the "Neeva for Safari" extension. This allows clearer tracking of extension usage._

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
I manually tested this with a couple of searches from the address bar. This functionality is not very complex, so I don't feel it justifies extensive testing.

https://user-images.githubusercontent.com/41495718/156910008-348c11a1-4c39-43c5-9bc1-266ea162b128.mov

## Issues addressed
* https://github.com/neevaco/neeva-ios/issues/2526


NeevaForSafari/NeevaForSafariExtension/Resources/content.js
```diff
@@ -7,7 +7,7 @@ browser.runtime.sendMessage({ "getPreference": "neevaRedirect"}).then((response)
     referrerURL = window.document.referrer;
 
     if (searchQuery != null && response["value"] && !possibleReferalURLs.includes(referrerURL)) {
-        const url = `https://neeva.com/search?q=${searchQuery}`;
+        const url = `https://neeva.com/search?q=${searchQuery}&src=ios_ext`;
```
darinwf, line 10: Thanks so much for the PR! I believe we should just use `src=nvobar` instead as described in the issue. That way we simulate how a browser, w/ Neeva set as default search engine through opensearch, would communicate to neeva.com.

jonathangarelick: Got it, I must have misinterpreted the issue. I believe this should be addressed now. Let me know if you need anything else!

darinwf (approved):
LGTM, thanks for the patch!

PR #3000, opened 2022-03-06
Add optional client-side logging of httpd~login cookie

### LINKED ISSUE #2999: Some users are reporting having to sign-in frequently
This is a bug for tracking the investigation of this issue.

darinwf:
Appends to log file once per request to neeva.com. This is intended
to help us isolate and better understand how some users are having
to login repeatedly.

Fixes #2999

### How was this tested?
- [x] I tested this manually

### Manual test cases
Observed log entries as expected when loading a neeva.com URL.


darinwf:
I decided to add a new "auth" logger category so we can have a log file dedicated to authentication related logging. It might make sense to move some other login related logging into this logger, but I'll save that for another time. For now, I just wanted to make it so that when we ask vron to enable this that the result is a very focused log file.

macy-n (approved):

Issue #2483, opened 2021-12-21
ShareTo extension's "load in background" feature does nothing


darinwf:
ShareTo extension's "Load in Background" feature does nothing

From a code inspection, it looks like it might try to write the tabs to the profile with the idea that the app will pick them up when next launched. But that doesn't seem to be working.

I stumbled upon this as I noticed the `loadQueuedTabs` function in `BrowserViewController` has no callers. I suspect if we look at the old Firefox code, we'd see that it was called and that this is the function used to support the "Load in Background" feature.

PR #3004, opened 2022-03-06
Correct "load in background" functionality

jonathangarelick:
This PR corrects the "load in background" functionality inherited from Firefox iOS.

Meta-note: Apologies if this is a bit messy, I'm new to Swift. Happy to make any adjustments as needed. Also, if you'd prefer this functionality be hidden (as mentioned in the issue), I can take care of that as well.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
See video below.

https://user-images.githubusercontent.com/41495718/156946416-795ff72e-fd9a-4dcc-8e50-ea859eecc8b3.mov

## Issues addressed
* https://github.com/neevaco/neeva-ios/issues/2483


jonathangarelick:
Here's an example with multiple tabs queued in the background. I'm not sure if the lazy loading is intentional or relevant, but you can see it at the end.

https://user-images.githubusercontent.com/41495718/156946782-e2727b01-d25d-4f97-9b91-864626c1cf21.mov

jonathangarelick:
After doing some more research, my code is a bit wrong. The `loadQueuedTabs` function doesn't seem to do much -- the real action takes place in `dequeueQueuedTabs`. I'll have to spend some more time digging into this.

jonathangarelick:
Alright, if someone could take a look at this, it's a bit cleaner now. A couple of notes:
- Dead code removed from AppDelegate.swift
- This code now differs from Firefox's in that the queued tab loading triggers in the `sceneDidBecomeActive` lifecycle hook rather than `applicationDidBecomeActive` [link to FF code for reference](https://github.com/mozilla-mobile/firefox-ios/blob/e155ab6555e03805ca1b363e9a89f7499ff559e5/Client/Application/AppDelegate.swift#L366)
- I don't know if it's still necessary to run the tab loading "on the background queue" due to the lifecycle hook change, but I left that functionality the same, along with the accompanying assertion.
- I don't quite understand the whole purpose of `receivedURLs`. There is a comment about receiving URLs from push notifications, but I can't imagine that this is part of the Neeva app right now (perhaps related to Pocket integration?)

Testing again --

https://user-images.githubusercontent.com/41495718/157125794-ab7409b5-0f77-4ed5-b76e-1e411127af0f.mov

darinwf (approved):
Makes sense! Thanks, LGTM

PR #3014, opened 2022-03-08
Remove and Update NeevaMenu and Overflow Menu Tests

### LINKED ISSUE #2892: Delete code for the Neeva menu
When ios.cheatsheet_query is enabled, the Neeva menu is replaced with the cheatsheet. The tab overflow menu gains access points for some of the features that were previously found in the Neeva menu.

We should configure the client to assume this is the only configuration: that the Neeva menu is no more and the tab overflow always has those items.

We should also assume that ios.cheatsheet_query will be set to true and perhaps reuse that flag or create a new one that can be used to server-side disable the feature.

A big part of this task will be updating all of the tests to not use the Neeva menu.

@EdwardJTL - can you take this task? It can be done on `main` now that we have branched for the 1.28.x release. Thanks!

EdwardJTL:
Fix and update tests following removal of neevamenu code.

Also removed test for new tab button in overflow menu which is no longer there.

### How was this tested?
- [ ] I tested this manually
- [x] I added automated tests

## Issues addressed
closes #2892


XCUITests/BaseTestCase+GoToExtension.swift
```diff
@@ -66,6 +71,8 @@ extension BaseTestCase {
     /// Launches from tab page
     func goToHistory() {
         showAppNavigationMenu(for: "History")
+
+        sleep(1)
```
darinwf, line 79: `sleep` added to a test is usually a recipe for flakiness. the CI can have large unexpected wall clock delays due to running on a VM. perhaps this can be avoided by adding code to wait for the `History` button to become available.

EdwardJTL: yea i tried the wait methods but they are flakey too. It is strange because `showAppNavigationMenu` already calls `waitForExistence` before returning but sometimes the button is not available in `goToHistory`

evancrow: Going to agree with @darinwf and highly suggest avoiding using `Sleep`. Think there's a `waitForTappable` method that might work?

darinwf: Can we use waitForHittable?

EdwardJTL: waitForHittable keeps failing for me

EdwardJTL: I changed it to waitForExistence. let's see if its flakey on the bots

EdwardJTL: looks like its passing on bots 🤷‍♂️

darinwf: Winner!

darinwf (approved):
Ship it!

PR #3015, opened 2022-03-08
add tab pinning tests

chassu:
Add tests for tab pinning.

### How was this tested?
- [ ] I tested this manually
- [x] I added automated tests

After this PR and #2986, I think we can flip the flag.


ClientTests/CardTests.swift
```diff
@@ -305,6 +305,81 @@ class CardTests: XCTestCase {
...
+        let tab1 = manager.addTab()
+        let tab2 = manager.addTab()
+
+        tab2.isPinned = true
+        tab2.pinnedTime = Date().timeIntervalSinceReferenceDate
+        tabCardModel.onDataUpdated()
+
+        let buildRowsTwoTabs = tabCardModel.buildRows(
+            incognito: false, tabGroupModel: self.tabGroupCardModel, maxCols: 2)
+        /*
```
darinwf, line 325: nit: Usually best to delete commented out code since it can bit-rot easily. Alternatively, you can define a local boolean variable that is set to false by default but when true would run this code. Then add a comment about how it can be enabled when you want to debug the test.

ClientTests/CardTests.swift
```diff
@@ -305,6 +305,81 @@ class CardTests: XCTestCase {
...
+        buildRowsTwoTabs.forEach {
+            print(">>> \($0.cells[0].id)")
+        }
+        */
+        XCTAssertEqual(buildRowsTwoTabs[0].cells.count, 2)
+        XCTAssertEqual(buildRowsTwoTabs[0].cells[0].id, tab2.id)
+
+        manager.tabs.removeAll()
+
+        /*
```
darinwf, line 336: Please create separate `testFoo` functions for each of these scenarios. That way if one succeeds but the other fails, the log message about a particular failed test will be more helpful.

darinwf (commented):
Looks pretty good. Appreciate you using unit tests for this.

It may still be worth an integration test to validate the long press menu entry points. You don't need to validate all of the scenarios in that test. Just confirm that the menu items work as advertised. The unit test can provide the more detailed coverage for the buildRows function.

ClientTests/CardTests.swift
chassu: Fixed

ClientTests/CardTests.swift
```diff
@@ -305,6 +305,80 @@ class CardTests: XCTestCase {
...
+        let buildRowsThreeTabs = tabCardModel.buildRows(
+            incognito: false, tabGroupModel: self.tabGroupCardModel, maxCols: 2)
+
+        XCTAssertEqual(buildRowsThreeTabs[0].numTabsInRow, 2)
+        XCTAssertNotEqual(buildRowsThreeTabs[0].cells[0].id, tab5.id)
+    }
+
+    func testPinnedTabsBeforeNonPinnedTabs() throws {
+        // TODO: (Charles) enable this test after #2986 is merged
+        var condition = false
```
darinwf, line 355: nit: You might just add a `return` statement instead of defining and checking this variable. That'd be simpler and also when you are ready to activate the test, you only need to delete that `return` statement. No need to reformat the other lines, etc.

darinwf (approved):

Issue #3009, opened 2022-03-07
Space Detail should be using NavigationView-NavigationLink


yozuysal:
As prep work for Profile UI work, we need to make sure we can construct a working stack of NavigationLinks as the user clicks through Space detail-> Profile UI -> Space detail again.

1) Replace Grid-> Space detail transition with a NavigationLink transition. This should also cleanup the current custom DraggableDetail
2) Cleanup anything that is related with the animation after the above

PR #3017, opened 2022-03-08
NavigationView for DetailView

evancrow:
### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/37356984/157173449-128f6c4c-0634-4ea3-8696-7a46bfe1e317.MP4

## Issues addressed
Fixed: #3009


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -660,6 +659,10 @@ class SpaceCardModel: CardModel {
                 Defaults[.seenBlackFridayNotifyPromo] = true
             })
     }
+
+    func openedSpace() {
```
yozuysal, line 661: not needed?

Client/Frontend/Browser/Card/SpaceDetailList.swift
```diff
@@ -74,182 +74,180 @@ struct SpaceDetailList: View {
...
+                    if let space = space, !space.isDigest {
+                        SpaceCommentsView(space: space, model: spaceCommentsModel)
+                            .modifier(ListSeparatorModifier())
+                            .id("CommentSection")
                     }
-                    .useEffect(deps: spaceCommentsModel.addingComment) { addingComment in
-                        self.addingComment = addingComment
+                }
+                .modifier(ListStyleModifier(isDigest: space?.isDigest ?? false))
+                .navigationBarTitle("Space")
```
yozuysal: why do we need this?

Shared/API/SpaceStore+CoreSpotlight.swift
```diff
@@ -74,7 +76,7 @@ extension SpaceStore {
     }
 
     /// Remove spaces from corespoltight by id
-    class func removeSpacesFromCoreSpotlight(
+    public class func removeSpacesFromCoreSpotlight(
```
yozuysal, line 79: I am not sure why we need this change in this PR

Client/Frontend/Browser/Card/CardGrid.swift
```diff
@@ -92,7 +92,24 @@ struct CardGrid: View {
...
+    }
+
+    @ViewBuilder
+    var detailedSpaceView: some View {
+        if let detailedSpace = detailedSpace {
+            DetailView(primitive: detailedSpace)
+                .onDisappear {
+                    gridModel.closeDetailView()
+                }
+        } else {
```
yozuysal, line 106: we shouldnt need the else here, if with a ViewBuilder should give and empty view when the condition is not met

evancrow:
Found a few more crashes I need to fix...

Client/Frontend/Browser/Card/GridModel.swift
```diff
@@ -24,7 +24,7 @@ class GridModel: ObservableObject {
         }
     }
 
-    @Published var showingDetailView = false
+    @Published var showingDetailView = true
```
yozuysal, line 27: tbh I am not really sure why we need this state at all. This should always be equivalent to spaceModel.detailedSpace != nil, and keeping it as a separate state may leave them prone to getting out of sync. Hmm, I wonder if there is a way to not need this at all.

yozuysal (commented):
Thanks for the changes! Test drove it and it actually looks pretty good!

There are a few things about the state handling that I think we will need to cleanup, but tbh I am torn between taking this change as is and over the course of the week slowly tweaking the state handling logic vs changing all in place. It is quite a large PR.. Let's chat tomorrow.

yozuysal (commented):
thanks for these changes, but if you have zoomed in on the crash itself, I would make this only about the crash. As I mentioned, I think we will need to think through the state dependencies a bit before we make a call here.

Right now this slices through the isShowingDetails logic (which may be a good way) but it does it by adding a circular dependency to spaceModel. I would suggest reverting these changes and only addressing unfollow crash.

evancrow:
Unfortunately, the state change was my fix to the crash :(. I did add an optional option to `manager` which avoided the crash but broke the functionality.

yozuysal:
Umm, frankly that sounds like an overkill for a crash to unfollow.. there should be a way to bandaid fix that functionality without changing detail showing logic for all spaces. (Again, I think that logic should change, but this is gonna be in a rush)

evancrow:
Yeah haha, was a bit overkill. Pretty much put the logic back to the way it was, just changed how detailedSpace was set.

yozuysal:
not sure what changed with Fixes for iPad, but now the drag from the edge to close is not working on phones.

evancrow:
This is what you would see on iPad when the app launched before my fix:

Client/Frontend/Browser/Card/CardGrid.swift
```diff
@@ -148,100 +151,28 @@ struct CardGrid: View {
...
-                }.modifier(
-                    DraggableDetail(
-                        detailDragOffset: $detailDragOffset,
-                        width: geom.size.width))
+                NavigationLink(
+                    destination: detailedSpaceView,
+                    isActive: $gridModel.showingDetailView
+                ) {}.useEffect(deps: spaceModel.detailedSpace) { detailedSpace in
+                    gridModel.showingDetailView = detailedSpace != nil
+                }
```
yozuysal, line 159: I think this is missing .environment(\.onOpenURLForSpace), we are falling back to openURL which loses setting the parentID and rootID to spaceID

yozuysal: also the shareURL is missing so share button crashes.

I think we pretty much need to carry over all important environemnt variables :)

evancrow: I follow that logic :)

evancrow: `onOpenURLForSpace` is passed down via BrowserView (as is `openURL`). `ShareURL` was missing but I just added it, will push in a sec.

yozuysal: onOpenURLForSpace is definitely not getting through though. When you click any link, look for the back button being enabled, if it is not, then there was no onOpenURLForSpace and it fell back to use openURL instead.

yozuysal (approved):
I think this is in good shape now. Thanks Evan! 🤞

evancrow:
Tests not related to this and failing from main, going to merge

PR #3024, opened 2022-03-08
Use SwiftUI based logging for empty cheatsheet

EdwardJTL:
The method of logging based on loading status logs cases where we exited fetching early due to supported URLs.
Moving back to logging based on SwiftUI onAppear so that we are sure that empty page was meant to be shown instead of education or error pages.
Added variable to track if a fetch request has been made so that we don't log false positives when opening a link from cheatsheet.

@macy-n the url i have been using to bring up empty cheatsheet now returns data. could we find another obscure website?

I left the old combine pipeline in the code. we might need to return to it when we eventually switch to using a single cheatsheet view model

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested manually by observing logs.
Tested opening link from cheatsheet that will create a new tab or switch to an existing one; verified that no log was produced.


darinwf (approved):
Makes sense to me. Would be good to get @macy-n's eyes on this too.

macy-n (approved):

Issue #2838, opened 2022-02-09
Missing thumbnails


darinwf:
From Ario:

<img src="https://user-images.githubusercontent.com/79867437/153308521-579cbab5-3c8b-4f5e-8e1c-41bd894135c8.png" width="300">

I've seen this too.

darinwf:
I added some logging and observed the bug. Here's what I learned:

1. It looks like a newly added set of tabs did not have their screenshots saved.
2. It looks like the background thread where the screenshots are written out to disk did not run (see DiskImageStore.swift).
3. I did not see those background tasks getting interrupted. They just didn't seem to run.

Now, my logging was not as complete as it could have been. I don't see confirmation that we even attempted to start saving the screenshots, but I suspect we probably did.

It seems quite plausible that the app could get backgrounded when these screenshots have not yet been written out.

For other tabs state we take care to ensure background IO happens before we signal to the OS that we are ready to be killed. We should probably do that for DiskImageStore as well.

darinwf:
Unfortunately, I don't think https://github.com/neevaco/neeva-ios/pull/3029 was sufficient to resolve this issue. I believe I observed the bug using 1.32.0, which included (see https://github.com/neevaco/neeva-ios/commits/Build-187) that change.

There must be more to this puzzle. Will need to add more logging and try to again catch it in the act.

EdwardJTL:
> The times I've seen missing thumbnails recently may just be explained by this: #3089
> 
> Might just be that we are not taking a screenshot when we should.

in #3078 we don't have screenshots when opening link from cheatsheet because `createOrSwitchToTab` don't take screenshot. `BrowserViewController.finishEditingAndSubmit` may also call that function. Maybe we should look at when we should be taking screenshots in the code in general?

darinwf:
@EdwardJTL - yeah, it's likely that one patch would fix both issues. I see that `TabManager.createOrSwitchToTabForSpace` calls `takeScreenshot`, so one approach would be to just mimic that code inside `createOrSwitchToTab`. I think that might be a reasonable place to start.

darinwf:
Going to call this fixed now.

darinwf:
Actually, nevermind... just re-read the last few comments.

jedireza:
Moved to ClickUp 🔒 https://app.clickup.com/t/3vxg36y

PR #3029, opened 2022-03-09
Use beginBackgroundTask to protect DiskImageStore writes

darinwf:
Refactors BackgroundFileWriter so it can be used by DiskImageStore.

Issue #2838

### How was this tested?
- I updated existing automated tests


Client/Frontend/Browser/Tabs/SavedTab+ConfigureExtension.swift
```diff
@@ -57,7 +57,7 @@ extension SavedTab {
             let imageStore = imageStore
         {
             tab.screenshotUUID = screenshotUUID
-            imageStore.get(screenshotUUID.uuidString) >>== { screenshot in
+            imageStore.get(screenshotUUID.uuidString) { screenshot in
```
chung-lai, line 60: hmm...not sure what `>>==` operator is, is that we don't do any sql operation anymore for image?

Client/Frontend/Browser/Tabs/TabManagerStore.swift
```diff
@@ -58,22 +58,19 @@ class TabManagerStore {
...
-                savedUUIDs.insert(screenshotUUID.uuidString)
-
-                imageStore?.put(screenshotUUID.uuidString, image: screenshot)
+            if let image = tab.screenshot, let uuid = tab.screenshotUUID {
+                screenshots.append(.init(key: uuid.uuidString, image: image))
             }
         }
-        // Clean up any screenshots that are no longer associated with a tab.
-        _ = imageStore?.clearExcluding(savedUUIDs)
+        imageStore?.putAll(screenshots)
```
chung-lai, line 72: nit: probably the method name should mention about it clears the unused screenshots

Storage/DiskImageStore.swift
```diff
@@ -39,70 +32,97 @@ open class DiskImageStore {
...
+
+    /// Updates the store with the full set of entries that should be stored. Removes
+    /// any existing entries that are not included.
+    open func putAll(_ entries: [Entry]) {
+        processor.performTask {
+            var keysToKeep = Set<String>()
+
+            for entry in entries {
+                keysToKeep.insert(entry.key)
+                self.putOne(key: entry.key, image: entry.image)
```
chung-lai, line 83: What do you think about logging here when an existing entry is found for debugging purpose?

chung-lai: Or add logging when a file is successfully written?

chung-lai (approved):
Awesome refactoring! LGTM, left a few minor comments.

Client/Frontend/Browser/Tabs/SavedTab+ConfigureExtension.swift
darinwf: This is from `DeferredUtils.swift`. It just runs the callback only when there is success. I got rid of using `Deferred` and instead just made the second parameter to `DiskImageStore.get` be an ordinary callback function that gets passed a `UIImage?` that will be nil if there is a lookup error. I think this is a simpler API.

Client/Frontend/Browser/Tabs/TabManagerStore.swift
darinwf: I was kind of intentionally burying the detail about cleanup. I'll give some more thought to the name....

Storage/DiskImageStore.swift
darinwf: I didn't feel a strong need for it and it can add noise to the log file when we are trying to study problems.

chung-lai:
Looks like test fails to build.

PR #3033, opened 2022-03-09
Remove Indexed Browsing Activities When Deleting Browsing History

EdwardJTL:
Add calls to delete indexed `NSUserActivity`s from Spotlight when browsing history has been cleared.
This applies to both clearing from `HistoryPanel` and `DataManagementView`(Settings).

HistoryPanel can clear history by recent time period. Since we don't keep identifiers to the activity items, we delete all browsing items regardless of the time period.

Also added an abstract interface from UserActivityHandler to remove items.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Opened some webpages. Check they are in spotlight. 
Delete history in settings. Check they are removed from spotlight (a new search is needed since spotlight keeps results between launches)
Repeat and clear in history panel.


Client/Frontend/Library/HistoryPanel.swift
```diff
@@ -266,6 +266,7 @@ class HistoryPanel: SiteTableViewController {
                     ofTypes: types, modifiedSince: date, completionHandler: {})
 
                 self.profile.history.removeHistoryFromDate(date).uponQueue(.main) { _ in
+                    UserActivityHandler.clearIndexedItems()
```
darinwf, line 269: nit: Would be good to add a comment here about why we just wipe out everything

darinwf (approved):
Thanks for catching and fixing this issue!

PR #3036, opened 2022-03-09
Switch to use keys for node urls

yozuysal:
Switch from hardcoded urls to the ones from config for node urls. 

### How was this tested?
- [X] I tested this manually
- [ ] I added automated tests

### Manual test cases
Enable wallet and test node related functionality


WalletCore/CryptoConfig.swift
```diff
@@ -62,17 +62,23 @@ public enum EthNode: String, CaseIterable, Identifiable {
...
         case .Polygon:
-            return "https://polygon-mainnet.infura.io/v3/25b0f8c5b9004da38236104927e630b5"
+            return "CRYPTO_POLYGON_URL"
         }
     }
 
+
+
+    public var url: URL {
+        URL(string: Bundle.main.object(forInfoDictionaryKey: configKey) as! String)!
```
chung-lai, line 79: Given not all devs are going to set up these api keys in their environment. We should return `URL?` and remove `!`

yozuysal: that is a great point :)

It did require a bit more surgery to use URL? here though. PTAL

WalletCore/CryptoConfig.swift
```diff
@@ -112,58 +114,24 @@ public class CryptoConfig {
         return "My Neeva Crypto Wallet"
     }
 
-    public func sendEth(amount: String, sendToAccountAddress: String, completion: () -> Void) {
-        do {
```
chung-lai, line 116: I assume these are all just clean up? Not sure how this relate to changing to `URL?` :P

chung-lai (approved):
Just one comment about the refactoring change. otherwise, LGTM!

WalletCore/CryptoConfig.swift
yozuysal: a nil URL creates a nil web3 object and the path that I am calling now is guarded against having a nil web3 (it refreshes it)

PR #3038, opened 2022-03-10
Fix Sign-in Close Padding

evancrow:
Fixes the padding of the IntroView by using the safe area. It also creates a better way of passing the SafeArea down from the outermost view, since calculating the SafeArea from within a container isn't always accurate.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

Before:

After:


darinwf (approved):
LGTM

I also test drove on an iPhone SE simulator as well, and it looks great.

evancrow:
Tests flakey, unrelated

PR #3040, opened 2022-03-10
NYTimes TabGroup Case

evancrow:
Adds support for creating a tab group when a user navigates back to a tab that opened a child tab. (See test cases before or see the Tab Group paper doc for more info)

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
1:
- Create tab (nytimes.com)
- Navigate to link (nytimes.com/article)
- Create a new tab with original URL (nytimes.com)
- Check tab group created ✅

2: (tab already in tab group)
- Search new york times
- Open tab from SRP (nytimes.com) (creating tab group)
- Navigate to link (nytimes.com/article)
- Create a new tab with original URL (nytimes.com)
- Check nytimes.com inserted intro already existing tab group ✅

Everything else seems to perform normally (using `<a target=_blank ...>`) but LMK if there are other paths to try!

1st case:

https://user-images.githubusercontent.com/37356984/157582039-9e8f8e45-1f5e-48c6-b5c4-cf4c58d4e6a8.mp4

2nd case:

https://user-images.githubusercontent.com/37356984/157582078-c61bdecf-d57f-4b46-94d0-597171f0edda.mp4


Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -133,6 +153,32 @@ extension TabManager {
...
+        let shouldCreateTabGroup = possibleChildTab.originURL == newTab.url
+
+        if shouldCreateTabGroup {
+            if possibleChildTab.parent == nil {
+                // Create a tab group by setting the child tab's parent.
+                possibleChildTab.parent = newTab
+                possibleChildTab.parentUUID = newTab.tabUUID
+                possibleChildTab.rootUUID = newTab.rootUUID
+            } else {
+                // Set the new tab's parent the same as the current tab,
```
darinwf, line 171: Hmm, I thought parent / parentUUID is used to enable the simulated swipe to go back gesture. It is just the rootUUID property that indicates inclusion in a tab group.

I think this is confusing because when a tab group is formed through target=_blank, the parent / parentUUID will be set to the tab where the link was tapped.

For this case where we are trying to re-add the hub site, I don't think we need to be configuring the parent / parentUUID properties. It should just be enough to set rootUUID.

(Also, I don't think we currently set `Tab.parent` when restoring tabs, which could be another bug, but it would cause complications if we depend on `Tab.parent` here.)

Probably all we need to really do here is just this:
```
newTab.rootUUID = possibleChildTab.rootUUID
```

In other words, put the `newTab` in the same tab group as `possibleChildTab`, and then ... insert it into `TabManager.tabs` in the right spot, which you are doing.

darinwf (commented):
I think this PR calls for some automated tests. Thanks!

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -109,6 +109,11 @@ class Tab: NSObject, ObservableObject {
         }
     }
 
+    // MARK: - Navigation Properties
+    /// URL of the page that opened the current URL
+    var originURL: URL? {
+        backList?.last?.initialURL
```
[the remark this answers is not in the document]
evancrow: Isn't the backlist is backwards? The most recent item is last in the list, may be mistaken though.

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -111,11 +111,31 @@ extension TabManager {
...
 
-    func insertTab(_ tab: Tab, atIndex: Int? = nil, parent: Tab? = nil, notify: Bool) {
+    private func insertTab(_ tab: Tab, atIndex: Int? = nil, parent: Tab? = nil, notify: Bool) {
         if let atIndex = atIndex, atIndex <= tabs.count {
             tabs.insert(tab, at: atIndex)
         } else if parent == nil || parent?.isIncognito != tab.isIncognito {
-            tabs.append(tab)
+            var insertIndex: Int? = nil
+
+            if !isIncognito {
```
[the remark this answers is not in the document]
evancrow: Sure

Client/Frontend/Browser/Tabs/Tab.swift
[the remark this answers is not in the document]
darinwf: *oh*... maybe we are thinking about the design differently.

I was thinking about the following user journey:
1. Each morning, I hit the URL bar and select NYTimes from my Suggested Sites
2. I scroll through the headlines and click on links to read stories
3. Next day I hit the URL bar and repeat the above steps
4. When I look in the tab switcher, I see a tab group containing all of the stories I read on nytimes.com

I think there is a variation of this scenario where a user goes to nytimes.com, clicks on an article and then clicks again to go to another URL from that page. So they go from A to B to C.

In order to cluster a tab that is just on URL C with a new tab for URL A, we need to check the first URL that was ever opened in the tab that now contains URL C. It isn't sufficient to just look at the previous URL -- URL B -- that was in that tab.

Does that make sense? This is why I thought we should use `backList?.first`.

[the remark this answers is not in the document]
evancrow: Yup, makes sense now. Will change it to use first now.

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
evancrow: Yeah should create a separate issue for the restore tabs->parent bug

XCUITests/BaseTestCase.swift
```diff
@@ -142,22 +142,11 @@ class BaseTestCase: XCTestCase {
             app.buttons["Add Tab"].tap()
         }
 
-        UIPasteboard.general.string = url
```
evancrow, line 145: For some reason, pasteboard stopped working on tests for me... switched over to `typeText`

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -133,6 +152,30 @@ extension TabManager {
...
+    @discardableResult func addTabToTabGroupIfNeeded(
+        newTab: Tab, possibleChildTab: Tab
+    ) -> Bool {
+        let childTabOriginalURL = possibleChildTab.originalURL?.normalizedHostAndPathForDisplay
+        let newTabURL = newTab.url?.normalizedHostAndPathForDisplay
+        let shouldCreateTabGroup = childTabOriginalURL == newTabURL
+
+        if shouldCreateTabGroup {
+            if possibleChildTab.parent == nil {
+                // Create a tab group by setting the child tab's parent.
```
darinwf, line 167: nit: comment needs to be updated

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -133,6 +152,30 @@ extension TabManager {
...
+                possibleChildTab.rootUUID = newTab.rootUUID
+            } else {
+                // Set the new tab's parent the same as the current tab,
```
darinwf, line 170: Ditto. please update the comments

darinwf (approved):

PR #3041, opened 2022-03-10
[ios] add news query and places query

macy-n:
- removed getAllProviders api endpoint because it's being deprecated and no result will be return from server
- added query for news widget
- added query for local results


Shared/API/ProviderList.swift
```diff
@@ -15,26 +15,7 @@ public class ProviderList: ObservableObject {
...
-                        }
-
-                    }
-                }
-            case .failure(let error):
-                logger.error(error)
-            }
-            self.isLoading = false
-        }
+        // Todo remove Provider Related code because API is deprecated.
```
EdwardJTL, line 18: do we have an issue created for this?

EdwardJTL (approved):

Shared/API/ProviderList.swift
macy-n: Thanks for reminding, created a clickup ticket https://app.clickup.com/t/2h61f4j

Issue #3035, opened 2022-03-09
TabManagerStoreTests.testAddedTabsAreStored is flaky


darinwf:
This is concerning as this is a unit test and should be rock solid.

darinwf:
I've noticed that now the incognito variant is failing. This makes me wonder if some other test that runs before these tests is leaving behind some bad state that is messing up this test.

darinwf:
@EdwardJTL - looks like you may have fixed this issue w/ your change here: 128e60d98814c87c43e7a532f7732a938b4595a0

We can probably re-enable `testAddedTabsAreStored` to see.

darinwf:
Marking fixed now

PR #3046, opened 2022-03-10
Attempt to reduce flakiness of testAddedTabsAreStored

darinwf:
Switch from loading yahoo.com to example.com. Ideally, we'd use
the test-fixture that we have for XCUITests here, but that's not
setup.

Surprisingly the 1.1 second wait time was being exceeded before
the scheduled task w/ 0.1 sec delay that would fulfill the
expectation was run. Bumping up the timeout for this to 5 sec.

Issue #3035


macy-n (approved):

darinwf:
@macy-n - please take another look. went with the approach of fetching a local resource instead of fetching from an actual website. I think this may be an approach we should be using elsewhere to root out flakiness.

macy-n (approved):

macy-n:
Looks good to me!

Issue #3001, opened 2022-03-06
PDF thumbnails are not saved


darinwf:
As part of investigating why thumbnails are sometimes blank (see https://github.com/neevaco/neeva-ios/issues/2838), I've noticed that PDF thumbnails are never saved.

darinwf:
It looks like the `jpegData` calls made when storing the thumbnails to the `DiskImageStore` fail in this case.

darinwf:
Chrome and Firefox also have this bug.

I see that `UIImage.jpegData` is failing and yet there is a non-nil and non-empty underlying `CGImage`. Hmm...

EdwardJTL:
This function may return nil if the image has no data or if the underlying CGImageRef contains data in an unsupported bitmap format. perhaps the bitmap format is not supported by UIImage? we could redraw the CGImage or ask the CGImage to write to a png/jpg file and then ask UIImage to read from it?

darinwf:
Perhaps we can just log details about the CGImage to see how it differs in the PDF case from the normal case. Maybe that will give some clues. I like the suggestion about redrawing the CGImage or what not. Seems like there should be a solution here.

PR #3048, opened 2022-03-10
Show PDF tab titles on cold start

### LINKED ISSUE #3043: PDF tabs do not have a title (or display a URL) when restoring tabs on cold start
This is in addition to the recently fixed screenshots issue (https://github.com/neevaco/neeva-ios/issues/3001).

@EdwardJTL FYI

jonathangarelick:
Shows PDF tab titles on cold start. A couple notes:
- I don't think I have ever seen proper titles on PDF tabs? (Regardless of cold start)
- Still seeing the issue in #3001 (see video -- edit: Maybe this is an issue with iOS simulator. I seem to be able to view some thumbnails on my device 😕 )

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
See video.

https://user-images.githubusercontent.com/41495718/157772293-aa4efba7-a713-43bb-82d4-7ad93ab36e3d.mov

## Issues addressed
* Fixes #3043


Client/Frontend/Browser/Tabs/SavedTab+ConfigureExtension.swift
```diff
@@ -65,7 +65,8 @@ extension SavedTab {
         }
 
         tab.sessionData = sessionData
-        tab.lastTitle = title
+        // Use current URL as lastTitle when the tab loads a PDF, for example.
+        tab.lastTitle = (title ?? "").count > 0 ? title : sessionData?.currentUrl?.absoluteString
```
EdwardJTL, line 69: maybe we should also guard against white space titles, something like
`(title?.isEmptyOrWhitespace() ?? false) ? title : sessionData?.currentUrl?.absoluteString`

EdwardJTL (commented):
Thanks for making the change!
re: thumbnails not showing. I noticed that on simulators, the image captured from WKWebView is blank/black when the url is a pdf. So that is a different issue than #3001.

Client/Frontend/Browser/Tabs/SavedTab+ConfigureExtension.swift
jonathangarelick: Yeah, I figured that was the case with the simulator. Fortunately, I have my physical device working now (#3066). Made your suggested changes -- thanks!

EdwardJTL: ah sorry i think we need to flip the output from the check `!title?.isEmptyOrWhitespace()`

jonathangarelick: I see what you mean. I'll have to look at this with fresh eyes tomorrow. Too much punctuation, not enough coffee :)

jonathangarelick: Alright, I've made a change which I think is a bit more of an elegant solution. Instead of a multi-line conditional or this cryptic statement

`tab.lastTitle = (!(title?.isEmptyOrWhitespace())! ?? false) ? title : sessionData?.currentUrl?.absoluteString`

I added a utility method to String. The statement should now properly handle nil, empty Strings, and whitespace Strings. Let me know if this works for you.

EdwardJTL (approved):

PR #3050, opened 2022-03-11
Add Support Option to Cheatsheet

EdwardJTL:
Add button to bring up support from cheatsheet
Add utility to take a full page screenshot of the cheatsheet content

## Caveats
the method used to take a screenshot is expensive and does not preserve view states, such as the horizontally scrollable content or the collapsible recipe. 
one alternative would be to restructure the code so that the content of the cheatsheet is presented in a UIHostingView. That is quite a bit of effort, and we'd need to still set up a bridge to trigger the render command and retrieve the image for support. 
another alternative is to write a proper view modifier that wraps the view in a hosting controller and provide some hooks for triggering and receiving the screenshot. To make this a proper view modifier, I believe we would still need to use UIViewRepresentables such that the output is a SwiftUI View again on which further modifiers can be applied.
Either way, it is going to be a bit of effort. I propose we wait and see if this feature actually gets used.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested opening support from cheatsheet repeatedly. Tested that the support routes in other parts of the app still take the right screenshot.

https://user-images.githubusercontent.com/44557675/157782269-6ed7aea0-79bd-4099-b892-12ae0c121cb6.MP4

## Issues addressed
clickup


Client/Frontend/Overlay/Content/Menu/OverflowMenu/CardGridOverflowMenuView.swift
```diff
@@ -35,7 +35,7 @@ public struct CardGridOverflowMenuView: View {
             GroupedCell.Decoration {
                 VStack(spacing: 0) {
                     GroupedRowButtonView(label: "Support", symbol: .bubbleLeft) {
-                        menuAction(.support)
+                        menuAction(.support())
```
EdwardJTL, line 38: unfortunate side effect of adding associated value

Client/Extensions/UIViewExtensions.swift
```diff
@@ -45,3 +46,36 @@ extension UIView {
         return screenshot(size, offset: offset, quality: quality)
     }
 }
+
+extension UIView {
```
darinwf, line 50: nit: you could also just these new functions into the existing `extension UIView {` block. It seems to be the convention above with the two existing `screenshot` methods.

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuView.swift
```diff
@@ -329,6 +331,29 @@ public struct CheatsheetMenuView: View {
                 }
             } else {
                 cheatsheetContent
+                    .background(
+                        GeometryReader { proxy in
```
darinwf, line 335: Any performance issues to be worried about here? Hopefully this geometry is not getting updated when the cheatsheet is being scrolled.

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuView.swift
```diff
@@ -363,7 +388,32 @@ public struct CheatsheetMenuView: View {
...
             reviewURLSection
             memorizedQuerySection
+
+            supportSection
+        }
+    }
+
+    @ViewBuilder
+    var supportSection: some View {
+        VStack(alignment: .leading, spacing: 10) {
```
darinwf, line 398: Just want to confirm. Is this spacing to spec w/ Ria's mock. I went looking for the mock, but couldn't locate it. Just thought I'd double check with you.

darinwf (approved):

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuView.swift
EdwardJTL: ah. i did not check with the mock. I copied this from other sections of the cheatsheet. will verify

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuView.swift
EdwardJTL: this part does get called when cheatsheet is being scrolled. but i think it is probably very low cost since it is only drawing `Color.clear`, and nothing in the `onChanged` closure is evaluated until the button is pressed

PR #3052, opened 2022-03-11
Possible Fix for AutocompleteTests

evancrow:


chung-lai:
Interesting, so the change we are making is to tap on a cancel button? When I look at the failure, it says something about extra character added at the end after typing the url. 🤔

evancrow:
I've seen it fail sometimes opening the address bar... guess the extra character is a separate issue too? I'll try to find a way to prevent that.

chung-lai:
Ah, interesting, trying another one here to address the extra character issue: https://github.com/neevaco/neeva-ios/pull/3056. May be we can combine these two? Feel free to pick mine onto this branch if it helps anyway

evancrow:
I think yours is the best fix honestly. Looking at other instances of the test failure, I only saw the address bar issue once. I'll just close this PR, and can push my change to your PR/branch if needed.

PR #3055, opened 2022-03-11
Spotlight Setting Clearing Fixes

EdwardJTL:
I forgot to clear a SwiftUI State and clearing all index button keeps spinning.
Also added logging when clear index errors. Updated the use of class functions to support delegate

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Observed that the button works, and logging function can be called via breakpoints.


Shared/API/SpaceStore+CoreSpotlight.swift
```diff
@@ -144,31 +146,48 @@ extension SpaceStore {
...
     // Remove all space items in the index
-    public class func removeAllSpacesFromCoreSpotlight(completionHandler: ((Error?) -> Void)? = nil)
-    {
-        Self.CSIndexQueue.async {
-            searchableIndex.deleteSearchableItems(
-                withDomainIdentifiers: [CSConst.spaceDomainIdentifier],
-                completionHandler: completionHandler
-            )
+    public func removeAllSpacesFromCoreSpotlight(completionHandler: ((Error?) -> Void)? = nil) {
+        Self.CSIndexQueue.async { [weak self] in
```
darinwf, line 150: Why use `[weak self]` here? It's a one-time callback, so it should be okay to just strongly bind to `self`, no?

Shared/API/SpaceStore+CoreSpotlight.swift
```diff
@@ -144,31 +146,48 @@ extension SpaceStore {
...
-    public class func removeAllSpaceURLsFromCoreSpotlight(
+    public func removeAllSpaceURLsFromCoreSpotlight(
         completionHandler: ((Error?) -> Void)? = nil
     ) {
-        Self.CSIndexQueue.async {
-            searchableIndex.deleteSearchableItems(
-                withDomainIdentifiers: [CSConst.spacePageDomainIdentifier],
-                completionHandler: completionHandler
-            )
+        Self.CSIndexQueue.async { [weak self] in
```
darinwf, line 167: ditto

Shared/API/SpaceStore+CoreSpotlight.swift
```diff
@@ -144,31 +146,48 @@ extension SpaceStore {
...
+            }
         }
     }
 
     // Clear all items in the index
-    public class func clearIndex(completionHandler: ((Error?) -> Void)? = nil) {
-        Self.CSIndexQueue.async {
-            searchableIndex.deleteAllSearchableItems(completionHandler: completionHandler)
+    public func clearIndex(completionHandler: ((Error?) -> Void)? = nil) {
+        Self.CSIndexQueue.async { [weak self] in
```
darinwf, line 183: ditto

Shared/API/SpaceStore+CoreSpotlight.swift
EdwardJTL: ah i copied the code from the other parts in the file, but i forgot that I meant to make all these callbacks to file on the main thread. fixed now!

darinwf: I see, but why `[weak self]`? What's wrong with using a normal strong reference here? Since this is a one-time async dispatch, there isn't a concern about a reference cycle, right?

EdwardJTL: yes, it is not a concern about ref cycle/mem leak. It is more so a precaution in case we become sensitive about the life cycle of the object since a strong reference will keep the object alive for as long as the closure is in memory.

darinwf: I see, but it is a singleton... `SpaceStore.shared`... by using weak references we also risk flakiness about whether or not some of the steps happen. Maybe not a concern.

darinwf: I'd recommend using strong references just to simplify logic here. We generally avoid `deinit` methods (good practice!), so extending the lifetime of objects shouldn't be a big concern.

EdwardJTL: done

darinwf (approved):

PR #3057, opened 2022-03-11
Create onAppUpdate Method

evancrow:
Adds a function in which we can put logic to run when the user updates the app.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


Client/Application/SceneDelegate.swift
```diff
@@ -473,4 +475,23 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+                onAppUpdate(version: version)
+            }
+
+            Defaults[.lastVersionActiveOn] = version
+        }
+    }
+
+    /// Called when the user updates the app, and then opens it.
+    /// Useful for updating flags, migrating data, etc.
+    private func onAppUpdate(version: String) {
```
EdwardJTL, line 494: maybe it will be helpful to have the previous version too? in development we might be going backwards, or if sometimes user skips an update but a data migration should still be carried out?

EdwardJTL (approved):
LGTM

PR #3058, opened 2022-03-11
Flip Spotlight Flag

EdwardJTL:
Enable spotlight by default

- Flip default value in defaults so that new installs will have it enabled
- add temporary worker to flip the flag on first launch after update for existing devices.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested manually on device


Client/Frontend/Settings/Debug/SpotlightSettingsView.swift
```diff
@@ -115,6 +116,13 @@ struct SpotlightSettingsView: View {
                 }
                 .disabled(spaceIndexBusy)
             }
+            
+            Section {
+                Toggle(
+                    String("Enable worker to overwrite flags"),
```
darinwf, line 122: nit: "Enable *worker*..."? Who / what is the worker? Maybe this should be called "Force enable Spotlight integration"?

darinwf (approved):

EdwardJTL:
@evan i didn't use the on app version updated because i realized that it would trigger every time the app updates, but we only want this to happen once and respect the setting afterwards.
We need a util to parse the version strings that adopts the Comparable protocol so i can say `if prevVersion < <some version> { overwrite flags } `

evancrow:
> @evan i didn't use the on app version updated because i realized that it would trigger every time the app updates, but we only want this to happen once and respect the setting afterwards.
> We need a util to parse the version strings that adopts the Comparable protocol so i can say `if prevVersion < <some version> { overwrite flags }`

Ah right, I'll put up another PR that converts those to `int`.

PR #3059, opened 2022-03-12
Add Logic for onAppUpdate

evancrow:


Client/Application/SceneDelegate.swift
```diff
@@ -491,7 +491,13 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
-    private func onAppUpdate(version: String, previousVersion: String) {
-        // Nothing here yet!
+    private func onAppUpdate(previousVersion: String) {
+        // Remove the periods from the version string.
+        // (e.x.) 1.1.1 -> 111
+        let parsedPreviousVersion = Int(previousVersion.replacingOccurrences(of: ".", with: ""))
+        
+        // Example usage:
+        // (e.x.) current version: 1.x.x
+        // if parsedPreviousVersion < 1xx { your code }
```
EdwardJTL, line 501: i'm not sure if this is right? for example going from 1.10.0 to 2.1.0 will break.
there's a couple of suggestions here
https://stackoverflow.com/questions/27932408/compare-two-version-strings-in-swift
perhaps we should write a library and have some test cases for it

EdwardJTL:
can we add some test cases?

EdwardJTL (approved):
LGTM! thanks!

darinwf:
Please add a description to PRs that explains why. It is not obvious from reading this PR why you all wanted to implement this method. Thanks!

PR #3061, opened 2022-03-12
Enable simulator support on an M1 mac

darinwf:
WIP

Removing the excluded architectures from `carthage_command.sh` seems to be the key change needed.

However, in my testing, I'm finding that `carthage bootstrap` sometimes still decides to exclude arm64 simulator support for some of the packages. Not all. Just some. And the set is not consistent from run to run.

If I use `carthage update` instead, it seems to always reliably generate arm64 simulator support. Of course we can't just use that since it will rev the packages to their latest versions, which is not what we want the bots doing, etc.


bootstrap.sh
```diff
@@ -9,39 +9,6 @@
 # already exists then nothing is done. This speeds up builds on
 # CI services where the Carthage directory can be cached.
 #
-# Use the --force option to force a rebuild of the dependencies.
```
darinwf, line 12: Just included the deletion of this script here. We don't use it.

darinwf:
I think this is a good step forward for people trying to build with M1. There may still be issues to work through as described above, but I think this is a good step.

EdwardJTL (approved):
nit: could we keep the force option? it's easier than manually deleting all the carthage products

darinwf:
> nit: could we keep the force option? it's easier than manually deleting all the carthage products

Oh, sure thing. I didn't realize it was being used. Thanks for catching that!

darinwf:
Perhaps `~/Library/Caches/org.carthage.CarthageKit` explains the issue I was seeing with this PR not working reliably. I wasn't deleting that.

chung-lai (approved):
LGTM. Wish I have upgraded to M1 mac to test drive this.

carthage_command.sh
```diff
@@ -1,13 +1,11 @@
 #!/bin/sh
 
-echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64=arm64 arm64e armv7 armv7s armv6 armv8' > /tmp/tmp.xcconfig
```
chung-lai, line 3: Just curious, do you notice any difference in terms of build time? Or in CI?

darinwf: Yes, build time increases a bit :-(

I'm not sure about impact to CI, but at least carthage dependencies are cached by CircleCI.

PR #3065, opened 2022-03-12
tab group drag-n-drop

chassu:
Apologies that I probably have messed up when rebasing the old branch so I'm creating a new PR. This support drag-n-drop for non-pinned tabs. I also fixed the preview issue, but will only available for iOS 15 and after. While I'm cleaning this up, I can sense plenty of nits suggestion coming... bring them on if I miss anything :D

FYI, things are now behind a flag.

While I'm implementing this I have a question for interaction for drag-and-dropping pinned tabs. Do we overwrite the ordering logic and overwrite "pinned time in chronological order" if a drag and drop event happens? I realized this problem when drag-and-dropping pinned tabs and realized the sorting logic basically cancel the drag and drop operation.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/87332917/158000123-4d996a23-5025-45f3-ad75-76cf261038df.mp4


Client/Frontend/Browser/Card/Card.swift
```diff
@@ -287,3 +303,31 @@ struct Card<Details>: View where Details: CardDetails {
...
+        } else {
+            content
+                .onDrag({
+                    tabCardDetail?.tabCardModel?.draggingDetail = tabCardDetail
+                    return NSItemProvider(contentsOf: URL(string: "\(tabCardDetail!.id)")!)!
+                })
+        }
+    }
+}
+extension View {
```
darinwf, line 329: nit: I'd probably leave out the `View` extension since it is not really saving on code. We only have one spot where we are creating the DragModifier.

Client/Frontend/Browser/Card/Card.swift
```diff
@@ -287,3 +303,31 @@ struct Card<Details>: View where Details: CardDetails {
         }
     }
 }
+
+struct TabDetailOnDragModifier: ViewModifier {
```
darinwf, line 307: nit: consider just defining this alongside ActionsModifier above within the definition of Card, since it is just a helper for Card. I'd probably then also just give it a shorter name like DragModifier.

Client/Frontend/Browser/Card/Card.swift
```diff
@@ -287,3 +303,31 @@ struct Card<Details>: View where Details: CardDetails {
...
+    var tabCardDetail: TabCardDetails?
+    var preview: AnyView
+
+    func body(content: Content) -> some View {
+        if #available(iOS 15, *) {
+            content
+                .onDrag {
+                    tabCardDetail?.tabCardModel?.draggingDetail = tabCardDetail
```
darinwf, line 315: Why did you choose to store `draggingDetail` on the `TabCardModel`? Why should `TabCardModel` need to know about UI interactions like drag-n-drop?

Is it that you were just looking for some global place to store it?

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -692,3 +720,9 @@ class TabGroupCardDetails: CardDetails, AccessingManagerProvider, ClosingManager
         }
     }
 }
+
+extension Array {
```
darinwf, line 724: This should live in an ArrayExtensions.swift file

Client/Frontend/Browser/Card/Card.swift
darinwf: Actually... the whole point of the `NSItemProvider` is to pass information to `dragEntered` and `dragUpdated` via the `DropInfo` parameter. You are storing the ID of the `TabCardDetail` that is being dragged there, so why not just extract the ID from there and look it up? Then there is no need to store draggingDetail anywhere.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -366,4 +366,8 @@ class TabManager: NSObject {
         assert(Thread.isMainThread)
         configuration.processPool = WKProcessPool()
     }
+
+    func rearrangeTabs(fromIndex: Int, toIndex: Int) {
+        tabs.rearrange(from: fromIndex, to: toIndex)
```
darinwf, line 371: I'm surprised you don't need to add a `tabsUpdatedPublisher.send()` here. I get that you would have already updated allDetails properly though, so maybe there should just be an optional notify parameter here that has a default value of true but that you pass false to when called from the drag-n-drop code. That way any future callers will not trip over the fact that this doesn't notify about changes to `tabs`.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -175,33 +176,60 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
...
 
-        let items = info.itemProviders(for: ["public.url"])
-        for item in items {
-            _ = item.loadObject(ofClass: URL.self) { url, _ in
-                if let url = url {
-                    DispatchQueue.main.async {
-                        self.manager.get(for: self.id)?.loadRequest(URLRequest(url: url))
-                    }
-                }
+    public func dropEntered(info: DropInfo) {
```
darinwf, line 194: nit: I think you should add comments for this function and `dropUpdated` to explain what they are for and why they are subtlety different.

darinwf (commented):
Works quite well! There are some things we can probably improve about the drag animation, but this is great.

Client/Frontend/Browser/Card/Card.swift
chassu: It seems that the item inside `DropInfo` can't be properly retrieved unless we're inside performDrop(), which is only called when the the user actually "dropped" the item. In order to show a preview of how the switcher is going to look like, we might still have to store it somewhere where we can read it from.

From the swiftUI documentation:
```

    /// This function is only valid during the `performDrop()` action.
    ///
    /// - Parameter contentTypes: The uniform type identifiers to query for.
    /// - Returns: The item providers that conforms to `contentTypes`.
    @available(iOS 14.0, macOS 11.0, *)
    public func itemProviders(for contentTypes: [UTType]) -> [NSItemProvider]

```

chassu: It seems to be working without having to make draggableDetail a published var though.

Client/Frontend/Browser/Card/Card.swift
```diff
@@ -287,3 +303,31 @@ struct Card<Details>: View where Details: CardDetails {
         }
     }
 }
+
+struct TabDetailOnDragModifier: ViewModifier {
+    var tabCardDetail: TabCardDetails?
+    var preview: AnyView
```
EdwardJTL, line 309: can you make this a view builder?

Client/Frontend/Browser/Card/Card.swift
```diff
@@ -206,6 +207,22 @@ struct Card<Details>: View where Details: CardDetails {
                             alignment: .top
                         )
                         .clipped()
+                        .if(dragNDrop && tabCardDetail != nil && !tabCardDetail!.isChild) { view in
```
EdwardJTL, line 210: does this need to be inside an if? this modifier changes the view hierarchy. is it possible to guard against these conditions elsewhere? maybe in `TabDetailOnDragModifier` which is defined in this file anyways?

Client/Frontend/Browser/Card/Card.swift
chassu: I added `@ViewBuilder` in the func body inside DragModifier. Is this what you're suggesting?

EdwardJTL: no, i meant that `preview` is declared as an `AnyView`
i think the body function automatically gets view builder because of inheritance, but you can leave that one

EdwardJTL: no, i meant that `preview` is declared as an `AnyView`
i think the body function automatically gets view builder because of inheritance, but you can leave that one

Client/Frontend/Browser/Card/Card.swift
EdwardJTL: there is a difference. the drag modifier checks on ios 15, which i believe is a compile time check.
the check with the `if` view modifier is a run time check. it will rebuild the view hierarchy when the check condition changes. 
i guess my question is, can we always apply the on drag view modifier?

chassu:
https://user-images.githubusercontent.com/87332917/158709113-66618f81-7794-40d5-ade5-43f012085dfc.MP4

chassu:
stunning.... @darinwf 👀 👀 🔥 🔥

chassu:
Raising an immediate issue that I see - since we're disabling drag-n-drop interaction to and from tab groups, if the auto-scroll will get terminated if there's a tab group and it'll never go above/below that row.

chassu:
Some code turns out not necessary after I tried on my iOS 15 testing device. The scroll looks smooth to me.

https://user-images.githubusercontent.com/87332917/159091689-b917b486-30b2-49c9-9f6a-206e9130415b.MP4

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -174,10 +178,11 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
 
     // Avoiding keeping a reference to classes both to minimize surface area these Card classes have
     // access to, but also to not worry about reference copying while using CardDetails for View updates.
-    init(tab: Tab, manager: TabManager, isChild: Bool = false) {
+    init(tab: Tab, manager: TabManager, isChild: Bool = false, tabCardModel: TabCardModel? = nil) {
```
darinwf, line 181: No longer any need to plumb `TabCardModel` if you use the static fields instead.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -133,7 +133,11 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
     public let id: String
     private var subscriptions: Set<AnyCancellable> = []
 
+    static var draggingDetail: TabCardDetails?
```
darinwf, line 136: nit: Looks like you can delete these static fields now.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -174,10 +178,11 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
 
     // Avoiding keeping a reference to classes both to minimize surface area these Card classes have
     // access to, but also to not worry about reference copying while using CardDetails for View updates.
-    init(tab: Tab, manager: TabManager, isChild: Bool = false) {
+    init(tab: Tab, manager: TabManager, isChild: Bool = false, tabCardModel: TabCardModel? = nil) {
         self.id = tab.id
         self.manager = manager
         self.isChild = isChild
+        self.tabCardModel = tabCardModel
```
darinwf, line 185: Looks like this needs to be a `weak` reference to avoid a reference cycle between `TabCardModel` and `TabCardDetails`

Client/Frontend/Browser/Card/CardDetails.swift
darinwf: I think this comment can be ignored since you are needing to plumb `TabCardModel` and are not using the static fields. Sorry for the confusing comment.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -133,6 +133,9 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
     public let id: String
     private var subscriptions: Set<AnyCancellable> = []
 
+    static var gridModel: GridModel?
```
darinwf, line 136: nit: This `gridModel` static field looks like it can just be deleted.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -32,6 +32,7 @@ class TabCardModel: CardModel {
     private(set) var allDetails: [TabCardDetails] = []
     private(set) var allDetailsWithExclusionList: [TabCardDetails] = []
 
+    var draggingDetail: TabCardDetails?
```
darinwf, line 35: Rather than define this as a field here, I think you should introduce a new type (probably a `struct`) named something like `DragState` that you use to hold all of the extra information needed when performing a drag-n-drop. The `DragState` instance can be stored as a static field on `TabCardDetails` just as you are doing for the other fields. When the drag-n-drop is done, you'd just clear that field. I think `DragState` could have two members: `tabCardModel` and `draggingDetail`.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -133,6 +133,11 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
     public let id: String
     private var subscriptions: Set<AnyCancellable> = []
 
+    struct DragState {
```
darinwf, line 136: nit: I think it would work better to store a static instance of DragState instead of DragState having a set of static fields. Also I think TabCardModel does not need to be weak in this case.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -196,25 +201,66 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
...
+        if fromIndex != toIndex {
+            manager.rearrangeTabs(fromIndex: fromIndex, toIndex: toIndex, notify: false)
         }
 
         return true
     }
 
+    // This function is called right when an item is dragged onto another item
+    public func dropEntered(info: DropInfo) {
+        guard let tabCardModel = TabCardDetails.DragState.tabCardModel,
```
darinwf, line 231: nit: when you have a static member, you can do `Self.dragState.tabCardModel`
You don't have to repeat the name of the class.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -133,6 +133,13 @@ public class TabCardDetails: CardDetails, AccessingManagerProvider,
     public let id: String
     private var subscriptions: Set<AnyCancellable> = []
 
+    struct DragState {
+        var tabCardModel: TabCardModel?
+        var draggingDetail: TabCardDetails?
+    }
+
+    static var dragState = DragState()
```
darinwf, line 141: nit: I recommend making this an optional so you can just assign it to `nil` when you want to remove it. That will work better if we ever add more fields in the future. It's just a more standard approach to this kind of thing.

darinwf (approved):
LGTM w/ nit

Issue #3069, opened 2022-03-12
Regression with navigating back to FastTap


darinwf:
I've started noticing that I cannot always go back to edit / revise my query in FastTap.

[Using 1.32.1]

Steps:

1. Type in "chocolate"
2. Hit enter
3. Hit back
4. Append " chip"
5. Hit enter
6. Hit back

Result:

You are now on the SRP for "chocolate" but you instead see FastTap pre-filled with the "chocolate chip" query.

If you hit back once more, you get to FastTap pre-filled with the "chocolate" query.

There's been some regression in our ability to handle query revision within a single tab. @evancrow FYI

(Tagging w/ release_blocker because this is a fairly serious usability issue.)

PR #3072, opened 2022-03-12
Fixes for Suggestion Back Button

evancrow:
The issue was caused by the search query not being set, so we never attached it to the navigation. Also added a test to prevent this from happening again!

### How was this tested?
- [ ] I tested this manually
- [x] I added automated tests

Video from test:

https://user-images.githubusercontent.com/37356984/158030025-e25ba839-8642-439f-b065-2491cdc794dd.mp4

## Issues addressed
Fixed: #3069


darinwf (approved):
LGTM, thanks for jumping on this!

PR #3073, opened 2022-03-12
Fix Preview Background Cutoff in Dark Mode

evancrow:
Before: 

Fixed:


darinwf (approved):
LGTM

Thanks for fixing this. I had just noticed this too and was about to file a bug on it. Thanks for beating me to it!

Issue #3080, opened 2022-03-13
All tabs get merged into a single tab group on cold start / via restoreTabs


darinwf:
Looks like this is a regression from 71bb935e52a2f378593b3715e5bb4cae051eaeb0

PR #3082, opened 2022-03-13
Fix Giant Tab Group

evancrow:
Happened to see this bug earlier and noticed an issue was just filed. This bug was caused by Tab URLs being nil on restore, which when compared in `shouldCreateTabGroup` returned true. Now the URLs are unwrapped before being compared.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Added a bunch of tabs/tab groups and restarted the app, everything works as expected now.

## Issues addressed
Fixed: #3080


Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -158,8 +158,13 @@ extension TabManager {
     @discardableResult func addTabToTabGroupIfNeeded(
         newTab: Tab, possibleChildTab: Tab
     ) -> Bool {
-        let childTabOriginalURL = possibleChildTab.originalURL?.normalizedHostAndPathForDisplay
-        let newTabURL = newTab.url?.normalizedHostAndPathForDisplay
+        guard
+            let childTabOriginalURL = possibleChildTab.originalURL?.normalizedHostAndPathForDisplay,
```
darinwf, line 162: btw, it occurs to me that we probably need to record `originalURL` to SavedTab, right?

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -158,8 +158,13 @@ extension TabManager {
...
+            let newTabURL = newTab.url?.normalizedHostAndPathForDisplay
+        else {
+            return false
+        }
+
```
darinwf, line 167: Should we set a flag to ensure that we only promote at most one `possibleChildTab` to be in the tab group? It seems like it could be possible to add more than one tab into the tab group if many happen to match, right? Not sure that is what we want to do. Maybe it should just merge with the first one that is found.

darinwf (approved):

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
evancrow: It's a computed property so we don't need to! Also wouldn't be possible.

darinwf: Oh right!

darinwf (approved):

PR #3085, opened 2022-03-14
Use url encoding to receive urls from the config file

### LINKED ISSUE #3083: Web3 object is nil because the url can't be read from config
Looks like #3036 was not properly working. This makes all transactions unusable and also balances not show...

Taking a look. @chung-lai FYI re release. (Backup solution is to revert, but I think this should be an easy fix today/tomorrow)

yozuysal:
The current implementation was producing valid URLs (which was what was tested before), but they were all "https" only. Using url encoding in the string and then removing url encoding avoid losing the string in the process.

Note: I also updated the relevant doc for the build as a part of this. If we have any other copies that we use for build verification, we should update those too.

### How was this tested?
- [X] I tested this manually
- [ ] I added automated tests

Confirm that not only we have a valid url, but also a valid web3 as a result of read from config.

## Issues addressed
Fixes #3083


chung-lai:
Not sure I fully understand the problem here. Is that when we get the value from the info plist, the string gets modified? Do we need to change what we set in the config file?

chung-lai (commented):
Left a comment

yozuysal:
Yes, without the url encoding, what is returned is just "https" and nothing else. I did modify the doc as well and added the url encoded values.

yozuysal:
So we need https%3A%2F%2Fmainnet.infura.io/....... instead of https://mainnet.infura.io/.............. inside the file

chung-lai:
> So we need https%3A%2F%2Fmainnet.infura.io/....... instead of https://mainnet.infura.io/.............. inside the file

Ah, got it!! Ok, lets do this for now to  unblock and I will revisit it and do something like what this thread suggested: https://stackoverflow.com/questions/21317844/how-do-i-configure-full-urls-in-xcconfig-files

chung-lai (approved):

PR #3087, opened 2022-03-14
Add SpaceEntity types

### LINKED ISSUE #3007: Add PreviewEntity Note
Right now section title and notes (Space entities with no url) are special cased inside SpaceDetailList. This usually causes issues with feature compat and we keep having to maintain code twice. We should

1) Add a new PreviewEntity for Note that will cover these cases
2) Move this code to inside SpaceEntityDetail

buracules:
- Added images and contents from list as separate files. Also i removed markdown snippet from SpaceDetailList
- Created SpaceListContainerView and put the body of SpaceEntityDetailView there

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
* Fixes #3007


Client/Frontend/Browser/Spaces/SpaceListContentView.swift
```diff
@@ -0,0 +1,226 @@
...
+        .buttonStyle(.reportsPresses(to: $isPressed))
+        .padding(.vertical, 12)
+        .padding(.horizontal, 16)
+        .background(
+            shouldHighlightAsUpdated
+                ? Color.ui.adaptive.blue.opacity(0.1) : Color.DefaultBackground)
+        Spacer(minLength: 0)
+    }
+
+    @ViewBuilder private var thumbnailView: some View {
```
yozuysal, line 145: Here is an idea: How about we make these three Views of their own that take base types and the previewEntity enum only? That way we can build previews for those Views as well.

yozuysal: So SpaceEntityTitleView, SpaceEntityThumbnailView and SpaceEntityDescriptionView. And SpaceEntityDescriptionView takes snippet as a String for example

yozuysal (commented):
This is actually looking pretty good as is. One suggestion re organizing things a bit more.

The only tricky thing to check out here is spaceLink and whether that is still working or not.

Client/Frontend/Browser/Spaces/SpaceListContentView.swift
buracules: Actually i was thinking about the same thing. Yes i can do that. They will only depend previewEntity enum only and the description will take extra snippet as string.  I will also add previews.

Client/Frontend/Browser/Spaces/SpaceEntityTitleView.swift
```diff
@@ -0,0 +1,50 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Defaults
+import Shared
+import SwiftUI
+
+struct SpaceEntityTitleView: View {
```
yozuysal, line 9: nit: This is a misnomer actually since there is more than the title inside. SpaceEntitySummaryView?

Client/Frontend/Browser/Spaces/SpaceEntityDescriptionView.swift
```diff
@@ -0,0 +1,56 @@
...
+            case .techDoc(let doc) = previewEntity,
+            let body = doc.body
+        {
+            Text(AttributedString(body))
+                .withFont(.bodyLarge)
+                .modifier(DescriptionTextModifier())
+        } else if let snippet = snippetToDisplay,
+            showDescriptions, !snippet.isEmpty
+        {
+            if #available(iOS 15.0, *),
```
yozuysal, line 37: Actually I think this is SpaceMarkdownSnippet with showDescriptions = true

yozuysal (approved):
lgtm with 2 nits and a small modification.

PR #3090, opened 2022-03-14
Avoid Force Unwrapping in CardModels.buildRows filter

EdwardJTL:
Saw quite a few crash report on this force unwrapping here. Not sure if the force unwrapping is intended; if not, this removes it.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
I didn't test this. Not sure how to reproduce a tab for which `TabManager.get(for:)` fails.

### DIFF
--- a/Client/Frontend/Browser/Card/CardModels.swift
+++ b/Client/Frontend/Browser/Card/CardModels.swift
@@ -124,7 +124,9 @@ class TabCardModel: CardModel {
 
         var partialResult: [Row] = []
         var allDetailsFiltered = allDetails.filter { tabCard in
-            let tab = tabCard.manager.get(for: tabCard.id)!
+            guard let tab = tabCard.manager.get(for: tabCard.id) else {
+                return false
+            }
             return
                 (tabGroupModel.representativeTabs.contains(tab)
                 || allDetailsWithExclusionList.contains { $0.id == tabCard.id })


darinwf:
Hmm, this was not supposed to be possible with the latest changes. Can you point me to some of the crash dumps? Can you confirm the crashes are in a build that includes 09a509a32797ef78e32ea5906882797692a3a460?

EdwardJTL:
> Hmm, this was not supposed to be possible with the latest changes. Can you point me to some of the crash dumps? Can you confirm the crashes are in a build that includes [09a509a](https://github.com/neevaco/neeva-ios/commit/09a509a32797ef78e32ea5906882797692a3a460)?

Screen Shot 2022-03-14 at 1 05 16 PM

sorry i'm not sure how to share a crash log from organizer

darinwf:
The issue here is caused by the fact that we haven't called `TabCardModel.onDataUpdated` yet, which runs from `TabManager.tabsUpdatedPublisher`. The sequence is: `removeTabs` -> `selectTab` -> `tabsUpdatedPublisher`. During `selectTab`, we notify code that then calls `buildRows`.

darinwf:
@chassu and I discussed a proper fix for this. He's going to take a stab at it.

Issue #3088, opened 2022-03-14
Tab group created through new navigation to hub site breaks existing tab group


darinwf:
Steps to repro:

1. Start with no tabs
2. Open a tab to news.ycombinator.com
3. Click on a link
4. Open a new tab to news.ycombinator.com
5. Click on a link
6. Open a new tab to news.ycombinator.com

Expected outcome:
Should see one tab group with 3 tabs in it

What happens instead:
See one tab group with 2 tabs in it and another tab on its own

What's going on:
When looking for a tab with matching originalURL, we assign the first tab we find to a new tab group with the newly opened tab. That has the effect of removing the existing tab out of its existing tab group.

Instead we should:
If a matching tab is already in a tab group, then just add the newly opened tab to that tab group.

One of the challenges here is that at the TabManager + Tab level, the only indication of a Tab being in a group is its rootID property. I think the only way to tell if it is not a solo tab is to check if other tabs share the same rootID property.

@evancrow FYI

yozuysal:
TabGroupManager keeps a running map of rootID to children. Maybe we need to move the ownership of TabGroupManager to TabManager so that this query is easily accessible?

darinwf:
@yozuysal - That sounds like a great idea!

darinwf:
I see... TabGroup and TabGroupManager are defined in BrowserPrimitive.swift, so perhaps we'd extract the core bits of each and leave behind a set of extensions that implement AccessingManager, ClosingManager, etc. Makes sense.

evancrow:
Or we could pass TabGroupManager to TabManager and leave it as a separate class? It's defined in BVC so it would be easy to pass on `init`.

evancrow:
Another option would be to declare `TabGroupManager` inside `TabManager` (and pass `TabManager` as a `weak` reference to avoid circular dependencies). I kind of like that one since we can cut `BVC` out of the picture.

darinwf:
I still see this bug in TestFlight build 1.32.6 (192.1). @chung-lai FYI

PR #3098, opened 2022-03-14
TabGroup Hot Fix

evancrow:
I somehow found a way to break everything this week!

### How was this tested?
- [ ] I tested this manually
- [x] I added automated tests

## Issues addressed
Fixed: #3088


Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -168,9 +168,17 @@ extension TabManager {
         }
 
         let shouldCreateTabGroup = childTabOriginalURL == newTabURL
+        let childTabIsParentOfTabGroup: Bool = {
+            for tabGroup in tabGroupManager.tabGroups.values
```
darinwf, line 172: I think the point of plumbing TabGroupManager here is to avoid needing to do an enumeration here. You should be able to query `tabGroupManager.tabGroups[possibleChildTab.rootUUID]` to efficiently see if `possibleChildTab` is a member of a tab group.

darinwf (commented):
Thanks for working on this so quickly!

However, I'm not so sure about plumbing TabGroupManager this way into TabManager. It creates a dependency from Tabs -> Card, which creates a circular dependency between those two modules. We should ideally avoid doing that.

So this is creating some technical debt. The right fix is to move TabGroupManager into Tabs/, and to factor out the BrowserPrimitive pieces into extensions as we have for TabManager.

A short term fix instead could also just be to enumerate all Tabs in TabManager and look for other tabs with the same rootUUID as possibleChildTab. That's a super simple fix and would be very isolated.

Then we can come back and do the proper refactoring I described above.

Makes sense?

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
evancrow: Oh okay, the RootID is the key to the dictionary

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -168,9 +168,16 @@ extension TabManager {
         }
 
         let shouldCreateTabGroup = childTabOriginalURL == newTabURL
+        let childTabIsParentOfTabGroup: Bool = {
+            for tab in tabs where tab.rootUUID == possibleChildTab.rootUUID {
```
darinwf, line 172: Hmm... does this work? `possibleChildTab` is part of `tabs`, so won't this always return true? I think we need to exclude `tab` if it is equal to `possibleChildTab`, right?

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -168,9 +168,16 @@ extension TabManager {
         }
 
         let shouldCreateTabGroup = childTabOriginalURL == newTabURL
+        let childTabIsParentOfTabGroup: Bool = {
```
darinwf, line 171: nit: Please add a comment mentioning how a better design would use TabGroupManager for more efficiently answering the question of if there are other tabs in the same group.

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -168,9 +168,16 @@ extension TabManager {
...
+            for tab in tabs where tab.rootUUID == possibleChildTab.rootUUID {
+                return true
+            }
+
+            return false
+        }()
 
         if shouldCreateTabGroup {
-            if possibleChildTab.parent == nil {
+            if possibleChildTab.parent == nil && !childTabIsParentOfTabGroup {
```
darinwf, line 180: By the way, I don't think we should be checking `parent` as that will not be set after session restore. That's a separate bug, but even still, `parent` is not meant to be the way check for inclusion in a tab group.

I think what you want to do here is answer the question, is the `possibleChildTab` all by itself, not part of a tab group, yet? If so, then form a new tab group with newTab as hub tab.

Else, make newTab be part of the same group as possibleChildTab.

So checking `parent` shouldn't happen. All we should do is check if `possibleChildTab` is part of a tab group already. And `childTabIsParentOfTabGroup` should be renamed to something that expresses that.

Neeva.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/apollographql/apollo-ios",
         "state": {
           "branch": null,
-          "revision": "d328c0ed3f50c856f8611f54191601a4d2be8533",
-          "version": "0.49.0"
+          "revision": "84c8fc2448d6b7285481bfa25da12668083859eb",
```
darinwf, line 9: whoops, probably want to avoid changing Package.resolved. Maybe this was just caused by Xcode when switching between branches and you happened to `git push` at a bad time?

darinwf (approved):

Neeva.xcworkspace/xcshareddata/swiftpm/Package.resolved
evancrow: Yeah I think that is what happened

XCUITests/TabGroupTests.swift
```diff
@@ -0,0 +1,69 @@
...
+
+class TabGroupTests: BaseTestCase {
+    // MARK: - NYTimes Test Case
+    // The NYTimes case is where a URL is opened from say nytimes.com,
+    // and a sublink i.e. nytimes.com/article is opened.
+    //
+    // Then the user would open up a new tab to the orignal URL (nytimes.com,
+    // which in that case we should create a tab group with the two tabs.
+
+    private func goToMoreInformation(tapLink: Bool = true) {
```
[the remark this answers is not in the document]
evancrow: Wrong PR? Doesn't really make sense in this case since the functions method goes to "More information..." on the `example.com` URL.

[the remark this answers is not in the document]
darinwf: I was picking a name that corresponds to what the intent of the function is. I was thinking it would help the tests read a bit better.

The reader does need to have in mind that openURLInNewTab is going to navigate a new tab to a particular URL. The `andNavigateAway` parameter was to highlight the intent to have the tab then navigate to a different URL, which is accomplished in this case by just clicking on the "More information" link.

XCUITests/TabGroupTests.swift
```diff
@@ -0,0 +1,71 @@
...
+    // The NYTimes case is where a URL is opened from say nytimes.com,
+    // and a sublink i.e. nytimes.com/article is opened.
+    //
+    // Then the user would open up a new tab to the orignal URL (nytimes.com,
+    // which in that case we should create a tab group with the two tabs.
+
+    /// Navigates to the test URL and if `andNavigateAway` is true, then it will tap a link
+    /// to open the sublink (similar to nytimes.com/article).
+    private func openTestURLInNewTab(andNavigateAway: Bool = true) {
+        openURLInNewTab()
```
darinwf, line 18: seems like openURLInNewTab should also be called openTestURLInNewTab too! but definitely for another PR ;-)

darinwf (commented):
Looks like we are missing a test case that covers the exact steps of the bug.

XCUITests/TabGroupTests.swift
```diff
@@ -0,0 +1,71 @@
...
+    func testNYTimesCaseCreatesTabGroup() {
+        openTestURLInNewTab()
+        openTestURLInNewTab(andNavigateAway: false)
+
+        goToTabTray()
+        waitForExistence(app.buttons["Tab Group, https://example.com/"])
+    }
+
+    /// Tests the case above, with multiple tabs.
+    func testNYTimesCaseCreatesTabGroupRepeated() {
```
evancrow, line 43: @darinwf Nope! I added this test which follows the exact same steps you laid out in the issue. Only difference is that it uses `example.com` and not `yCombinator`.

PR #3101, opened 2022-03-14
Show TopBar in CardGrid

evancrow:
In a recent change, we hid the `TopBar` when preview mode is being shown but it also hid the `TopBar` when opening the `CardGrid`.  Now we override that hiding if the `CardGrid` is visible.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
-  Delete the app and reinstall
- Close sign-in UI
- Go to `CardGrid`


darinwf (approved):
Good catch

PR #3102, opened 2022-03-14
Better URL String Parsing in SearchController

EdwardJTL:
Follow up to #3049 

I noticed that Mozilla has a more robust util for parsing URL from strings, which is why the url mentioned in #3049 could be opened when pasted in address bar but not be directly constructed to a URL object.

This change moves the two URL util files to shared module with appropriate access levels.

There is still a little bit of residue hard URL parsing in product list item sellers and reviews. Not quite sure what is going on in the logic there. Will take a look later.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Verified that links are parsed and can still be opened.


darinwf:
Hmm... usually URL fix-up is reserved for user input. Are we getting malformed URLs that need fix-up from the Neeva API endpoint for cheatsheet?

Generally good to limit fix-up to just the few cases where it is needed and usually just as a processing done to user input.

EdwardJTL:
> Hmm... usually URL fix-up is reserved for user input. Are we getting malformed URLs that need fix-up from the Neeva API endpoint for cheatsheet?
> 
> Generally good to limit fix-up to just the few cases where it is needed and usually just as a processing done to user input.

ah that's a good point.

yea with neeva endpoint we were getting URLs with `#` character in them, which must be put into percent encoding.
i suppose we can write a new class for just handling URL strings coming from neeva endpoints? However, we are no longer force unwrapping and instead just skipping the items with malformed URLs, I am concerned that we won't be able to discover new cases where edits are needed.

darinwf:
I see... the `URL(string:)` constructor is not able to handle input with a `#`? That's surprising since that is a valid character in a URL. It is not clear that percent encoding that is correct though. The fragment portion of the URL is important and if %-encoded, then the web page would not see it the same way. The page might not load properly, etc.

Do you have some more details on what URLs are failing to parse? Maybe share a few example URLs. Thanks!

EdwardJTL:
> I see... the `URL(string:)` constructor is not able to handle input with a `#`? That's surprising since that is a valid character in a URL. It is not clear that percent encoding that is correct though. The fragment portion of the URL is important and if %-encoded, then the web page would not see it the same way. The page might not load properly, etc.
> 
> Do you have some more details on what URLs are failing to parse? Maybe share a few example URLs. Thanks!

it was this url

https://apptronica.co.uk/www.cnet.com/home/kitchen-and-household/best-blenders#!=www.cnet.com-#listicle-9059a321-3809-49d6-9c48-68414d98b62d

when i stepped through `URIFixup.getURL(_:)`, it went into `punycodedURL()` and then `replaceHashMarks(url:)`

```
static func replaceHashMarks(url: String) -> String {
        guard let firstIndex = url.firstIndex(of: "#") else { return String() }
        let start = url.index(firstIndex, offsetBy: 1)
        return url.replacingOccurrences(of: "#", with: "%23", range: start..<url.endIndex)
    }
```

the resulting string was then parsed into a URL object and returned. and when i printed the contents of that URL object, it prints as # still

darinwf:
OK, I see... very interesting. I'm surprising `URL(string:)` would need the `#`'s chars %-encoded, but so be it. I would do a very focused fix just for the case you know this is needed since it is pretty exceptional to mutate the URL.

Shared/API/SearchController.swift
```diff
@@ -6,6 +6,47 @@ import Apollo
...
+
+extension URL {
+    fileprivate init?(parse string: String) {
+        if let url = URL(string: string), InternalURL.isValid(url: url) {
+            self = url
+            return
+        }
+
+        let trimmed = string.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard var escaped = trimmed.addingPercentEncoding(withAllowedCharacters: .URLAllowed) else {
```
darinwf, line 31: `URLAllowed` appears to include `"[]#"` ... maybe you should just define a variant of `URLAllowed` that excludes those characters? Then you would not need `replacing{Brackets,HashMarks}`.

I'm still quite surprised to learn that `URL(string:)` has trouble with these normal URL characters. I wonder if there is something we are missing about how `URL` works.

darinwf: I also think `URL(parse:)` is not very self-explanatory. The difference between this and `URL(string:)` is how it trims whitespace and percent-encodes certain characters. "parse" doesn't really convey that.

darinwf: Of course, you are just making these fileprivate so the name isn't so important. (I'd worry more about a global extension and broad confusion about which URL constructor to use.)

darinwf: Looks like this function could be useful: https://developer.apple.com/documentation/foundation/characterset/1780269-remove

EdwardJTL: looking at apple's [documentation](https://developer.apple.com/documentation/foundation/nsstring/1411946-addingpercentencoding), it seems like the most correct way to edit the string is to call `addingPrecentEncoding` with different character sets for different portions of the string; but that would require us to know which parts of the string represents which section.
`URL(string:)` returns `nil` when it finds illegal characters. In the example URL, there are two `#`'s. Looking at how the firefox routine works, it seems that only the first one is supposed to represent the fragment, and the second one should be percent encoded.

darinwf: Ok, that makes sense now. It being picky about the fragment needing to have `#`'s percent encoded does make sense. Maybe we should focus on that and not encode the first occurrence. Do we need to worry about trimming whitespace? Do we need to worry about the brackets? Were those just opportunistically included or do you know that they are a real issue for this code?

EdwardJTL:
we talked offline.
for this PR, we will limit to only replacing `#`, but we will log the presence of bad URL strings so we can pull the served URLs from the backend to investigate them later

darinwf (approved):
Lot of refactoring to support logging. Thanks for doing that!

PR #3107, opened 2022-03-15
make sure a new tab is selected tab when removing tab group

chassu:
There's a bug when removing tab group - when the selectedTab is not the last tab in the tab group, a new tab won't be selected, and clicking done button will show the deleted tab's webview.

### How was this tested?
- [x] I tested this manually

Before:

https://user-images.githubusercontent.com/87332917/158281905-e8687c3d-5e7c-44d1-bd1b-4ee8b55400e7.mp4

After:

https://user-images.githubusercontent.com/87332917/158281917-366fc884-cef0-4982-90f0-c807dd7db460.mp4


evancrow (approved):
LGTM

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -87,7 +87,7 @@ extension TabManager {
             DispatchQueue.main.async {
                 bvc.showTabTray()
             }
-        } else if tab == selectedTab {
+        } else if tab == selectedTab || tab.rootUUID == selectedTab?.rootUUID {
```
darinwf, line 90: I'm not sure I understand this check. It seems like one approach would be to just check if `selectedTab` is still part of `tabs`. If it is, then great. No work is needed. If it is not, then we need to update `selectedTab`. Simple and probably the most robust solution here.

```
} else if !tabs.contains(selectedTab) {
    // Update selectedTab
    ...
}
```

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -87,7 +87,7 @@ extension TabManager {
             DispatchQueue.main.async {
```
darinwf, line 87: Hmm... in this scenario, perhaps we should be setting selectedTab to nil? And shouldn't we also generate notifications for the selected tab changing?

chassu: I'm actually not sure what could break when we don't set it to nil in such scenario. Made some changes based on your comments though. Let me know if it looks right.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -85,9 +85,11 @@ extension TabManager {
 
         if closedLastNormalTab || closedLastIncognitoTab {
             DispatchQueue.main.async {
+                self.selectedTab = nil
+                self.selectedTabPublisher.send(nil)
```
darinwf, line 89: I think you need to honor `notify` here.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -85,9 +85,11 @@ extension TabManager {
...
                 bvc.showTabTray()
             }
-        } else if tab == selectedTab {
+        } else if let selectedTab = selectedTab, !viableTabs.contains(selectedTab) {
```
darinwf, line 92: Hmm... perhaps we should update the below to determine the next selected tab relative to the previously selected tab, instead of doing so relative to `tab`?

darinwf (commented):
This change deserves some test coverage.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
chassu: where are we passing the notify to?

darinwf: the `notify` parameter passed to this function is used to suppress notifications like the one you are sending here.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
darinwf: I think you should just call `selectTab(nil, notify: notify)` here instead

darinwf (changes requested):
See latest comments for what is needed.

darinwf (approved):

PR #3108, opened 2022-03-15
Only Updated Update Selected Tab, If No Tab Selected (RestoreTabs)

evancrow:
Pointed out by @chassu, but when restoring tabs it will override the selected tab which can be pretty annoying. Now will only pick the selected tab when restoring, if no tab is currently selected.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


chassu (approved):

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -25,7 +25,14 @@ class TabManager: NSObject {
 
     static var all = WeakList<TabManager>()
 
-    var tabs = [Tab]()
+    var tabs = [Tab]() {
+        didSet {
+            // Set the selected tab to nil if it was removed.
+            if let selectedTab = selectedTab, !tabs.contains(selectedTab) {
```
evancrow, line 31: Found a related issue where the `selectedTab` was not being reset if it was closed. Didn't cause any problems, but the state was still wrong.

darinwf: Shouldn't we just update the logic in `updateSelectedTabAfterRemovalOf` to handle whatever case is missing? It seems a bit confusing to have two places where we try to update the select tab after removing a tab. It also should be the case that we call `selectTab(nil)` when we want to set `selectedTab` to nil. That function takes care to store tabs state after changing which tab is selected. (Note, I think `selectTab(nil)` case might have some issues with updating incognito mode that would need to be addressed.)

darinwf (changes requested):

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -124,7 +124,10 @@ class TabCardModel: CardModel {
 
         var partialResult: [Row] = []
         var allDetailsFiltered = allDetails.filter { tabCard in
-            let tab = tabCard.manager.get(for: tabCard.id)!
+            guard let tab = tabCard.manager.get(for: tabCard.id) else {
```
darinwf, line 127: We have been trying to not add this change but to keep the force unwrapping to help us find cases where TabCardModel and TabManager are not in sync. Did you hit one of those cases as part of developing this PR?

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -83,6 +83,11 @@ extension TabManager {
...
         let bvc = SceneDelegate.getBVC(with: scene)
 
+        // Set the selected tab to nil if it was removed.
+        if let selectedTab = selectedTab, !tabs.contains(selectedTab) {
+            self.selectTab(nil)
+        }
+
         if closedLastNormalTab || closedLastIncognitoTab {
             DispatchQueue.main.async {
                 bvc.showTabTray()
```
darinwf, line 93: I think we should eventually move this logic to BVC and have it do this when it observes TabManager.tabs become empty, but that is for another change.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -83,6 +83,11 @@ extension TabManager {
         let viableTabs: [Tab] = tab.isIncognito ? incognitoTabs : normalTabs
         let bvc = SceneDelegate.getBVC(with: scene)
 
+        // Set the selected tab to nil if it was removed.
+        if let selectedTab = selectedTab, !tabs.contains(selectedTab) {
+            self.selectTab(nil)
```
darinwf, line 88: I think you might need to pass the `notify` parameter to `selectTab` here to avoid the crash that you were seeing in `buildRows`

darinwf (changes requested):

Client/Frontend/Browser/Card/CardModels.swift
evancrow: This is why the tests were failing. I you want to try it out, revert this change and run the `EmptyTabTrayTests`.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
evancrow: Ah okay!

darinwf:
@chassu - FYI, Maybe we should have made the `notify` parameter to `selectTab` a required parameter.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -83,6 +83,11 @@ extension TabManager {
         let viableTabs: [Tab] = tab.isIncognito ? incognitoTabs : normalTabs
         let bvc = SceneDelegate.getBVC(with: scene)
 
+        // Set the selected tab to nil if it was removed.
+        if let selectedTab = selectedTab, !tabs.contains(selectedTab) {
+            self.selectTab(nil, notify: false)
```
darinwf, line 88: I think you want to just forward the `notify` parameter here. That way the caller can control whether or not to allow notifications. Just like the other calls to `selectTab` below.

evancrow:
I'm going to close this PR, not really going anywhere, not really solving anything

PR #3126, opened 2022-03-15
Show default browser interstitial for "skip to browser" user

chung-lai:
* Show default browser interstitial to new users who skip to browser without sign in
* Add explicit skip to browser logging for analytics purpose
* Keep a new default flag to tell an user has seen the default browser interstitial via skip to browser flow so we can prompt the user again when they sign in

@darinwf, user still see the interstitial once. If they have seen the interstitial via skip to browser, they won't see it again when they sign in (but we can design for this case in the future as it uses separate flag for state)

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases

New user flow via skip to browser and sign in flow. They both should see the interstitial. Look at Proxyman for logging

  and dark mode

## Issues addressed


darinwf:
> @darinwf, user still see the interstitial once. If they have seen the interstitial via skip to browser, they won't see it again when they sign in (but we can design for this case in the future as it uses separate flag for state)

Sounds good!

darinwf (approved):

PR #3127, opened 2022-03-15
generate screenshot for new tabs via long press

chassu:
I played around and found out that when webview's frame is zero, the screenshot's frame will also be zero, which is causing the screenshot to be blank.

### How was this tested?
- [x] I tested this manually

Before:

https://user-images.githubusercontent.com/87332917/158483432-014a62d8-1030-4aba-91a1-b7c79f18ea43.mp4

After:

https://user-images.githubusercontent.com/87332917/158483435-cb1576bc-946b-4bb6-968d-5f8af0c5ff70.mp4


Client/Frontend/Browser/ScreenshotHelper.swift
```diff
@@ -34,6 +34,9 @@ class ScreenshotHelper {
         let configuration = WKSnapshotConfiguration()
         // This is for a bug in certain iOS 13 versions, snapshots cannot be taken correctly without this boolean being set
         configuration.afterScreenUpdates = false
+        // The screenshot's frame seems to be equal to webView's frame. The frame is manually
+        // set here to handle screenshot when a new tab is generated via long press menu.
+        webView.frame = controller?.tabManager.selectedTab?.webView?.frame ?? CGRect.zero
```
darinwf, line 39: Perhaps we should do this before calling takeScreenshot instead? Wouldn't it work to do this inside `BVC.navigateInTab`?

darinwf (approved):

PR #3136, opened 2022-03-16
Fix missed toggling create user activity flag

EdwardJTL:
Create user activity flag was default to true before so I didn't think to turn this on with the overwrite flag.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
it works :)


chung-lai:
@EdwardJTL, are we enabling spotlight for upcoming release? Wonder if we should cherry pick this.

EdwardJTL:
> @EdwardJTL, are we enabling spotlight for upcoming release? Wonder if we should cherry pick this.

no its okay. the flag in this PR was on by default. `main` will work fine for people who haven't been manually toggling these flags

Issue #3135, opened 2022-03-16
Suggested sites not showing up on compact state


chung-lai:
Suggested sites on ZeroQuery not showing up in compact state but shows up when it's fully expanded:

![IMG_5651](https://user-images.githubusercontent.com/87332578/158641361-b9b23bc0-0cb3-4c2b-9abf-98000d57bcfc.PNG)
![IMG_5652](https://user-images.githubusercontent.com/87332578/158641371-ef98bd3c-b07c-4a46-8b90-bd6ec23070c0.PNG)

Starts happening in 1.32.5. fyi @evancrow.

chung-lai:
Looking back at the set of changes, could it be related to #3067?

evancrow:
It seems to be related but I don't know how... suggested sites is inside a SwiftUI view which appears to be rendered just fine.

darinwf:
SnapKit was used by the horizontal scroller, so it could be a good one to explore.

Issue #3070, opened 2022-03-12
Regression: when removing a site from Suggested Sites, it is not visibly removed immediately


darinwf:
You have to open the zero-query UI again to see that the site was actually removed. It should instead refresh immediately.

This looks like a regression since I know this used to work :-(

jonathangarelick:
I'm definitely seeing this as well, but it's a bit tough to replicate.

PR #3137, opened 2022-03-16
Miscellaneous Fixes for Suggested Sites

evancrow:
### How was this tested?
- [ ] I tested this manually
- [x] I added automated tests

## Issues addressed
Fixed: #3135
Fixed: #3070


Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -240,15 +240,16 @@ class ZeroQueryModel: ObservableObject {
...
-        // if the default top sites contains the siteurl. also wipe it from default suggested sites.
+        // If the default top sites contains the site URL, also wipe it from default suggested sites.
         if TopSitesHandler.defaultTopSites().filter({ $0.url == url }).isEmpty == false {
             Defaults[.deletedSuggestedSites].append(url.absoluteString)
         }
+        
         profile.history.removeHostFromTopSites(host).uponQueue(.main) { result in
             guard result.isSuccess else { return }
             self.profile.panelDataObservers.activityStream.refreshIfNeeded(forceTopSites: true)
-            self.updateSuggestedSites()
```
chung-lai, line 251: Was this fixing suggested site not immediately removed from the screen? We don't need this anymore?

chung-lai (commented):
Just making sure I understand the change, we don't use autolayout for `FadingHorizontalScrollView` and let system to do autosizing?

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
evancrow: So what happened is that the actual updating of the `SuggestedSites` takes a while so it quickly returned the old values. So if we just delete it from the array instantly, and then update the array in the background for when the app actually updates it works.

evancrow:
> Just making sure I understand the change, we don't use autolayout for `FadingHorizontalScrollView` and let system to do autosizing?

We're still doing what we're always doing, using the UIKit layout constraints. But instead of attaching to the `superview` which is sometimes nil, we attach to the view it's actually embedded in.

chung-lai (approved):
LGTM, thanks!

PR #3138, opened 2022-03-16
Reduce Spotlight Logging Traffic

EdwardJTL:
Use Aggregated logging stats for indexing user activity and spaces. We lose some detail level information, but it doesn't impact our original goal of these logs was to understand how often we index.

We send these aggregated stats in the scene delegate.

Adding thumbnail and indexing errors are still logged per call. Aggregating these events will lose meaningful data.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Observed logs in proxyman

## Issues addressed


chung-lai:
Thanks @EdwardJTL for making this last minute change! 🙏

Client/Application/SceneDelegate.swift
```diff
@@ -519,4 +522,31 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
         return false
     }
+
+    func sendAggregatedSpotlightLogs() {
+        ClientLogger.shared.logCounter(
```
chung-lai, line 527: nit: may want to combine these three events into one and send the stats as attributes (e.g., `spotlight_stat` as the event name and attribute with `willIndexCount`, `userActivityCreationCount`, `didIndexCount`...), May be do this another time

Issue #3141, opened 2022-03-17
Landscape mode shows blank page with back button


chung-lai:
Starting from 1.32.0 build, landscape mode shows blank page with only a back button on my iPhone 13 Pro Max:

![IMG_5655](https://user-images.githubusercontent.com/87332578/158744209-2ad57f2c-b971-4800-b197-ad5417e30414.PNG)

Tapping on back button shows a screen takes half of the width:

![IMG_5656](https://user-images.githubusercontent.com/87332578/158744212-5e5a8f04-b463-4c7d-923b-884fc2643520.PNG)

chung-lai:
Looks like it only affects Pro Max phone/simulator. Other simulators work fine.

PR #3142, opened 2022-03-17
Quick NavigationView Fix

evancrow:
Just needed to tell the `NavigationView` to not use the large screen content type.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
- Use iPhone 13 Pro Max in Landscape, as well as other devices (including iPad)

## Issues addressed
Fixed: #3141


chung-lai (approved):
Thanks for the quick fix!

PR #3149, opened 2022-03-17
Add Fuzzy Matching Case for Wikipedia

EdwardJTL:
Per discussion, we add a special case for wikipedia's mobile host to normalize it by removing `.m.`

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
tested that we are filtering the url correctly in cheatsheet

## Issues addressed


darinwf (approved):
LGTM

darinwf:
It would be a good idea to include some unit tests for this though

darinwf:
> > It would be a good idea to include some unit tests for this though
> 
> are we fine with doing unit testing with hard coded urls as well? as opposed to actually loading a wikipedia url and comparing the redirected url to the input...

Yup! That's perfect. This is just about testing that our URL normalization code does what we intend, and it is about protecting that it continues to do so as others make changes to the code.

darinwf (approved):

PR #3150, opened 2022-03-17
run swift format after compile sources

chassu:
run swiftformat after compiling sources since we no longer let it fix the format for us

### How was this tested?
- [x] I tested this manually

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -4833,8 +4833,8 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = F84B21DD1A090F8100AAB793 /* Build configuration list for PBXNativeTarget "Client" */;
 			buildPhases = (
-				C9098F5F26B04DCE00D2B5F4 /* SwiftFormat */,
 				F84B21BA1A090F8100AAB793 /* Sources */,
+				C9098F5F26B04DCE00D2B5F4 /* SwiftFormat */,
 				F84B21BC1A090F8100AAB793 /* Resources */,
 				28CE83DE1A1D1E7C00576538 /* Frameworks */,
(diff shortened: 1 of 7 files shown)


EdwardJTL:
have you tried adding `--ignore-unparsable-files` option to the `swift-format` command in `Scripts/swift-format.sh`

EdwardJTL:
also re:description of the PR, i don't think we were letting swift-format format the files. we are just asking it to lint.

chassu:
I haven't tried `--ignore-unparsable-files` before, is it better to do so than swapping the order?

EdwardJTL:
> I haven't tried `--ignore-unparsable-files` before, is it better to do so than swapping the order?

yea i feel that running a linter after compiling is not very natural

chassu:
I'm not sure if I'm modifying in the right place... PTAL 🙏

EdwardJTL:
> 

that options needs to be passed to the actual swift format call

chung-lai:
> > I haven't tried `--ignore-unparsable-files` before, is it better to do so than swapping the order?
> 
> yea i feel that running a linter after compiling is not very natural

From what @chassu experienced, he was missing a curly brace in one file and that broke swiftformat so we may have hit a bug with swiftformat on unmatching braces. However, the linter was running before compilation and we couldn't see the actual compilation error. And imo, we usually run linter separately before commit and most of the time just let linter fix all the errors for us :). May be we can explore that given now we have swift-format circleci now?

EdwardJTL:
> 

when there's an unmatched braces, if the file could be parsed, the error log tells you where it is missing a brace (albeit not as nice as xcode highlighting).

and yes, i agree that we can take linter out of compile step. we can do a git hook or a separate circle ci job that will push to the branch?

PR #3151, opened 2022-03-17
Add a web3Mode mode flag to change post install and onboarding flows

yozuysal:
- Add a new flag that is distinct from enableCryptoWallet
- Skips signup/signin, default browser onboarding and shows a promo card for adding a wallet on zero query
- Adds neevaxyz suggestions on top of FastTap as a new suggestion type and fetches them when flas is on
- Changed default Suggested Sites to relevant ones for web3
- Removes immediate access to Spaces, and Neeva sign in related UI access points
- Shows this new zero query when there are no tabs instead of the tab tray
- Removes Neevascope button and swaps with Neeva Wallet button

This is intended to be a no-op when the flag is not enabled.

### How was this tested?
- [X] I tested this manually
- [ ] I added automated tests

Screen Shot 2022-03-17 at 12 22 11 PM


yozuysal:
Would be great to get both of your eyes on this as it involves onboarding+suggestions code and also information architecture modifications

chung-lai (commented):
Done a first pass on it, may have more comments

Client/Frontend/Suggestions/SuggestionModel.swift
```diff
@@ -49,6 +49,7 @@ class SuggestionModel: ObservableObject {
     @Published var urlSuggestions: [Suggestion] = []
     @Published var navSuggestions: [Suggestion] = []
     @Published private(set) var findInPageSuggestion: Suggestion?
+    @Published private(set) var xyzQuerySuggestions: [Suggestion] = []
```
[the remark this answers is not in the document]
yozuysal: Added .xyzQuery as the new suggestion enum type. Also added .XYZSearchSuggestion. The implementation un handleSuggestSelected is right below findInPage. Maybe I move it near .query to make it more prominent?

Client.xcodeproj/project.pbxproj
```diff
@@ -5399,7 +5402,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if [ -z \"$CI\" -a \"$CONFIGURATION\" = \"Debug\" ]; then\n    ./Scripts/swift-format.sh --check\nfi\n";
+			shellScript = "if [ -z \"$CI\" -a \"$CONFIGURATION\" = \"Release\" ]; then\n    ./Scripts/swift-format.sh --check\nfi\n";
```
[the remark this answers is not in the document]
yozuysal: oops this is trying to get my local build working... apologies.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -495,7 +495,11 @@ class BrowserViewController: UIViewController, ModalPresenter {
             } else if !Defaults[.didFirstNavigation] {
                 self.showPreviewHome()
             } else if self.tabManager.normalTabs.isEmpty {
```
darinwf, line 497: Preview mode also only makes sense if there are no tabs yet. The `didFirstNavigation` check is just to handle the case where someone opened some tabs but then closed them and now has no tabs. We don't want to show the preview experience in that case.

So, you could just move the `!Defaults[.didFirstNavigation]` check into the `normalTabs.isEmpty` block. That way you can only have to check `web3Mode` once, which is I think what you were achieving by suppressing the assignment to `didFirstNavigation`.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1595,85 +1599,85 @@ extension BrowserViewController {
...
-                        setTokenAndOpenURL(
-                            token: token, url: NeevaConstants.verificationRequiredURL)
-                    default:
-                        break
-                    }
+        let onDismiss: (FirstRunButtonActions) -> Void = { action in
+            self.introViewModel?.dismiss {
+                self.introViewModel = nil
 
-                    if NeevaUserInfo.shared.hasLoginCookie() {
```
darinwf, line 1635: I don't understand the motivation for this change.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1595,85 +1599,85 @@ extension BrowserViewController {
...
-                    case .oauthWithProvider(_, _, let token, _):
-                        // loading appSearchURL to prevent showing marketing site
-                        setTokenAndOpenURL(token: token, url: NeevaConstants.appSearchURL)
-                    case .oktaAccountCreated(let token):
-                        setTokenAndOpenURL(
-                            token: token, url: NeevaConstants.verificationRequiredURL)
-                    default:
-                        break
-                    }
+        let onDismiss: (FirstRunButtonActions) -> Void = { action in
```
darinwf, line 1602: Probably doesn't make sense to change any code inside this function. If you are trying to suppress the entire first run flow, then you'd want to avoid calling `showProperIntroVC` at all. I think it'd be better to do that then to conditionalize the body of this function.

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -118,7 +118,11 @@ class ZeroQueryModel: ObservableObject {
...
 
         if !Defaults[.signedInOnce] {
-            if Defaults[.didFirstNavigation] {
+            if FeatureFlag[.web3Mode] && Defaults[.cryptoPublicKey].isEmpty {
+                promoCard = .walletPromo {
+                    self.bvc.web3Model.showWalletPanel()
+                }
+            } else if Defaults[.didFirstNavigation] {
                 promoCard = .previewModeSignUp {
                     ClientLogger.shared.logCounter(
```
darinwf, line 127: It will be important to make sure that any logging uses a different ClientID so we don't mix all of the data together.

yozuysal: That should come after we actually have a different client ID though, right? I was thinking of taking care of logs after we have a new target/bundle ID etc

Client/Frontend/Browser/BrowserViewController.swift
yozuysal: this function should be the exact same impl (if not I missed something). But from the looks of it, the way our state machine is build for startup, we do need to call it for the mainContent to show.

Client/Frontend/Browser/BrowserViewController.swift
yozuysal: this is a bad rebase I think. thanks for catching it. Again, this dismiss call should be exactly as is if I am doing this right.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -485,7 +485,7 @@ class BrowserViewController: UIViewController, ModalPresenter {
         // that supports preview mode. They will have tabs already, so we don't
         // want to show them the preview home experience.
         // TODO: This is flawed as an existing user may have closed their tabs.
-        if !tabManager.tabs.isEmpty {
+        if !tabManager.tabs.isEmpty || FeatureFlag[.web3Mode] {
```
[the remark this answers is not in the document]
yozuysal: I had perceived didFirstNavigation as inPreviewMode from our current state. Is that not the case? I worry if we wrote the preview code with that assumption in mind, I will need to sprinkle over if(web3) to all of them and will miss some inevitably. Isn't that going to be more fragile?

Client/Frontend/Browser/BrowserViewController.swift
yozuysal: actually, I will circle back on this after checking something quick,. Maybe I was misinterpreting the way the states work here. I will revert all changes in BVC and see if I can minimize diff to a few lines here.

Client/Frontend/Browser/BrowserViewController.swift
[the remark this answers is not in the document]
darinwf: It's just a state we keep to know if a navigation has been performed. I think that's a useful state to keep. We use it to help us know if we should trigger preview mode, but it isn't mean to be a flag that says we should use preview mode. Preview mode has slightly more logic to its triggering (e.g., has no tabs).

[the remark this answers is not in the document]
darinwf: We might want to use this state for other things in the future, and so that's why i'm worried about disabling this state tracking in web3 mode. Certainly web3 mode can have a first navigation.

yozuysal:
@darinwf ended up reverting almost all BVC changes and replacing them with a cleaner way to simulate a dismiss. I do think didFirstNavigation = true is a more robust change though. Looking up the uses of it, I find 5-6 different places I have to sprinkle web3Mode conditions. It does seem to be the equivalent of isPreviewMode as far as the code is concerned now. PTAL

Client/Frontend/Browser/BrowserViewController.swift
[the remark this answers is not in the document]
yozuysal: But right now we do have a few `!Defaults[.didFirstNavigation] ? .previewHome : .blank` inside TabContainerHost and also we show the PreviewMode promo card if the user doesnt have this set.

I am OK with changing all` !Defaults[.didFirstNavigation] ` to `!Defaults[.didFirstNavigation] || FeatureFlag[.web3Mode]` but I worry this is gonna be more fragile long term.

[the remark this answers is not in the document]
yozuysal: Alright, I concede my point :) It wasn't as extensive and I ended up using it in few places.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1526,13 +1524,20 @@ extension BrowserViewController {
...
     ) {
         if alwaysShow || !Defaults[.introSeen] {
-            showProperIntroVC(
-                signInMode: signInMode,
-                onOtherOptionsPage: onOtherOptionsPage,
-                marketingEmailOptOut: marketingEmailOptOut,
-                completion: completion,
-                onDismiss: onDismiss
-            )
+            if FeatureFlag[.web3Mode] {
```
darinwf, line 1527: I think it is a bit hacky to intercept and override `presentIntroViewController`. We should back up the stack to where the decision is made to show the first-run experience (which is what this is about), and avoid doing it there. In this case, I think that's just `viewDidAppear`, and then it comes down to removing access points for signing into neeva.com, right?

(I'm a little confused about how this app uses neeva.com for search but hides the account details in the settings.)

yozuysal: Hiding the settings was part of my thinking for removing access points to sign into neeva.com. Should we keep that section in settings for now?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -492,16 +486,25 @@ class BrowserViewController: UIViewController, ModalPresenter {
...
+                }
             }
         }
     }
 
     override func viewDidAppear(_ animated: Bool) {
-        presentIntroViewController()
+        if FeatureFlag[.web3Mode] {
+            Defaults[.introSeen] = true
+            overlayManager.hideCurrentOverlay(ofPriority: .fullScreen) {}
```
darinwf, line 504: What overlay is this hiding? Hmm, do we really need to set `introSeen`? (It looks like introSeen is state internal to presentIntroViewController, but I may be missing something.)

yozuysal: Yeah, I think it is only for internal state within intro, we seem to be using it as a way to check "Is this first run?" while deciding within the intro though. My thinking was, if the user somehow sees this screen, it shouldn't behave as first run but should use the setup you would get when you run into it in daily use.

yozuysal: I had thought hideCurrentOverlay was a precaution to dismiss any existing overlays, but maybe that is actually hiding the intro fullscreen overlay in which case it is not needed.

darinwf: `introSeen` is just meant to indicate if we should show the first run experience, but since we are skipping it entirely in web3 mode, I don't think you need to worry about setting that state. And since the intro screen isn't seen, it is probably better to not set the state. We are going to be refactoring a bunch of this probably next week, so trying to keep the streams separated so don't need to worry about web3 cases when evolving the way first run works.

If there is no overlay to hide, then probably don't need to call hideCurrentOverlay yeah :)

chung-lai: We do use `introSeen` in the latest default browser experiment for signed out users to determine whether to show interstitial after first run screen. But that code only runs when user interacts with the first run screen which we don't show in wallet flow anyway.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1526,13 +1529,13 @@ extension BrowserViewController {
...
     ) {
         if alwaysShow || !Defaults[.introSeen] {
-            showProperIntroVC(
-                signInMode: signInMode,
-                onOtherOptionsPage: onOtherOptionsPage,
-                marketingEmailOptOut: marketingEmailOptOut,
-                completion: completion,
-                onDismiss: onDismiss
-            )
+                showProperIntroVC(
```
chung-lai, line 1532: nit: indentation?

chung-lai (commented):
After looking through the feature flag checks and knowing we will probably fork the app. Does it make sense we use compilation flag and start forking the code with `#if WALLET` and that will make a cleaner cut between wallet and browser app. (like using certain classes that won't be there in browser code)

In that case, may be we will at some point convert all these feature flags into compilation flag check. May be we haven't crossed that bridge yet, just food for thought.

Client/Frontend/Browser/BrowserViewController.swift
yozuysal: ok, removed all this. Now we simply dont call this in first run with the flag and not set/do anything else

Client/Frontend/Browser/BrowserViewController.swift
yozuysal: indent fixed.

re compilation flag vs feature flag, as we are talking through this I am leaning that way as well. This change is only to see what the experience would look like.

chung-lai (approved):

darinwf (approved):

PR #3154, opened 2022-03-17
Carry over query for navigation when opening from SRP

EdwardJTL:
When we open a result from SRP, the new navigation item doesn't get a query value associated in `QueryForNavigation`.
This change carries over the query value, for both opening in place and in a new tab. This way, a page opened from SRP will have the same query value as pages opened from fast tap, which gives better fallback for cheatsheet.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually checked by opening links from SRP, both in new tab and in place (by changing it in the account settings). Observed that query value is attached and correct.


darinwf:
Does this interfere with back navigation to the SRP at all?

EdwardJTL:
> Does this interfere with back navigation to the SRP at all?

oh darn, it looks like `Tab.goBack` uses the query to check if it should open the suggestion UI instead... that makes things complicated :/

darinwf:
> > Does this interfere with back navigation to the SRP at all?
> 
> oh darn, it looks like `Tab.goBack` uses the query to check if it should open the suggestion UI instead... that makes things complicated :/

Yeah. that's what i was worried about. It is okay, we just need to add an attribute on the stored query to reflect the mode.

EdwardJTL:
> > > Does this interfere with back navigation to the SRP at all?
> > 
> > 
> > oh darn, it looks like `Tab.goBack` uses the query to check if it should open the suggestion UI instead... that makes things complicated :/
> 
> Yeah. that's what i was worried about. It is okay, we just need to add an attribute on the stored query to reflect the mode.

what should we call this attribute? I believe we need it to indicate whether suggestion UI should be opened for the current navigation item.
or we could have cheatsheet model look for query using the tab's backlist and parent properties, but the parent tab property is not restored on launch, so this won't work sometimes.

darinwf:
> what should we call this attribute? I believe we need it to indicate whether suggestion UI should be opened for the current navigation item. or we could have cheatsheet model look for query using the tab's backlist and parent properties, but the parent tab property is not restored on launch, so this won't work sometimes.

Be creative :) What would you like to call it? I think you have the right idea about what the flag should indicate.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -349,6 +349,16 @@ class TabManager: NSObject {
             self.selectTab(popup)
         }
 
+        // if we open from SRP, carry over the query
```
darinwf, line 352: So this is for the scenario of a user who has configure neeva to open links in a new tab, right?

darinwf (approved):

chung-lai (approved):

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
EdwardJTL: yes. the other one in `QueryForNavigation` is for opening in place

Client/Frontend/Browser/Tabs/QueryForNavigation.swift
```diff
@@ -3,15 +3,21 @@
 // found in the LICENSE file.
 
 import Foundation
+import Shared
 import WebKit
 
 class QueryForNavigation {
     struct Query {
+        enum Location: Int, Codable {
+            case suggestion
```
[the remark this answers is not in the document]
EdwardJTL: added comments :)

Issue #3155, opened 2022-03-17
Large memory usage when we index space items for spotlight


chung-lai:
Noticed a few memory crash when I used the testflight app 1.32.10 and saw one in my log with `cannot allocate memory`. Looking at simulator with memory usage pattern, notice that after using the app for a while, the memory keeps going up from 150mb to 1.9gb in my case. 

Putting a breakpoint in some spotlight indexing code, every time the memory spikes, that piece of code got triggered (https://github.com/neevaco/neeva-ios/blob/main/Shared/API/SpaceStore%2BCoreSpotlight.swift#L94-L97)

@darinwf, this may explain some crashes you experience earlier in the week without crash stack. fyi @EdwardJTL

EdwardJTL:
@darinwf when you looked for on device logs for the mysterious crashes, did you check if there's any jetsam events logged during the time period of interest?
App killed by memory might be the reason why there was no testflight crash report.

chung-lai:
Didn't know about jetsam before, that's cool! Somehow I don't see any got generated from today, may be I got a crash report instead.

chung-lai:
Remove the release_blocker label as that should be addressed by the PR. Will leave this open until we find a fix for the thumbnail or feel free to open another issue if you like, @EdwardJTL . Thanks for fixing it so quickly!

PR #3156, opened 2022-03-17
Skip Adding Thumbanil Data to Spotlight Spaces URLs

EdwardJTL:
Repeated calls to getThumbnailData seems to grow the memory without releasing objects, a potential cause for #3155.

Skipping setting and fetching thumbnail data for now. Will investigate why the memory is growing.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
@chung-lai and I observed that after this change the memory seems to be no longer growing at a problematic rate.

## Issues addressed
a temporary fix for #3155


chung-lai (approved):
After applying this change, the memory is stable so the changes are good. @darinwf, by applying this change, space entries won't show a thumbnail in spotlight. imo, I think it's ok for initial launch and have a quick follow up to fix it. Flagging it in case this is a feature blocker.

darinwf (approved):
I did see another instance of the app crashing randomly yesterday and without generating a crash report. I think that is the fingerprint of an out of memory crash, and I did just re-enable spotlight integration before that (the other day).

I'm glad you found this issue!

PR #3160, opened 2022-03-18
Remove remainder of legacy Download Manager

### LINKED ISSUE #3153: Remove option to clear "Downloaded Files" from Settings > Clear Browsing Data
This is leftover from how Firefox has a Downloads manager. We just put downloads into the user visible file system managed via the Files app, and so we let the Files app be the way the user manages their downloads.

jonathangarelick:
This PR removes the remainder of the FF iOS Download Manager, including the option to clear downloaded files in Settings > Clear Browsing Data

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases

## Issues addressed
* Fixes #3153


jonathangarelick:
@darinwf It might make sense to just merge this, then create a separate issue for testing download functionality later. What do you think?

Client/Frontend/Settings/Privacy/DataManagementView.swift
```diff
@@ -12,7 +12,6 @@ enum ClearableDataType: String, Identifiable, Codable, CaseIterable {
     case cache = "Cache"
```
jonathangarelick, line 12: calling this out -- expand to line 9 -- "removing an entry from this enum will cause the user’s settings to be cleared"

jonathangarelick:
@darinwf This is ready for review. There's a comment in DataManagementView that says removing an entry from the ClearableDataType enum clears the user's settings, but I can't seem to replicate it.

Client/Frontend/Settings/Privacy/DataManagementView.swift
```diff
@@ -59,7 +55,7 @@ enum ClearableDataType: String, Identifiable, Codable, CaseIterable {
 extension Defaults.Keys {
     fileprivate static let clearDataTypes = Defaults.Key<Set<ClearableDataType>>(
         "profile.dataManagement.clearTypes",
-        default: Set(ClearableDataType.allCases).filter { $0 != .downloads })
+        default: Set(ClearableDataType.allCases))
```
darinwf, line 58: Yeah, I can see how there might be a data migration issue. I don't know how an enum type is serialized by UserDefaults. I can imagine that changing the set of possible enum cases would change the schema in such a way that it would not be backwards compatible to remove an enum case. I'll do a quick test to double check...

darinwf: OK, yeah, seems to work!

darinwf (approved):

Issue #3143, opened 2022-03-17
Tab switcher display issue after rotating the phone (Pro Max phones)


chung-lai:
Now the issue is happening on tab switcher after switching between landscape and portrait. In landscape, the tab switcher is taking half of the screen. And the layout is pretty mess up in portrait. This happens after the fix for #3141.

![IMG_5657](https://user-images.githubusercontent.com/87332578/158752812-dbc5fcba-e532-4ba9-8c27-e2060522d944.PNG)

![IMG_5658](https://user-images.githubusercontent.com/87332578/158752822-c19ef427-e0d4-48e1-b72b-8e9150a72219.PNG)

fyi @evancrow

evancrow:
@chung-lai I saw this once, but cleaning the build file fixed it for me. Can you do the same?

chung-lai:
> @chung-lai I saw this once, but cleaning the build file fixed it for me. Can you do the same?

Just tried cleaning the build and removing the derived folder but it doesn't help on simulator. And this was happening on the Testflight build.

darinwf:
I think we have to re-open this bug now that we reverted the PR for it.

PR #3161, opened 2022-03-18
*new* Rotation Fix

evancrow:
Fixed: #3143

@chassu If you could make sure the TabGroup bug you found is not present that would be great!


darinwf (approved):
LGTM

I did a bit of test driving as well, and it seems solid. Thank you @evancrow !

PR #3170, opened 2022-03-19
make notify required in selectTab

chassu:
Making notify a required argument in selectTab, as suggested from https://github.com/neevaco/neeva-ios/pull/3108

### How was this tested?


darinwf (approved):
Thanks for doing this!

PR #3167, opened 2022-03-18
move to xcode13.3

EdwardJTL:
Move CircleCI to Xcode 13.3
For compatibility with xcode 13.3:
- Move swift-format to new [release](https://github.com/apple/swift-format/releases/tag/0.50600.0) 
- Move apollo to [0.51.0](https://github.com/apollographql/apollo-ios/issues/2146)

Verified that circle ci still builds and runs swift-format

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests
- [x] Existing automated tests are enough to cover my changes

PR #3182, opened 2022-03-21
Add Place API to Search Query

EdwardJTL:
This is to implement this type of results in cheatsheet
Screen Shot 2022-03-21 at 5 08 42 PM

This change depends on #3167 because I am working with xcode 13.3 and needs a new version of apollo for codegen

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,7 +143,7 @@ commands:
 jobs:
   build:
     macos:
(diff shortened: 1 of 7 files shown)


EdwardJTL:
this needs to be put back to xcode 13.2

EdwardJTL:
see #3187

PR #3183, opened 2022-03-21
Update Apollo Codegen to Use schema.json

EdwardJTL:
After updating to xcode 13.3 and using apollo 0.51.x (see #3167), the initializer we were using for apollo code gen options defaults to `schema.graphqls`. 
This change initializes the options exactly like the initializer we were calling, except it specifies schema.json 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
this is how i made #3182


darinwf (approved):
Makes sense!

PR #3184, opened 2022-03-21
Use interstitial screen for the education screen

chung-lai:
As the new interstitial screen with instructions combined perform better and have better UX. We are changing the existing education screens with the new interstitial.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases

* Settings *

Screen Shot 2022-03-21 at 1 45 41 PM

* Promo Card *

Screen Shot 2022-03-21 at 1 45 50 PM

## Issues addressed


hwiatneeva (commented):
LGTM from design. SWE should approve the code changes. 

Thanks much, Chung!!!!

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1693,7 +1693,9 @@ extension BrowserViewController {
                 } buttonAction: {
                     self.overlayManager.hideCurrentOverlay()
                     UIApplication.shared.openSettings(
-                        triggerFrom: .defaultBrowserPromptMergeEduction
+                        triggerFrom: fromSkipToBrowser
+                            ? .defaultBrowserPromptSkipToBrowser
+                            : .defaultBrowserPromptMergeEduction
```
darinwf, line 1698: I can see it getting a bit complicated when we want to convey state like this through a series of actions. Probably keeping it simple as you have done is the way to go for now. Makes me wonder if there is a better way though.

Client/Frontend/DefaultBrowserOnboarding/DefaultBrowserIntestitialOnboardingView.swift
```diff
@@ -6,8 +6,74 @@ import Defaults
 import Shared
 import SwiftUI
 
+class DefaultBrowserInterstitialOnboardingViewController: UIHostingController<
```
darinwf, line 9: We should probably be using the OverlayManager these days. Probably a question for Evan though.

darinwf (approved):

Client/Frontend/Browser/BrowserViewController.swift
chung-lai: Yes, agree. Would love to spend a bit of time to think about the state as the default browser interstitial stablize.

Client/Frontend/DefaultBrowserOnboarding/DefaultBrowserIntestitialOnboardingView.swift
chung-lai: Yes, I tried a few options with OverlayManager and couldn't get one with full sheet presented. Let me add a TODO to figure this out with Evan about whether we have support on that.

PR #3188, opened 2022-03-22
Wallet theme support

buracules:
- Add Web3Toolbar
- Add Web3Theme enum for support multiple themes
- Add OpenSea asset support


Client/Frontend/NeevaWallet/Web3Theme+Extensions.swift
```diff
@@ -0,0 +1,161 @@
...
+
+import SFSafeSymbols
+import Shared
+import SwiftUI
+import Defaults
+import WalletCore
+import SDWebImageSwiftUI
+import UIKit
+
+extension Web3Theme {
```
yozuysal, line 13: can this live inside WalletCore?

WalletCore/AssetStore.swift
```diff
@@ -46,8 +49,13 @@ public class AssetStore: ObservableObject {
                 self.assets = result.assets
                 self.assets.forEach({
                     guard let collection = $0.collection else { return }
+                    if let theme = Web3Theme.allCases.first(
+                        where: { $0.rawValue == collection.openSeaSlug }) {
+                        self.availableThemes.insert(theme)
+                    }
                     self.collections.insert(collection)
                 })
+                print(self.availableThemes.count)
```
yozuysal, line 58: nit leftover print? If we wanna log, we can use Logger.browser

Client/Frontend/TabChrome/Toolbar/TabToolbarButtons.swift
```diff
@@ -41,16 +43,29 @@ enum TabToolbarButtons {
...
-                    .arrowBackward, size: 20, weight: weight,
-                    label: .TabToolbarBackAccessibilityLabel),
-                action: onBack,
-                longPressAction: onLongPress
-            )
-            .disabled(!model.canGoBack && !model.canReturnToSuggestions)
-            .accessibilityAction(named: "Show Recent Pages", onLongPress)
+            if FeatureFlag[.web3Mode] {
+                TabToolbarButton(
+                    label: Web3Theme(with: currentTheme).backButton,
```
yozuysal, line 52: let's do:

```
@ViewBuilder var label: some View {
    if FeatureFlag[.web3Mode] {
        // Web3Theme....    
    } else {
        //default logic
    }
}
```

above so that we dont have to copy over anything else. Same for all buttons

yozuysal (commented):
looking pretty good. a few comments about structure

yozuysal (approved):
one last nit. Thank you!

yozuysal:
One last thing before we land. We might want to update the PR description here to give more context

yozuysal:
one thing I noticed while playing with this. We should import the CC assets as pdf, then render them as template and set foregroundColor(.label). This way it will work with dark mode better.

yozuysal:
another thing I justy caught on. When we remove the wallet (Inside WalletDashboard), we need to cleanout anything related in defaults

buracules:
> one thing I noticed while playing with this. We should import the CC assets as pdf, then render them as template and set foregroundColor(.label). This way it will work with dark mode better.

I can render them as template while they are png is there any another reason for pdf? Also i think we have to update all assets they all have different spacing etc.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -102,6 +103,8 @@ extension BrowserViewController: ToolbarDelegate {
                 self.showTabTray()
             case .share:
                 self.showShareSheet(buttonView: self.view)
+            case .showZeroQuery:
+                self.showZeroQuery()
```
yozuysal, line 107: actually this should be openLazyTab instead of showZeroQuery. you can call with no arguments.

yozuysal:
let's ship it!

yozuysal (approved):
taking this. I am unsure about the swiftformat errors being related with this change and this is too large to contain random swift format fixes imo.

PR #3189, opened 2022-03-22
Add "Search with Neeva" to Shortcuts and Siri Suggestions

EdwardJTL:
- Edit Search with Neeva shortcut item (the display strings)
- When we submit text to the url bar
  - add the shortcut to shortcut suggestions (can be viewed in Shortcuts app under Gallery)
  - donate the interaction (can be viewed by turning on "Display Recent Shortcuts" and "Display Donations on Lock Screen" in your phone's developer's settings)

@darinwf  we could put the search text in the donated object, which will open the same search in a new tab when it is opened; my understanding is that our goal is to open the app on a new tab with this suggestion.

putting this out now to get some feedback

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Do a search, verify that a short cut suggestion is created, and that a siri suggestion was donated.
Open the shortcut, verify that it opens a blank search.
Open the suggestion, verify that it opens a blank search.

With the shortcut, provide a search text, run shortcut, verify that it opens a new search with that query


Client/Application/SceneDelegate.swift
```diff
@@ -250,12 +250,15 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
-        {
-            self.bvc.openURLInNewTab(url)
-            return true
+        if let intent = userActivity.interaction?.intent as? SearchNeevaIntent {
+            if let query = intent.text,
+               let url = SearchEngine.current.searchURLForQuery(query)
+            {
+                self.bvc.openURLInNewTab(url)
+                return true
+            } else {
```
darinwf, line 259: nit: no need for `else` after `return`. less indenting can be helpful.

Client/Info.plist
```diff
@@ -96,12 +96,12 @@
 	<string>Websites you visit may request your location.</string>
 	<key>NSUserActivityTypes</key>
 	<array>
-		<string>co.neeva.app.ios.browser.browsing</string>
 		<string>IntentHandler</string>
 		<string>IntentIntent</string>
 		<string>NSUserActivityTypeBrowsingWeb</string>
 		<string>OpenURLIntent</string>
 		<string>SearchNeevaIntent</string>
+		<string>co.neeva.app.ios.browser.browsing</string>
```
darinwf, line 104: Is this reordering significant?

darinwf (approved):

Client/Info.plist
EdwardJTL: its what xcode automatically did after i modified the SearchNeevaIntent file. i think it will build either way; but if xcode is going to keep generating this change, we might as well check it in

Client/Application/SceneDelegate.swift
```diff
@@ -264,7 +264,15 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
                 self.bvc.openURLInNewTab(url)
             } else {
                 // open a new search
-                self.bvc.openLazyTab()
+                // Wait for animations that run when returning to foreground to finish
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) { [self] in
```
darinwf, line 268: Can we avoid hard-coded delays like this? These lead to flakiness in behavior. Is there a trigger like an `onAppear` that we can use instead?

EdwardJTL: i copied this from `QuickActions.handleOpenNewTab`. https://github.com/neevaco/neeva-ios/pull/2082
i removed the delay and it works mostly fine (the suggested sites sometimes jitter).

PR #3200, opened 2022-03-23
Add Impression Timers to cheatsheet education screens

EdwardJTL:
Add two logging cases, one for on SRP and one for on non-SRP pages
2 seconds impression (i find that 1 second may give false results due to how our overflow menu destroys the view)

Also refactored info view to new file since menu view file is getting too long

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
verified that logs show up


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetInfoView.swift
```diff
@@ -0,0 +1,184 @@
...
+                attributes: EnvironmentHelper.shared.getAttributes()
+            )
+            buttonAction()
+        }
+        .onAppear {
+            ClientLogger.shared.logCounter(
+                .ShowCheatsheetEducationOnSRP,
+                attributes: EnvironmentHelper.shared.getAttributes()
+            )
+            impressionTimer?.invalidate()
```
chung-lai, line 29: We are doing the 2 second impression logging elsewhere in the code. Feel free to generalize the `PromoCardImpressionModifier` from `ZeryQueryView` for this purpose :)

Client/Frontend/ZeroQuery/ZeroQueryView.swift
```diff
@@ -125,14 +125,12 @@ struct PromoCardImpressionModifier: ViewModifier {
 
     let attributeValue: String
 
-    func startImpressionTimer(with attributeValue: String) {
```
chung-lai, line 128: Don't think we need the `impressionTimer` anymore in `PromoCardImpressionModifier`. Indeed, we don't even need `PromoCardImpressionModifier` here and we can directly use the `ImpressionLoggerModifier` within the promo card view itself

Client/Logging/ImpressionLoggerModifier.swift
```diff
@@ -0,0 +1,49 @@
...
+// found in the LICENSE file.
+
+import Foundation
+import Shared
+import SwiftUI
+
+struct ImpressionLoggerModifier: ViewModifier {
+    @State var impressionTimer: Timer? = nil
+
+    let timeInterval: TimeInterval
```
chung-lai, line 12: We can default this interval to 2 seconds and let people override this interval if they want. That way, this should bring more consistent across app when no one overrides this :P

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetInfoView.swift
```diff
@@ -0,0 +1,164 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Shared
+import SwiftUI
+
+private let impressionOnSRPTimeInterval: TimeInterval = 2
```
chung-lai, line 8: if we make default to be `2` in `ImpressionLoggerModifier` then we don't need to specify this constant.

chung-lai (commented):
Thanks for refactoring the impression timer. left a few comments

Client/Logging/ImpressionLoggerModifier.swift
EdwardJTL: done, i used the default in zero query view, but i kept my own for cheatsheet because the overlay thing is finicky.

chung-lai:
```suggestion
    let timeInterval = defaultTimeInterval
```

And this variable will be optional?

EdwardJTL: but if i do that, i still need to keep the initializer around, else it wont compile properly if we pass a value to override it, so i feel that it suffices to put the default value as args to the initializer

chung-lai (approved):
LGTM

PR #3203, opened 2022-03-23
Ensure switcher shows the correct state

### LINKED ISSUE #3191: Tab switcher can at times incorrectly indicate the wrong state
I've seen a few times now where the tab switcher control will indicate normal tabs but I'm seeing incognito tabs, or it will show spaces, but indicate normal tabs. Here's a video of it having just happened:

https://user-images.githubusercontent.com/79867437/159585536-bd649c1d-e064-459a-bf69-a115815f030c.MP4

darinwf:
I've noticed that it is possible to not get an `.onEnded` notification
from a drag operation. I can repro by dragging around the top of the
device and causing the OS UI to kick in.

This change ensures that whenever we enter the tab switcher, the
drag offset of the switcher is reset. This way it will be updated to
show the correct state.

Also addresses an issue with `selectTab(nil)` that could result in
forcing the user out of incognito mode, which is not intended in our
UX design. We don't call `selectTab(nil)` so this was avoided, but
this change just helps remove this landmine for the future.

Fixes #3191

### How was this tested?
- [x] I tested this manually


chung-lai (approved):
LGTM

PR #3204, opened 2022-03-23
bye, tabGroupCardModel and TabGroupManager

chassu:
Merge tabGroupManager with tabManager and tabGroupCardModel with tabCardModel. Hopefully, there will be performance improvement by combining the two and reducing the updates caused by `onDataUpdated()` and the 
 various subscriptions.

### How was this tested?


chassu:
I will work on a PR that aims to simplify `buildRows`s logic and process the rows in onDataUpdated() and feed it to SingleLevelTabCardsView directly.

darinwf:
Seeing an odd thumbnail bug:



I started with a single cnn.com tab, and then opened a new tab to hacker news. I clicked on a story, hit back. Then hit "show tabs" and saw this.

darinwf:
Not able to repro :-(

Client/Frontend/Browser/Card/BrowserPrimitive.swift
```diff
@@ -296,8 +296,8 @@ struct TabGroup {
     var id: String
```
darinwf, line 296: The definition of `TabGroup` should also be moved to Browser/Tabs.

We want to avoid circular dependencies between Browser/Tabs and Browser/Card

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -567,17 +567,19 @@ class SiteCardDetails: CardDetails, AccessingManagerProvider {
 }
 
 // TabGroupCardDetails are not to be used for storing data because they can be recreated.
-class TabGroupCardDetails: CardDetails, AccessingManagerProvider, ClosingManagerProvider,
-    ThumbnailModel
-{
+class TabGroupCardDetails: ObservableObject {
     typealias Item = TabGroup
-    typealias Manager = TabGroupManager
+    typealias Manager = TabManager
```
darinwf, line 572: Are these type aliases needed?

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -24,14 +24,21 @@ protocol CardModel: ThumbnailModel {
...
     private(set) var allDetailsWithExclusionList: [TabCardDetails] = []
 
+    // Tab Group related members
+    private(set) var representativeTabs: [Tab] = []
+    private(set) var allTabGroupDetails: [TabGroupCardDetails] = [] {
+        didSet {
+            allTabGroupDetailsWithExclusionList = allTabGroupDetails
+        }
+    }
+    private(set) var allTabGroupDetailsWithExclusionList: [TabGroupCardDetails] = []
```
darinwf, line 40: Why do we need this variable if it is just a clone of allTabGroupDetails?

chassu:
I did remember seeing similar thumbnail bug but thought one of the change fixed it... 😞

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -259,7 +269,13 @@ class TabCardModel: CardModel {
     }
 
     func onDataUpdated() {
-        groupManager.updateTabGroups()
+        manager.tabGroups = manager.getAll()
```
darinwf, line 272: Updating tabGroups and childTabs should be done by TabManager. It is backwards for a downstream consumer to have to maintain these fields for TabManager.

Whenever we want to calls tabsUpdatedPublisher.send(), we should before doing so, update these fields of TabManager.

I think there is a caller outside of TabManager that pokes tabsUpdatedPublisher.send() directly. That would need to change.

darinwf: I think it's the code that removes a tab from a tab group that pokes this publisher directly.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -27,6 +27,11 @@ class TabManager: NSObject {
     var tabs = [Tab]()
     var tabsUpdatedPublisher = PassthroughSubject<Void, Never>()
 
+    // Tab Group related variables
+    @Default(.tabGroupNames) private var tabGroupDict: [String: String]
+    var tabGroups: [String: TabGroup] = [:]
+    var childTabs: [Tab] = []
```
darinwf, line 33: I wonder if we can get away with making childTabs be a computed property instead.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -269,21 +285,29 @@ class TabCardModel: CardModel {
...
         }
         .map { TabCardDetails(tab: $0, manager: manager) }
 
+        // Tab Group related updates
+        representativeTabs = manager.getAllTabGroup()
+            .reduce(into: [Tab]()) { $0.append($1.children.first!) }
+        allTabGroupDetails = manager.getAllTabGroup().map {
+            TabGroupCardDetails(tabGroup: $0, tabManager: manager)
+        }
+        manager.cleanUpTabGroupNames()
```
darinwf, line 298: This should also just be done by the manager before calling tabsUpdatedPublisher.send().

This method should just be private to TabManager in fact.

darinwf (commented):
Good stuff!

Client/Frontend/Browser/Card/CardModels.swift
darinwf: Well, we should look at all of those spots and try to understand why they are calling `tabsUpdatedPublisher.send()` directly. Instead, it should be a method on `TabManager` that they are calling that mutates `TabManager` and those methods should know to generate the `tabsUpdatedPublisher.send()` call. Or maybe there is code that is modifying a `Tab`. Maybe those mutating functions should be expressed as a method on `TabManager` so we can control how updates are generated.

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -567,17 +567,17 @@ class SiteCardDetails: CardDetails, AccessingManagerProvider {
 }
 
 // TabGroupCardDetails are not to be used for storing data because they can be recreated.
-class TabGroupCardDetails: CardDetails, AccessingManagerProvider, ClosingManagerProvider,
-    ThumbnailModel
-{
-    typealias Item = TabGroup
-    typealias Manager = TabGroupManager
+class TabGroupCardDetails: ObservableObject {
```
darinwf, line 570: Note, we'll eventually want this to have a different name for this at some point since it no longer "is a CardDetails"

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -751,3 +752,47 @@ class TabWebView: WKWebView, MenuHelperInterface {
         super.evaluateJavaScript(javaScriptString, completionHandler: completionHandler)
     }
 }
+
+struct TabGroup {
```
darinwf, line 756: nit: this deserves its own TabGroup.swift file

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -751,3 +752,47 @@ class TabWebView: WKWebView, MenuHelperInterface {
...
+    var children: [Tab]
+    var id: String
+}
+
+extension TabGroup: BrowserPrimitive {
```
darinwf, line 761: Avoid circular dependencies between Browser/Tabs and Browser/Card. Card depends on Tabs, so we cannot have `BrowserPrimitive` here. Maybe there is no reason for TabGroup to inherit from BrowserPrimitive. Bet you can just delete this dependency and many of the fields below. Please do a pass to eliminate those that are unnecessary.

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -150,7 +150,7 @@ extension TabManager {
         }
 
         if notify {
-            tabsUpdatedPublisher.send()
+            updateTabGroup(notify: notify)
```
darinwf, line 153: Hmm... updateTabGroup does more than send notification events. That said it is nice to avoid extraneous calls to updateTabGroup, so this is probably a good fit. Probably should rename updateTabGroup to something more descriptive like `updateTabGroupsAndSendNotifications`

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -27,6 +27,17 @@ class TabManager: NSObject {
     var tabs = [Tab]()
     var tabsUpdatedPublisher = PassthroughSubject<Void, Never>()
 
+    // Tab Group related variables
+    @Default(.tabGroupNames) private var tabGroupDict: [String: String]
+    var tabGroups: [String: TabGroup] = [:]
+    var childTabs: [Tab] {
+        getAllTabGroup().flatMap(\.children)
+    }
+    var representativeTabs: [Tab] {
```
darinwf, line 36: I wonder if there are any performance concerns with computing this one.

darinwf (commented):
So beautiful :-)

Just a few more bits and bobs...

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: I think it does because it's used in `buildRows()` and `buildRows()` is spammy now

Client/Frontend/Browser/Tabs/Tab.swift
chassu: I can see that it's unnecessary for TabGroup to inherit from BrowserPrimitive, but I don't see any circular dependencies that came with it. Could you explain more?

darinwf: `BrowserPrimitive` is defined in `Browser/Card`. `TabGroup` is now defined in `Browser/Tab`.

`Browser/Card` uses `Browser/Tab`, so we should avoid making `Browser/Tab` use `Browser/Card`. That way we avoid a circular dependency. The codebase should be organized as a directed graph.

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Combine
+import Defaults
```
darinwf, line 6: nit: looks like you can drop this now

darinwf (approved):

PR #3208, opened 2022-03-24
Add NetworkManager to WalletCore

buracules:
- Created a NetworkManager for WalletCore
- Created 2 APIs (OpenSeaAPI and CurrencyStoreAPI)

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
I only tested with `.get` methods. if we wanna expand this manager it's better we use something like `Alamofire` otherwise we have to develop every encoding from scratch.


WalletCore/Network/Web3NetworkProvider.swift
```diff
@@ -0,0 +1,57 @@
...
+        target: URLRequestBuilder,
+        model: T.Type,
+        completion: @escaping (
+            Result<T, Error>
+        ) -> Void)
+}
+
+public class Web3NetworkProvider {
+
+    public static let `default`: Web3NetworkProvider = {
```
yozuysal, line 18: should we use lazy?

yozuysal (commented):
looks great overall! One q about initialization then we are good to go

yozuysal (approved):

WalletCore/Network/Web3NetworkProvider.swift
buracules: It says `'lazy' cannot be used on an already-lazy global` 😀

WalletCore/AssetStore.swift
```diff
@@ -55,8 +55,9 @@ public class AssetStore: ObservableObject {
                     })
                     self.state = .ready
                     if result.next?.isEmpty == false {
-                        self.refresh(cursor: result.next!)
+                        self.refresh(cursor: result.next!, onCompletion: onCompletion)
                     }
+                    onCompletion?()
```
yozuysal, line 60: we should only run onCompletion when we didn't get any next cursor in the result. Otherwise this will run it multiple times for every paginated fetch

yozuysal (approved):

WalletCore/Network/APIs/OpenSeaAPI.swift
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    public var headers: HTTPHeaders {
+        switch self {
+        case .assets, .collections:
+            if let apiKey = Bundle.main.object(forInfoDictionaryKey: "OPENSEA_API_KEY") as? String {
+                return [
+                    "Accept": "application/json",
+                    "X-API-KEY": apiKey,
```
yozuysal, line 35: I dont think we need this for collections btw

WalletCore/AssetStore.swift
```diff
@@ -19,84 +19,98 @@ public class AssetStore: ObservableObject {
...
-                string:
-                    "https://api.opensea.io/api/v1/assets?order_direction=desc\(cursor == nil ? "" : ("&cursor=" + (cursor?.addingPercentEncoding(withAllowedCharacters: .alphanumerics))!))&limit=50&owner=\(Defaults[.cryptoPublicKey])"
-            )
-        else {
+    public func refresh(cursor: String? = nil, onCompletion: (() -> Void)? = nil) {
+        guard !Defaults[.cryptoPublicKey].isEmpty else {
             return
         }
-        if cursor == nil {
-            assets = []
```
yozuysal, line 32: we might want to make sure we reset this below or in some other way

yozuysal (changes requested):

yozuysal (approved):
we need to add limit=50 and we are good to go.

PR #3211, opened 2022-03-24
Add Swift format to git hook

EdwardJTL:
- Add git pre-push hook to check for swift format on the current state of the repo
- Remove swift format step from client build phases
- Add bootstrap code to link the githooks

the script is based off https://stackoverflow.com/a/61621713
except we don't need to do this check per commit, we can just run it on the current state of the repo.

## Discussion
- This allows us to only check on the files that are committed. errors in the un-committed, modified files won't prevent you from pushing.
  - The downside is that because we stash, you won't be able to undo in xcode :( 

- The alternative is to not stash and check directly on the current status of the repo. but this means your uncommitted changes must also be clean. I feel that this is inferior because you will likely want to stash before fixing the linter error anyways
 
### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested
- No errors, push a commit
- Errors in a previous commit, cannot push
- Errors in uncommitted changes, can push

## Issues addressed


.githooks/pre-push
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# An example hook script to verify that each commit that is about to be pushed
```
darinwf, line 3: nit: This comment looks like it is residual from an example script. Would be good to clean up.

darinwf (approved):
Looks great!

Is there a way to force push a change even if it has lint errors?

EdwardJTL:
> Looks great!
> 
> Is there a way to force push a change even if it has lint errors?

yea you can run 'git push --no-verify' (i added this to the comment)

PR #3214, opened 2022-03-25
Unwrap Reader mode URLs

EdwardJTL:
- Unwrap reader mode URLs for cheatsheet

- Also use `tab.url` which strips authorization

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
tested that it works on reader mode pages


darinwf (approved):
LGTM

There's also error page URLs to unwrap. Those can be done using `InternalURL`.

EdwardJTL:
> LGTM
> 
> There's also error page URLs to unwrap. Those can be done using `InternalURL`.

I took a look at internalURLs, but I wasn't sure how to unwrap those.
we have seen both `internal://local/errorpage?url=` and `internal://local/sessionrestore?history` urls on Mode

PR #3215, opened 2022-03-25
Plumbing to add apple oauth authentication code to /login-mobile calls

seth737:
After completing the native Apple oauth flow, we were previously only sending back the `identityToken` to the `/login-mobile` endpoint. We passed this as the `serverAuthCode` URL param.

With this change, we:

1. Rename the `serverAuthCode` URL param to `identityToken`, for clarity (the server API will still accept either param, preferring `identityToken` over `serverAuthCode` if both are set)
2. Add the authentication code to this request, as the new `authenticationCode` URL param.

The authentication code is used by httpd to exchange with Apple's servers for an authentication token, which is then encrypted and stored in mysql. A daily job (usertable-cleanauth) runs to verify the token has not been revoked by the user. See this doc for more details: https://docs.google.com/document/d/1F8Kwpexb481eIOBXxVuK9noBVCR8cHfZQR7HL6g06IA.

I also removed some legacy logic that was storing the identity token in a cookie, after verifying no code in httpd reads this cookie.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases

  and dark mode

## Issues addressed

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1657,8 +1657,11 @@ extension BrowserViewController {
                     self.introViewModel = nil
 
                     switch action {
-                    case .signupWithApple(let marketingEmailOptOut, let identityToken, let authorizationCode):
-                        if let identityToken = identityToken, let authorizationCode = authorizationCode {
+                    case .signupWithApple(
+                        let marketingEmailOptOut, let identityToken, let authorizationCode):
+                        if let identityToken = identityToken,
+                            let authorizationCode = authorizationCode
+                        {
                             let authURL = NeevaConstants.appleAuthURL(
                                 identityToken: identityToken,
                                 authorizationCode: authorizationCode,
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1657,19 +1657,14 @@ extension BrowserViewController {
                     self.introViewModel = nil
 
                     switch action {
-                    case .signupWithApple(let marketingEmailOptOut, let serverAuthCode):
-                        if let serverAuthCode = serverAuthCode {
+                    case .signupWithApple(let marketingEmailOptOut, let identityToken, let authorizationCode):
+                        if identityToken != nil && authorizationCode != nil {
                             let authURL = NeevaConstants.appleAuthURL(
-                                serverAuthCode: serverAuthCode,
+                                identityToken: identityToken!,
+                                authorizationCode: authorizationCode!,
                                 marketingEmailOptOut: marketingEmailOptOut ?? false,
                                 signup: true)
-                            let httpCookieStore = self.tabManager.configuration.websiteDataStore
-                                .httpCookieStore
-                            httpCookieStore.setCookie(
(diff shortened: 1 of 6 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1657,19 +1657,14 @@ extension BrowserViewController {
                     self.introViewModel = nil
 
                     switch action {
-                    case .signupWithApple(let marketingEmailOptOut, let serverAuthCode):
-                        if let serverAuthCode = serverAuthCode {
+                    case .signupWithApple(let marketingEmailOptOut, let identityToken, let authorizationCode):
+                        if identityToken != nil && authorizationCode != nil {
```
darinwf, line 1661: nit: The more canonical Swift way of writing this is like so:

```
    if let identityToken = identityToken, let authorizationCode = authorizationCode {
        ...
```

This has the effect of unwrapping those optional variables, and then as a result you don't need the `!` suffix.

Client/Frontend/Intro/IntroViewModel.swift
```diff
@@ -199,20 +199,31 @@ extension IntroViewModel: ASAuthorizationControllerDelegate {
...
-                    if Defaults[.signedInOnce] {
-                        ClientLogger.shared.logCounter(
-                            .SignInWithAppleSuccess,
-                            attributes: EnvironmentHelper.shared.getFirstRunAttributes()
-                        )
-                    }
-                    self.onDismiss(.signupWithApple(self.marketingEmailOptOut, authStr))
-                }
+            guard let identityToken = appleIDCredential.identityToken else {
+                print("Unable to fetch identity token")
```
darinwf, line 203: nit: Instead of using `print` statements, use `Logger.auth` like so:

At the top of the file at global scope after the import statements add this:

```
private let log = Logger.auth
```

Then, in place of these `print` statements, use `log.error`.

By using `Logger` you can control whether logging goes to the console or to a file or if it should just be discarded. That way you can opt-into the logging you are interested in seeing and you can even get these logs from a production build that is misbehaving.

Client/Frontend/Intro/SignUpTwoButtonsPromptView.swift
```diff
@@ -130,21 +130,34 @@ struct SignUpTwoButtonsPromptView: View {
...
-                                    signup: true)
-                                ClientLogger.shared.logCounter(
-                                    .PreviewPromptSignupWithApple,
-                                    attributes: EnvironmentHelper.shared.getFirstRunAttributes())
-                                hideOverlay()
-                                openInNewTab(authURL, false)
-                            }
+                        // TODO(Seth): ideally reduce the code duplication here with IntroViewModel lines 200-217
+                        guard let identityToken = appleIDCredential.identityToken else {
+                            print("Unable to fetch identity token")
```
darinwf, line 135: same nit about using `Logger.auth`

darinwf (approved):

darinwf (approved):

darinwf:
To make `swift-format` happy just run this command from the `neeva-ios` directory:
```
$ ./Scripts/swift-format.sh
```

darinwf (approved):

PR #3220, opened 2022-03-25
Add a new app target

yozuysal:
This target contains all Client files, but excludes all extensions and the app clip.

It logs with the new bundle id. In a followup I will also override client and browser header as well.


Client.xcodeproj/project.pbxproj
```diff
@@ -5396,6 +6032,39 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "Conditionally Add Settings Bundle";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if [ \"${INCLUDE_SETTINGS_BUNDLE}\" = \"YES\" ]\nthen\n    cp -r \"${PROJECT_DIR}/${TARGET_NAME}/Application/Settings.bundle\" \"${BUILT_PRODUCTS_DIR}/${PRODUCT_NAME}.app\"\nfi\n";
```
chung-lai, line 6066: Do you recall adding anything new for these INCLUDE_SETTINGS_BUNDLE and swift-format phase? Or XCode add them automatically?

chung-lai (approved):
LGTM, just left one comment

Client.xcodeproj/project.pbxproj
```diff
@@ -7763,7 +7763,7 @@
 					"$(inherited)",
 					"$(SDKROOT)/usr/lib/swift",
 				);
-				MARKETING_VERSION = 0.0.1;
+				MARKETING_VERSION = 1.34.3;
```
yozuysal, line 7766: @chung-lai , do you think we will need to keep these to be the same?

chung-lai: Without this change, it will break the version update build script. Will address this in a separate PR as thinking what would be the best way to update version in the build script.

PR #3226, opened 2022-03-25
Tolerate nil pinnedTime

### LINKED ISSUE #3091: Crash @ [ TabGroupCardDetails.init ] Pinned tab may be causing a crash
Screen Shot 2022-03-14 at 12 55 36 PM

crash on `TabGroupCardModel.onDataUpdated()`->`map`->`init`->`init`. 
This seems to come from `TabGroupCardDetails.init`
checking this init routine,  we have force unwrapping `lhs.pinnedTime! < rhs.pinnedTime!`

we have a variable `isPinned: Bool` and `pinnedTime: TimeInterval?`. 

the simple fix for the crash itself is to use `if let pinnedTime` instead of `isPinned`. But really these should be more tightly coupled together to avoid them being out of sync. So perhaps we should use an enum with an associated value, or just use `@Published pinnedTime: TimeInterval?` and maybe a computed variable to check isPinned; I anticipate this will call for changes on the SwiftUI side as well.

darinwf:
Fixes #3091


chassu (approved):
LGTM

PR #3232, opened 2022-03-28
Move BuildRows inside onDataUpdated

chassu:
@darinwf, what do you think about this direction of change? It's yielding some weird result now, but theoretically I think it makes sense..


chassu:
This is definitely not ready for review yet. Sorry I misclicked..

darinwf:
> This is definitely not ready for review yet. Sorry I misclicked..

I clicked the "Convert to draft" link to mark it as such. Just click "Ready for review" when true.

chassu:
Seeing huge improvement. Gonna rebase from main and see how it looks along with browser view improvements.

chassu:
Did profiling before and after, timed from launching the app to seeing the card grid, after clicking the tab switcher

before:

Screen Shot 2022-03-27 at 11 08 00 PM

after:
Screen Shot 2022-03-27 at 11 10 15 PM

Client/Frontend/Browser/Card/CardGrid.swift
```diff
@@ -75,7 +76,11 @@ struct CardGrid: View {
 
     @State private var detailDragOffset: CGFloat = 0
     @State private var cardSize: CGFloat = CardUX.DefaultCardSize
-    @State private var columnCount = 2
+    @State private var columnCount = 2 {
+        didSet {
+            tabCardModel.updatePartialResult(incognito: incognitoModel.isIncognito, columnCount: columnCount, notify: true)
```
darinwf, line 81: I think you should avoid making CardGrid depend on IncognitoModel. Adding the `@EnvironmentObject` above is not strictly necessary. You should only do that for models that influence the SwiftUI View tree. You just need access to read the current value of the IncognitoModel.isIncognito state. TabCardModel already has access to this information via TabManager.isIncognito.

darinwf: I would call this method `updateColumnCount` to be more specific about what it is for.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -27,15 +27,27 @@ class TabCardModel: CardModel {
     private var subscription: Set<AnyCancellable> = Set()
 
     private(set) var manager: TabManager
+    private(set) var partialResult: [Row] = []
```
darinwf, line 30: nit: probably just call this `rows` to be more aligned with the data type and how it is used. It is not always a partial result :)

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -27,15 +27,27 @@ class TabCardModel: CardModel {
...
     private(set) var allDetails: [TabCardDetails] = []
     private(set) var allDetailsWithExclusionList: [TabCardDetails] = []
+    private var columnCount: Int = 2
+    private var incognito: Bool = false
```
darinwf, line 34: can just use manager.isIncognito

Client/Frontend/Browser/Card/SingleLevelTabCardsView.swift
```diff
@@ -23,8 +23,7 @@ struct SingleLevelTabCardsView: View {
 
     var body: some View {
         ForEach(
-            tabModel.buildRows(
-                incognito: incognito, maxCols: columns.count)
+            tabModel.partialResult
```
darinwf, line 26: Nice!!!!

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -27,15 +27,27 @@ class TabCardModel: CardModel {
...
     // Tab Group related members
     private(set) var representativeTabs: [Tab] = []
     private(set) var allTabGroupDetails: [TabGroupCardDetails] = []
 
     @Default(.tabGroupExpanded) private var tabGroupExpanded: Set<String>
 
+    func updatePartialResult(incognito: Bool, columnCount: Int, notify: Bool) {
+        if self.incognito != incognito || self.columnCount != columnCount {
+            partialResult = buildRows(incognito: incognito, maxCols: columnCount)
+            self.columnCount = columnCount
```
darinwf, line 45: probably better to assign `columnCount` first and then just have `buildRows` read that variable. no need for it to be a param right?

darinwf (commented):
You didn't ask for review feedback yet, but thought I might just prefetch it in case it is helpful. Sorry if any of this is redundant with what you already had in mind. Just trying to save some time :-)

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -27,15 +27,28 @@ class TabCardModel: CardModel {
...
     private(set) var allDetailsWithExclusionList: [TabCardDetails] = []
+    var columnCount: Int = 2
 
     // Tab Group related members
     private(set) var representativeTabs: [Tab] = []
     private(set) var allTabGroupDetails: [TabGroupCardDetails] = []
 
     @Default(.tabGroupExpanded) private var tabGroupExpanded: Set<String>
 
+    func updateRows(columnCount: Int? = nil) {
```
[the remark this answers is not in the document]
chassu: I made it optional to identify the caller who is calling because incognito mode is toggled. I'll try move away from this but make sure we don't forget the incognito path.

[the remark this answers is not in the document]
darinwf: Hmm... I'm not quite understanding. How is incognito part of this?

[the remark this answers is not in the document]
chassu: There was a bug when we switch from normal tab to incognito tab, the normal would still show because the rows are not refreshed. I had to add a change below:

```
    func switchToTabs(incognito: Bool) {
        switcherState = .tabs

        tabCardModel.manager.switchIncognitoMode(
            incognito: incognito, fromTabTray: true, openLazyTab: false)
        tabCardModel.updateRows()
    }

```

Client/Frontend/Browser/Card/CardGrid.swift
```diff
@@ -75,7 +75,11 @@ struct CardGrid: View {
 
     @State private var detailDragOffset: CGFloat = 0
     @State private var cardSize: CGFloat = CardUX.DefaultCardSize
-    @State private var columnCount = 2
+    @State private var columnCount = 2 {
```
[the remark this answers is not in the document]
chassu: Should we be careful about making TabCardModel.columnCount a published variable?

```
    func updateRows(columnCount: Int? = nil) {
        if let columnCount = columnCount, self.columnCount != columnCount {
            self.columnCount = columnCount
            rows = buildRows()
            self.objectWillChange.send()
        } else if columnCount == nil {
            rows = buildRows()
            self.objectWillChange.send()
        }
    }
```

What if `self.columnCount = columnCount` start invalidating views before `rows = buildRows()` actually happens?

Client/Frontend/Browser/Card/CardModels.swift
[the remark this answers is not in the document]
darinwf: OK, that seems reasonable, but how is this related to what I was saying above. I don't understand the connection. You can still just call `updateRows` here, and it should work.

Client/Frontend/Browser/Card/CardGrid.swift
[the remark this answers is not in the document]
darinwf: Are you worried about `columnCount` getting spammed? The `.useEffect` on `CardGrid` shows when that should be updated. We can also modify `updateCardSize` (which is the function that modifies `columnCount`) to take care to only set it when it has a new value. That's an easy change.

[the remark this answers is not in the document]
chassu: I was worried there is a synchronous path from TabCardModel.columnCount updating CardGrid before updateRows() actually updates `rows`, so columnCount gets modified -> tabCardModel.objectWillChange.send() -> CardGrid gets invalidated right away (before updateRows() modifies the rows array)

```
    @Published var columnCount: Int = 2 {
        didSet {
            updateRows()
        }
    }
```

Client/Frontend/Browser/Card/CardGrid.swift
```diff
@@ -134,11 +133,13 @@ struct CardGrid: View {
 
     func updateCardSize(width: CGFloat, topToolbar: Bool) {
         if width > 1000 {
-            columnCount = 4
+            tabCardModel.columnCount = 4
```
darinwf, line 136: As I mentioned earlier, this is the spot where you can avoid assigning to `columnCount` if the value hasn't changed. Can do this:

```
    var columnCount: Int
    if width > 1000 {
        columnCount = 4
    } else {
        columnCount = topToolbar ? 3 : 2
    }
    if tabCardModel.columnCount != columnCount {
        tabCardModel.columnCount = columnCount
    }
    self.cardSize = ...
```

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -27,15 +27,26 @@ class TabCardModel: CardModel {
...
+    }
 
     // Tab Group related members
     private(set) var representativeTabs: [Tab] = []
     private(set) var allTabGroupDetails: [TabGroupCardDetails] = []
 
     @Default(.tabGroupExpanded) private var tabGroupExpanded: Set<String>
 
+    func updateRows() {
+        buildRows()
```
darinwf, line 47: Instead of making `buildRows()` internally generate both `normalRows` and `incognitoRows`, how about giving it a parameter `incognito: Bool` and then have `updateRows()` call `buildRows` twice and store the result?

Perhaps `normalRows` and `incognitoRows` should be `@Published`? I'm surprised they are not.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -115,152 +126,163 @@ class TabCardModel: CardModel {
...
         // of this resolves a problem where TabGroupHeader doesn't hide arrows button
-        // when the number of tabs drops below maxCols.
-        tabGroupExpanded.forEach { groupID in
-            if let tabGroup = allTabGroupDetails.first(where: { groupID == $0.id }),
-                tabGroup.allDetails.count <= maxCols
-            {
-                tabGroupExpanded.remove(groupID)
+        // when the number of tabs drops below columnCount.
+        for incognito in [true, false] {
+            tabGroupExpanded.forEach { groupID in
```
darinwf, line 135: This update of `tabGroupExpanded` does not seem to depend on `incognito`, so not sure why it is part of this `for` loop.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -27,15 +27,26 @@ class TabCardModel: CardModel {
     private var subscription: Set<AnyCancellable> = Set()
 
     private(set) var manager: TabManager
+    private(set) var normalRows: [Row] = []
```
darinwf, line 30: Makes a lot of sense to have `normalRows` and `incognitoRows`!

Client/Frontend/Browser/Card/CardModels.swift
chassu: The two callers of buildRows() both ensures an objectWillChange.send() when it happens. You think we should make both of them published? I feel like that will introduce extra updates in the views.

chassu:
```
    @Published var columnCount: Int = 2 {
        didSet {
            updateRows()
        }
    }
```

```
        buildRows()

        objectWillChange.send()
```

chassu:
All xcui tests passed 😛

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -115,20 +127,21 @@ class TabCardModel: CardModel {
...
     }
 
-    func buildRows(incognito: Bool, maxCols: Int) -> [Row] {
+    func buildRows(incognito: Bool) -> [Row] {
         // When the number of tabs in a tab group decreases and makes the group
         // unable to expand, we remove the group from the expanded list. A side-effect
         // of this resolves a problem where TabGroupHeader doesn't hide arrows button
-        // when the number of tabs drops below maxCols.
+        // when the number of tabs drops below columnCount.
         tabGroupExpanded.forEach { groupID in
```
darinwf, line 135: Future optimization (or you can add it now), we can pull this out to `updateRows()` since it does not depend on the input parameter `incognito`. No reason to do this twice per call to `updateRows()`.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -283,6 +298,8 @@ class TabCardModel: CardModel {
             TabGroupCardDetails(tabGroup: $0, tabManager: manager)
         }
 
+        updateRows()
```
darinwf, line 301: I think we should move the `objectWillChange.send()` into `updateRows()`.... reason below:

Client/Frontend/Browser/Card/GridModel.swift
```diff
@@ -88,6 +88,7 @@ class GridModel: ObservableObject {
 
         tabCardModel.manager.switchIncognitoMode(
             incognito: incognito, fromTabTray: true, openLazyTab: false)
+        tabCardModel.updateRows()
```
darinwf, line 91: We probably need a call to `tabCardModel.objectWillChange.send()` here right?

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -27,15 +27,43 @@ class TabCardModel: CardModel {
...
+    func updateRows() {
+        // When the number of tabs in a tab group decreases and makes the group
+        // unable to expand, we remove the group from the expanded list. A side-effect
+        // of this resolves a problem where TabGroupHeader doesn't hide arrows button
+        // when the number of tabs drops below columnCount.
+        tabGroupExpanded.forEach { groupID in
+            if let tabGroup = allTabGroupDetails.first(where: { groupID == $0.id }),
+                tabGroup.allDetails.count <= columnCount
+            {
+                tabGroupExpanded.remove(groupID)
```
darinwf, line 55: I wonder if this is generating a bunch of updates at this point too.. could this be synchronously triggering a SwiftUI update? Maybe we should postpone writing to `tabGroupExpanded` until after we have updated all of our state. Just food for thought....

darinwf (approved):
LGTM!

PR #3233, opened 2022-03-28
Add support for matching the initialURL of (non-)restored tabs

### LINKED ISSUE #3224: addTabToTabGroupIfNeeded does not seem to match against zombie tabs
I've noticed that if I navigate a tab to nytimes.com, click on an article and then open a different tab to some other page, and then finally close the browser, if I then open the browser again and navigate to nytimes.com then that does not result in a tab group.

Perhaps the `Tab.originalURL` property is not restored until the WebView is created again. Or there's some other reason why `addTabToTabGroupIfNeeded` is failing to find the matching tab.

darinwf:
addTabToTabGroupIfNeeded was only capable of matching against a
tab that was created during the current session. Once the app
was killed, on next cold start any non-restored tabs or even
tabs that had been restored would not be candidates for matching.

The underlying issue is our use of internal:// URLs to handle
session restore. When we check the WebView for its history item's
initialURL property, that could be an internal:// URL. Or if we
haven't restored the tab yet, it could be nil. We then need to
check sessionData.urls[0].

Adds a convenience method on InternalURL to extract the nested
URL of a session restore internal URL.

Fixes #3224

### How was this tested?
- [x] I tested this manually

Unfortunately, I don't know how to write a good test for this yet. This
requires restarting the app to test.


darinwf:
Actually, I suspect I can write tests for this by just closing tabs and restoring them. Will work on that.

evancrow (approved):
LGTM

PR #3234, opened 2022-03-28
add options to add 500 tabs

chassu:
too lazy to press "create 100 tabs" 5 times 😛 I found out that on a testing device, creating 500 tabs could lead to crashes on startup, whereas that don't seem to happen on simulators.

### How was this tested?


darinwf (approved):
You are seeing a crash on device and not on the simulator due to memory limits of the respective devices. Your macbook pro has a lot more memory.

PR #3239, opened 2022-03-28
ZeroQueryViewForWeb3Mode

buracules:
- Update ZeroQueryView's UI for web3Mode
- Refactor codebase

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

Simulator Screen Shot - iPhone 13 Pro - 2022-03-28 at 22 52 38

Simulator Screen Shot - iPhone 13 Pro - 2022-03-28 at 22 52 25


WalletCore/CollectionViews.swift
```diff
@@ -119,7 +119,7 @@ struct CollectionWindowStatsMenu: View {
 }
 
 extension View {
-    fileprivate func gradientForeground() -> some View {
+    public func gradientForeground() -> some View {
```
yozuysal, line 122: lets move this to WalletStyle

Client/Frontend/ZeroQuery/View+Extensions.swift
```diff
@@ -0,0 +1,27 @@
...
+    func foregroundColorOrGradient(_ color: Color) -> some View {
+        if FeatureFlag[.web3Mode] {
+            self.gradientForeground()
+        } else {
+            self.foregroundColor(color)
+        }
+    }
+
+    @ViewBuilder
+    func backgroundColorOrGradient(_ color: Color?) -> some View {
```
yozuysal, line 20: let's make the default nil so that you can call with no arguments

`_ color: Color? = nil`

Client/Frontend/ZeroQuery/ZeroQueryView.swift
```diff
@@ -280,7 +238,7 @@ struct ZeroQueryView: View {
     @ViewBuilder
     private var searchesView: some View {
         ZeroQueryHeader(
-            title: "Searches",
+            title: FeatureFlag[.web3Mode] ? "Search on Ethereum (or the internet)" : "Searches",
```
yozuysal, line 243: nit: ...or the web)

Client/Frontend/ZeroQuery/SuggestedSitesViews.swift
```diff
@@ -40,7 +40,7 @@ struct SuggestedNavigationView<Content: View>: View {
                         width: SuggestedSiteUX.IconSize, height: SuggestedSiteUX.IconSize,
                         alignment: .center
                     )
-                    .background(Color(light: .ui.gray97, dark: .systemFill))
+                    .modifier(SuggestedItemBackgroundModifier())
                     .cornerRadius(SuggestedSiteUX.IconCornerRadius)
```
yozuysal, line 44: I think this cornerRadius should also if inside the if and applied to the else part only when web3Mode is not on.

Client/Frontend/ZeroQuery/View+Extensions.swift
```diff
@@ -25,3 +25,38 @@ extension View {
...
+            self
+                .clipShape(Hexagon().rotation(.degrees(30)))
+                .blendMode(.multiply)
+        }
+    }
+}
+struct Hexagon: Shape {
+    func path(in rect: CGRect) -> Path {
+        // hypotenuse (we make it fit inside the available rect
+        let h = Double(min(rect.size.width, rect.size.height)) / 2.0
```
yozuysal, line 45: let's use full words for the variables here and I think we can actually drop most of the comments here. So height instead of h, center instead of c, point for pt etc....

Client/Frontend/ZeroQuery/View+Extensions.swift
```diff
@@ -25,3 +25,38 @@ extension View {
...
+            Image("hexagon")
+                .resizable()
+                .scaledToFit()
+            self
+                .clipShape(Hexagon().rotation(.degrees(30)))
+                .blendMode(.multiply)
+        }
+    }
+}
+struct Hexagon: Shape {
```
yozuysal, line 42: Let's move this to ViewExtensions as well?

yozuysal (approved):
all styling nits. lgtm otherwise

Client/Frontend/ZeroQuery/SuggestedSitesViews.swift
```diff
@@ -81,7 +82,8 @@ private struct SuggestedItemBackgroundModifier: ViewModifier {
 }
 
 struct SuggestedHomeView: View {
-    let title = "Home"
+    let title: String =
+        FeatureFlag[.web3Mode] ? Defaults[.cryptoPublicKey].isEmpty ? "You" : "Your NFTs" : "Home"
```
yozuysal, line 86: let's use @Default(.cryptoPublicKey) here and turn this into a computed var `var title:String: {`. Otherwise during first run even after wallet setup, this will say You because it wont get the update

Client/Frontend/ZeroQuery/SuggestedSitesViews.swift
```diff
@@ -168,6 +170,7 @@ struct SuggestedSiteView: View {
 
 struct SuggestedSitesView: View {
     let isExpanded: Bool
+    var withHome: Bool = true
```
yozuysal, line 173: now we can use withHome: false for the 2nd SuggestedSItes section with marketplaces.

yozuysal (approved):
two small fixes. then this is ready to go

Issue #3209, opened 2022-03-24
Issuing a search asks the user to sign-in, but then after sign-in the user is navigated to neeva home


darinwf:
Instead, we should run the search the user was trying to enter.

PR #3245, opened 2022-03-29
Misc. IntroView Fixes

evancrow:
Few fixes for the `IntroView`:

1. No longer cancel the `WebView` navigation when displaying, which would result in a `Tab` that did nothing.
2. Cleaned up the model, and moved the code for displaying into the dedicated model.
3. Created a separate file for the `IntroView` stuff.
4. Cleaned up the code for dismissing the view, now the onDismiss code won't run until the view is closed (@chung-lai I believe this was the issue for the infamous FullscreenOverlay bug). I ran through a couple of different paths to make sure it worked and it should be fine, but if you have any specific cases for me to check, happy to do it.
5. If the `IntroView` opened from a search, don't open the Neeva webpage when the user signs in, instead just refresh the search page.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #3209


chung-lai:
We may reshuffle the first run experience a bit. Would love to hold off this PR before the reshuffling which should happen very soon.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -704,7 +704,7 @@ extension BrowserViewController: WKNavigationDelegate {
                         && Defaults[.signedInOnce]
                 {
                     self.presentIntroViewController(true, signInMode: true)
-                    decisionHandler(.cancel)
+                    decisionHandler(.allow)
```
chung-lai, line 707: Is this change intentional? If so, how does this relate to intro view?

darinwf:
I'm going to defer to @chung-lai's review on this one. @evancrow - Please test first-run flows thoroughly. It is easy to have a regression slip through since we don't see these flows that often in our normal browsing. Thanks!

Client/Frontend/Browser/BrowserViewController/BrowserViewController+IntroView.swift
```diff
@@ -0,0 +1,245 @@
...
+            }
+
+            if NeevaUserInfo.shared.hasLoginCookie() {
+                if let notificationToken = Defaults[.notificationToken] {
+                    NotificationPermissionHelper.shared
+                        .registerDeviceTokenWithServer(deviceToken: notificationToken)
+                }
+            }
+
+            if let arm = NeevaExperiment.arm(for: .defaultBrowserInterstitialFirst),
```
chung-lai, line 79: With the latest code, this piece of code goes away as we will be showing the interstitial as first run screen

Client/Frontend/Browser/BrowserViewController/BrowserViewController+IntroView.swift
```diff
@@ -0,0 +1,245 @@
...
+        onboardingVC.modalPresentationStyle = .formSheet
+
+        if let modalTransitionStyle = modalTransitionStyle {
+            onboardingVC.modalTransitionStyle = modalTransitionStyle
+        }
+
+        present(onboardingVC, animated: true, completion: nil)
+    }
+
+    func presentDBPromptView(fromSkipToBrowser: Bool = false) {
```
chung-lai, line 206: this is gone in the latest `main` as well

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
chung-lai: Discussed offline and this is intentional for neeva sign in

Client/Frontend/Intro/IntroViewModel.swift
```diff
@@ -75,11 +77,39 @@ class IntroViewModel: NSObject, ObservableObject {
...
+            completion()
+        }
+    }
+
+    public func dismiss(_ firstRunButtonAction: FirstRunButtonActions?) {
         Defaults[.introSeen] = true
 
         overlayManager.hideCurrentOverlay(ofPriority: .fullScreen) {
-            completion()
+            browserLog.info("Dismissed introVC")
```
chung-lai, line 105: So we don't need to call completion anymore?

chung-lai:
@evancrow, I test drove the change a bit and found this issue:

1. Fresh install the app
2. Do 5 searches
3. Sign up modal should pop up
4. Click Other sign up options
5. Sign in with existing account
6. Once the page reloads itself, the sign up modal shows up again

Screen Shot 2022-04-01 at 7 09 07 PM

Could it be relate to the change of reloading the current search page?

evancrow:
Yeah that appears to be the issue, also I don't think that's necessary anymore anyways.

chung-lai (approved):
LGTM! Let's merge this after branch cut if possible, would like to get it bake for a bit

PR #3247, opened 2022-03-29
Onboarding and build modifications for new target

yozuysal:
- Updated entitlements and splits them across dev and release
- Adds a dedicated launch screen file
- Updates Neeva Wallet wordmark asset to say Neevaxyz instead
- Adds a first step to wallet onboarding
- Updates copy for onboarding, community submissions and malicious site warning
- Moves to a new graphQL api for community submissions that doesn't require login
- Changes onboarding flow to end with showing the wallet panel half screen
- Gracefully fails on import fail without crashing

### How was this tested?
- [X] I tested this manually
- [ ] I added automated tests


Client.xcodeproj/project.pbxproj
```diff
@@ -728,6 +729,9 @@
 		4AB8160527EA64ED0017203D /* SuggestedXYZSearchesView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4AB8160427EA64ED0017203D /* SuggestedXYZSearchesView.swift */; };
 		4AB8160627EA662E0017203D /* ScreenCaptureHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 328585B127E8FE7D00717DD5 /* ScreenCaptureHelper.swift */; };
 		4AB8160727EA663A0017203D /* ServerLoggingSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 676685C427E50EAF003CEA6D /* ServerLoggingSettingsView.swift */; };
+		4AB8160827EA99F40017203D /* (null) in Sources */ = {isa = PBXBuildFile; };
```
chung-lai, line 732: Any idea what this `(null)` is?

Client.xcodeproj/project.pbxproj
```diff
@@ -7712,10 +7726,11 @@
 			buildSettings = {
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				ASSETCATALOG_COMPILER_APPICON_NAME = XYZAppIcon;
+				ASSETCATALOG_COMPILER_INCLUDE_ALL_APPICON_ASSETS = NO;
 				CODE_SIGN_ENTITLEMENTS = Client/Entitlements/Neevaxyz.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
 				CODE_SIGN_STYLE = Manual;
-				CURRENT_PROJECT_VERSION = 200;
+				CURRENT_PROJECT_VERSION = 19;
```
chung-lai, line 7733: Looks like the version has changed somehow but we are good for now. I'll fix the version script some time this week before `main` build

WalletCore/XYZIntroView.swift
```diff
@@ -0,0 +1,57 @@
...
+    @Default(.cryptoPublicKey) var publicKey
+    @State var isCreatingWallet: Bool = false
+    @Binding var viewState: ViewState
+
+    public init(viewState: Binding<ViewState>) {
+        self._viewState = viewState
+    }
+
+    public var body: some View {
+        VStack(alignment: .leading) {
```
chung-lai, line 20: Have we try this on various screens?

chung-lai (commented):
Overall the changes look good.

* By adding the entitlement, does the keychain access work now?
* You may not want the swiftformat change?

Client.xcodeproj/project.pbxproj
yozuysal: I was wondering about this too. I think this may be an after effect of bad cherry-picks. I can just remove and confirm that things still work?

Client.xcodeproj/project.pbxproj
yozuysal: oh I can change this back actually. I can always modify before archiving on my end

WalletCore/XYZIntroView.swift
yozuysal: yeah, at least for a few phone screens. Tablet is not looking great but I think we are OK with that for now.

yozuysal:
I actually have one more one liner to upload to make the keychain work OK.

I am very confused about what to do around swift format. Right now running it locally is giving me `The loaded '_InternalSwiftSyntaxParser' library is from a toolchain that is not compatible with this version of SwiftSyntax`. So I can't upload anything because git push is angry that I am not swift formatting

chung-lai:
> I actually have one more one liner to upload to make the keychain work OK.
> 
> I am very confused about what to do around swift format. Right now running it locally is giving me `The loaded '_InternalSwiftSyntaxParser' library is from a toolchain that is not compatible with this version of SwiftSyntax`. So I can't upload anything because git push is angry that I am not swift formatting

I saw that error as well when I was switching between XCode version. May be we should get @EdwardJTL or someone's help on understanding the swiftformat change. Feel like there is something to do with upstream branch of swiftformat.

EdwardJTL:
Looks like you committed a change to the swift-format submodule? that could cause the argument parse being out of sync with swift-format
and @chung-lai, when you switch between xcode versions, you should probably rebuild swift-format

yozuysal:
> Looks like you committed a change to the swift-format submodule? that could cause the argument parse being out of sync with swift-format and @chung-lai, when you switch between xcode versions, you should probably rebuild swift-format

hmm. which branch of swift-format should I be on when I rerun bootstrap? main? or does running bootstrap also switch to the right branch?

EdwardJTL:
> > Looks like you committed a change to the swift-format submodule? that could cause the argument parse being out of sync with swift-format and @chung-lai, when you switch between xcode versions, you should probably rebuild swift-format
> 
> hmm. which branch of swift-format should I be on when I rerun bootstrap? main? or does running bootstrap also switch to the right branch?

run 
`git checkout "tags/0.50600.0"`
and then
`swift build -c release`
in `swift-format` dir. you dont need to boot strap after that

yozuysal:
> > > Looks like you committed a change to the swift-format submodule? that could cause the argument parse being out of sync with swift-format and @chung-lai, when you switch between xcode versions, you should probably rebuild swift-format
> > 
> > 
> > hmm. which branch of swift-format should I be on when I rerun bootstrap? main? or does running bootstrap also switch to the right branch?
> 
> run `git checkout "tags/0.50600.0"` and then `swift build -c release` in `swift-format` dir. you dont need to boot strap after that
> 
> oh and make sure we don't check in the swift-format submodule change

Thank you! Working now!


@chung-lai I removed the extra project file changes with (nil) and version number. Without using appGroup, the keychain just works fine so I will just go with that. But I am keeping the entitlements as is, just in case we need to use the group for something else.

chung-lai (approved):

Issue #3198, opened 2022-03-23
NavigationRouter should bounce incoming URLs to other installed apps (e.g., Lyft)


darinwf:
If you have for example the Lyft app installed, and you have the Neeva app set as default browser, then when clicking on a link to lyft.com from another app (e.g., Google Maps), the Neeva app will just load the lyft.com URL. Instead, it should invoke the Lyft app just as Safari would do.

PR #3248, opened 2022-03-29
Expand LSApplicationQueriesSchemes + Fix OpenInAppOverlay

evancrow:
We already have the code to support bouncing URLs, just needed to expand the list of URLs to bounce. Happy to take suggestions of more to add, we get a maximum of 50. Also, fixed a bug where the `OpenInAppOverlay` would not dismiss.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
- Go to google maps
- Open Lyft link
 - Make sure `OpenInAppOverlay` appears

## Issues addressed
Fixed: #3198


jonathangarelick:
I'm not too familiar with this, but I think there might be a more scalable way to address this rather than using an allow list. It looks like [this](https://github.com/mozilla-mobile/firefox-ios/blob/ff98cf659c5858fea321b6ed7114edc954c888f2/Client/Frontend/Browser/BrowserViewController/BrowserViewController%2BWebViewDelegates.swift#L544) is how Firefox iOS handles this.

My understanding is that apps can link URLs like lyft://newride?from=xxx&to=yyy, and so NavigationRouter can manage this like it manages URLs with other schemes, like http. However, I might be wrong, so happy to get additional input.

EdwardJTL:
> I'm not too familiar with this, but I think there might be a more scalable way to address this rather than using an allow list. It looks like [this](https://github.com/mozilla-mobile/firefox-ios/blob/ff98cf659c5858fea321b6ed7114edc954c888f2/Client/Frontend/Browser/BrowserViewController/BrowserViewController%2BWebViewDelegates.swift#L544) is how Firefox iOS handles this.
> 
> My understanding is that apps can link URLs like lyft://newride?from=xxx&to=yyy, and so NavigationRouter can manage this like it manages URLs with other schemes, like http. However, I might be wrong, so happy to get additional input.

I second this! I noticed that both Chrome and Firefox gives popups for launching another app for a link, but i never figured out why we can't do the same. great find :)

evancrow:
So what you just shared is a part of the answer, a part which our app already handles. (our implementation is [here](https://github.com/neevaco/neeva-ios/blob/8c42389120881ecd901021f2f33f4ef591331b5e/Client/Frontend/Browser/BrowserViewController/BrowserViewController%2BWebViewDelegates.swift#L730))

The key to that code is the line ` UIApplication.shared.open(url, options: [:])`.  Running that code will just cycle back to our app (because we are the default browser) unless we tell the system that it should look to redirect the URL to an external app, if it matches a host inside `LSApplicationQueriesSchemes`. 

We inherited our list of `LSApplicationQueriesSchemes` from Firefox, it's just not a very complete list. 

See more [here](https://developer.apple.com/documentation/uikit/uiapplication/1622952-canopenurl) :)

EdwardJTL:
> So what you just shared is a part of the answer, a part which our app already handles. (our implementation is [here](https://github.com/neevaco/neeva-ios/blob/8c42389120881ecd901021f2f33f4ef591331b5e/Client/Frontend/Browser/BrowserViewController/BrowserViewController%2BWebViewDelegates.swift#L730))
> 
> The key to that code is the line ` UIApplication.shared.open(url, options: [:])`. Running that code will just cycle back to our app (because we are the default browser) unless we tell the system that it should look to redirect the URL to an external app, if it matches a host inside `LSApplicationQueriesSchemes`.
> 
> We inherited our list of `LSApplicationQueriesSchemes` from Firefox, it's just not a very complete list.
> 
> See more [here](https://developer.apple.com/documentation/uikit/uiapplication/1622952-canopenurl) :)

https://developer.apple.com/documentation/uikit/uiapplication/1648685-open
this one says 
```
 If the specified URL scheme is handled by another app, iOS launches that app and passes the URL to it. (Launching the app brings the other app to the foreground.) If no app is capable of handling the specified scheme, the completion handler is called with the success parameter set to false.
```

and it looks like it's what firefox is calling

jonathangarelick:
Hmm, it looks like they are doing something we're not, though. Here's [their Client Info plist](https://github.com/mozilla-mobile/firefox-ios/blob/5da67df3711b60ead0ff12e2d49c5c048a008c1c/Client/Info.plist#L52) (no Lyft). Here's what I see when I click on a Lyft rideshare suggestion from Google Maps:



In Neeva it just loads the mobile webpage for Lyft -- no prompt.

evancrow:
Ah I think I found the difference... maybe:

```
// Use for links, that do not show a confirmation before opening.
fileprivate func showOverlay(forExternalUrl url: URL) {
    // NOTE: This only considers schemes included in our LSApplicationQueriesSchemes
    // PList declaration. We should probably take that into consideration here and
    // still allow users to open other, unknown URLs.
    guard UIApplication.shared.canOpenURL(url) else {
        return
    }

    tabManager.selectedTab?.stop()

    showModal(style: .grouped) {
        OpenInAppOverlayContent(url: url)
    }
}
```

I guess we need to remove` UIApplication.shared.canOpenURL(url)`? That was probably added for a reason but the comment also makes it sound like it could be removed... any thoughts @darinwf?

darinwf (approved):

darinwf:
I think I stamped this prematurely. Didn't see all of the discussion on this PR....

darinwf:
I'm confused. Issue #3198 is about letting the Lyft app handle incoming `lyft.com` URLs. It isn't about `lyft:` URLs. Maybe there is also an issue with `lyft:` URLs?

jonathangarelick:
> I'm confused. Issue #3198 is about letting the Lyft app handle incoming `lyft.com` URLs. It isn't about `lyft:` URLs. Maybe there is also an issue with `lyft:` URLs?

I think the way it works is like this:
1. User clicks Lyft button in Google Maps
2. Lyft.com loads in the browser and [checks if the app is installed](https://stackoverflow.com/q/13044805)
3. If the app is installed, then Lyft.com does a redirect to a URL with the custom scheme (`lyft://`), otherwise use the mobile web version.

AFAIK, the only way to open a native app from the web browser is to use custom schemes.

Related link: [Defining a Custom URL Scheme for Your App](https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app)

evancrow:
@jonathangarelick @EdwardJTL 

Thought about this some more, and I think the best solution is to keep the `guard UIApplication.shared.canOpenURL(url)`. This will prevent us from showing the `OpenInAppOverlay` if an app is not installed, or there isn't one to handle the specific scheme. I would rather add schemes of trusted sites to the `info.plist` rather than risk the possibility of showing the `OpenInAppOverlay` that won't work.

darinwf (approved):
OK, let's go with this for now.

EdwardJTL:
> @jonathangarelick @EdwardJTL
> 
> Thought about this some more, and I think the best solution is to keep the `guard UIApplication.shared.canOpenURL(url)`. This will prevent us from showing the `OpenInAppOverlay` if an app is not installed, or there isn't one to handle the specific scheme. I would rather add schemes of trusted sites to the `info.plist` rather than risk the possibility of showing the `OpenInAppOverlay` that won't work.

would it just open the app store page for the app?

EdwardJTL:
> @jonathangarelick @EdwardJTL
> 
> Thought about this some more, and I think the best solution is to keep the `guard UIApplication.shared.canOpenURL(url)`. This will prevent us from showing the `OpenInAppOverlay` if an app is not installed, or there isn't one to handle the specific scheme. I would rather add schemes of trusted sites to the `info.plist` rather than risk the possibility of showing the `OpenInAppOverlay` that won't work.

Actually, `open(_:options:completionHandler:)` has a completion call back. how do we feel about opening the url anyways, and show an error alert when it doesn't open correctly (perhaps due to the app not installed).

This is also what firefox does (looking at their code).

evancrow:
Probably a question for design... but at the very least a follow-up for a different PR. Do you want to add that to the design open hours/mobile eng. weekly for discussion?

evancrow:
I'm going to merge this PR and we can figure out this discussion separately

PR #3253, opened 2022-03-29
update rows on drag and drop

chassu:
Due to the buildRows() changes, drag and drop needs to call updateRows() to work.

### How was this tested?
- [x] I tested this manually


darinwf (approved):
LGTM

Issue #48, opened 2021-03-08
Site specific settings


darinwf:
See issue #43. It would be nice to provide access here to site-specific settings that would include web-site capabilities such as location, camera and microphone.

However, we need to investigate if it is even possible to implement such UX. Neither Firefox nor Chrome have these options. Safari does. But it may not be something WKWebView readily supports.

PR #3256, opened 2022-03-29
Revert Reversion of Site Specific Settings + Bug Fix

evancrow:
Because we keep track if we allow trackers, but then store that in the tracking protection enabled variable (which is the opposite) the state was being set incorrectly. We were setting tracking protection enabled with is tracking allowed. 

Sort-of related, but while debugging I found that we're often calling `selectTab` for a `Tab` that was already selected. Preventing this should give us some performance back since we're not redundantly updating the state now.

### How was this tested?
- [x] I tested this manually
- [x] I added automated tests (<- working on)
 
## Issues addressed
Fixed: #48


UITests/SecurityTests.swift
```diff
@@ -63,7 +63,7 @@ class SecurityTests: UITestBase {
         let webView = tester().waitForView(withAccessibilityLabel: "Web content") as! WKWebView
 
         // Make sure the URL doesn't change.
-        XCTAssertEqual(webView.url?.path, "/errorpage")
+        XCTAssertEqual(webView.url?.absoluteString, testURL)
```
darinwf, line 66: See https://github.com/neevaco/neeva-ios/issues/3258. I don't think this is the correct fix for the test.

evancrow: I'll remove this change from the PR, not intentional probably just apart of the original PR.

evancrow:
I'm actually going to close this since we got different mocks for Cookie Cutter...

PR #3264, opened 2022-03-29
Add YourNFTs to WalletDashboard

buracules:
- Added Your NFTs section to WalletDashboard
- Updated AssetViews UI

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

Simulator Screen Shot - iPhone 13 Pro - 2022-03-29 at 23 12 11

Simulator Screen Shot - iPhone 13 Pro - 2022-03-29 at 23 12 01

## Issues addressed


Client/Frontend/NeevaWallet/WalletDashboard.swift
```diff
@@ -12,6 +12,31 @@ import WalletConnectSwift
 import WalletCore
 import web3swift
 
+struct WalletHeader: View {
+    let title: String
+    @State private var isExpanded: Bool = true
```
yozuysal, line 17: let's make this a @Binding instead of @State and we can remove onStateChange below. (Then we will need to send $showBalances, $showNFTs etc and they will be changed through the binding

Client/Frontend/NeevaWallet/Web3Model.swift
```diff
@@ -121,6 +123,9 @@ class Web3Model: ObservableObject, ResponseRelay {
         self.closeTab = { tab in
             tabManager.close(tab)
         }
+        self.openURLForSpace = {
```
yozuysal, line 126: what happens when you try?

`self.openURLForSpace = tabManager.createOrSwitchToTabForSpace`

Client/Frontend/NeevaWallet/AssetView.swift
```diff
@@ -27,54 +27,93 @@ extension Asset: SelectableThumbnail {
...
+            else {
+                return
+            }
+            walletDetailsModel.showingWalletDetails = false
+            browserModel.hideWithNoAnimation()
+            DispatchQueue.main.async {
+                openURLForSpace(url, web3Model.wallet?.publicAddress ?? "")
+            }
+        } label: {
+            Image("opensea-logo")
```
yozuysal, line 107: should we just use the url for this and turn this into a

`func button(url: URL) -> some View` for both to avoid code repetition?

Client/Frontend/NeevaWallet/AssetView.swift
```diff
@@ -27,54 +27,93 @@ extension Asset: SelectableThumbnail {
...
+            guard
+                let url = URL(
+                    string:
+                        "https://neeva.xyz/search?q=\(asset.contract.address)&contractAddress=\(asset.contract.address)&tokenID=\(asset.tokenID)"
+                )
+            else {
+                return
+            }
+
+            walletDetailsModel.showingWalletDetails = false
```
yozuysal, line 77: I dont think we need this and the line below

Client/Frontend/NeevaWallet/WalletDashboard.swift
```diff
@@ -306,25 +322,53 @@ struct WalletDashboard: View {
...
+                    WalletHeader(
+                        title: "Your NFTs",
+                        onStateChange: { state in
+                            self.showNFTs = state
+                        })
+                }
+            }
+        )
+        .padding(16)
+        .background(Color.clear)
```
yozuysal, line 368: no need to assign this

Client/Frontend/NeevaWallet/Web3Model.swift
buracules: It is still wants arguments and I can't pass them without closure block

yozuysal: ok. I was hoping the compiler would be smart enough since openURLForSpace and createOrSwitchToTabForSpace has the same parameter types and assign $0 and $1 accordingly. Does it not compile with this?

buracules: No it gives the following error. 

`Cannot assign value of type '(URL, String) -> TabManager.CreateOrSwitchToTabResult' to type '(URL, String) -> Void'`

yozuysal: Ah, never mind. I forgot about the return value.

yozuysal:
Can we put this UI behind a new flag in FeatureFlags? FeatureFlags.showNFTsInWallet and only show the section when it is enabled?

Client/Frontend/NeevaWallet/WalletDashboard.swift
```diff
@@ -329,7 +345,7 @@ struct WalletDashboard: View {
     var nftSection: some View {
         Section(
             content: {
-                if showNFTs {
+                if FeatureFlag[.showNFTsInWallet] {
```
yozuysal, line 348: this should probably be FeatureFlag[.showNFTsInWallet] && showNFTs so that collapsing still works

yozuysal: actually now that you have the one below, maybe we dont need this edit at all. This should just be showNFTs

EdwardJTL:
Next time please double check that changes to swift-format is not pushed and merged into main

PR #3265, opened 2022-03-29
expand tab group should update rows

chassu:
the expand/collapse button should also update rows.

### How was this tested?
- [x] I tested this manually


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -86,7 +86,7 @@ class TabCardModel: CardModel {
         }.store(in: &subscription)
 
         _tabGroupExpanded.publisher.sink { [weak self] _ in
-            self?.objectWillChange.send()
+            self?.updateRows()
```
darinwf, line 89: Hmm.... but `updateRows()` can modify `tabGroupExpanded`, so this could recurse. Isn't that a problem? Perhaps it would only do so once and then become stable. Actually, I think the solution is to move updating of `tabGroupExpanded` out of `updateRows` and into `onDataUpdated`.

darinwf (approved):
LGTM

PR #3266, opened 2022-03-29
Simplify versioning control

chung-lai:
* All versioning for browser inherited from project target (hence removing most VERSION strings in the project file)
* Introduce AppConfig.xcconfig at the project level
* Define version in AppConfig.xcconfig
* Change parent project version to use the `BROWSER_PROJECT_VERSION` and `BROWSER_MARKETING_VERSION`
* Update version script to only update version in AppConfig.xcconfig
* Remove ApiConfig.xcconfig usage from browser client

After this change, we will need to figure out how we should define API keys in the Client target because we no longer use ApiConfig which gets only generated after `bootstrap.sh`.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases

  and dark mode

## Issues addressed


Client/AppConfig.xcconfig
```diff
@@ -0,0 +1,9 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Configuration settings file format documentation can be found at:
```
darinwf, line 5: Thanks for adding this comment! Very helpful!

darinwf (approved):
LGTM!

PR #3269, opened 2022-03-30
Darin/log attribution

darinwf:


Client/EnvironmentHelper.swift
```diff
@@ -139,11 +140,18 @@ public class EnvironmentHelper {
             key: LogConfig.Attribute.PreviewModeQueryCount,
             value: String(Defaults[.previewModeQueries].count))
 
-        let attributes = [
+        var attributes = [
             getSessionUUID(), isUserSignedIn, deviceTheme, deviceName, firstRunPath,
             previewQueryCount,
         ]
 
+        if #available(iOS 14.3, *), let token = try? AAAttribution.attributionToken() {
```
chung-lai, line 148: Does it generate a new token every time? From documentation, it says this returns error when there is no network but I assume it doesn't make any network call. Reason I ask is making sure it's not a blocking network call :)

darinwf: I don't know for sure. That's a good question. I assumed it was just accessing some local information.

chung-lai: Discussed offline. Looks like it generates the same id between call and works in airplane mode.

chung-lai (approved):

PR #3270, opened 2022-03-30
Update tracking prevention as AdBlock

buracules:
- Created new easy_filter clone for Adblock (easy_filter_web3.json)
- Added missing cases for new filter

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested with "https://windowscentral.com" it blocked most of ads

Edit: Ads are random. I tought it blocked but it is not blocking.

#### Without Adblock
Simulator Screen Shot - iPhone 13 Pro - 2022-03-30 at 20 14 03

#### With Adblock
Simulator Screen Shot - iPhone 13 Pro - 2022-03-30 at 20 14 10


Client.xcodeproj/project.pbxproj
```diff
@@ -1394,6 +1394,8 @@
 		E67CD27527F33BCD007BFE58 /* ZeroQueryUX.swift in Sources */ = {isa = PBXBuildFile; fileRef = E64024E927F2437800D65B63 /* ZeroQueryUX.swift */; };
 		E67CD27627F33BD1007BFE58 /* SuggestedXYZSearchesView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E63DBA5627F339FC001B7A68 /* SuggestedXYZSearchesView.swift */; };
 		E67CD27727F33C3B007BFE58 /* TabGroup.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A78AA2427ED33C4002E7A86 /* TabGroup.swift */; };
+		E67CD27C27F49B1F007BFE58 /* easy_privacy_web3.json in Resources */ = {isa = PBXBuildFile; fileRef = E67CD27B27F49B1F007BFE58 /* easy_privacy_web3.json */; };
```
chung-lai, line 1397: Looks like we will be using different json file for different target. Could we only include this json in the new target? And make sure the browser target doesn't get this json file included in the binary. Just trying to make sure we don't increase the binary size.

chung-lai (commented):
LGTM! Could you talk a bit about how testing is done?

Client.xcodeproj/project.pbxproj
buracules: But if i do this Client target will crash with following error when Web3Mode is enabled.

` assertionFailure("Blocklists: bad file path.")`

buracules:
> LGTM! Could you talk a bit about how testing is done?

Updated first comment

Client.xcodeproj/project.pbxproj
chung-lai: Agree, we should gracefully fail and look for a backup json file.

buracules: Its probably gonna be fixed after compiler flags (because after merge you cant enable web3 on client anymore). Do we need backup json?

chung-lai: We don't necessarily need a backup json but would be great if we can fail the case gracefully instead of crashing. It's a nice to have and unrelated to this PR :D

buracules:
I will merge this one after compiler flags.

PR #3275, opened 2022-03-30
Adds logs to web3 mode actions.

yozuysal:
These logs are always taken for web3Mode and are not in any other mode


yozuysal:
@chung-lai wanted to check in with you re the setup here. This is supposed to work with the case with no neeva login. I followed other similar first run situations. But let me know if there is a better way to handle this.

chung-lai (approved):
The change looks good and the logging should work even without an account. Like we discussed, there is a concept of session id. We can talk about it and you can link events together.

PR #3279, opened 2022-03-30
Remove canOpenURL

evancrow:
Tested out an idea discussed in #3248.

Removes `canOpenURL` which checks if a scheme is inside our `info.plist` before showing the `OpenInExternalAppOverlay`. Instead, we show the overlay regardless (best way to account for all URLs) and fallback on an error `Toast`.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/37356984/160942635-56185d49-b0d0-45a9-9cac-b73ab4e48039.MP4


darinwf:
Looks good. Can you review the copy with @Ria-yu?

Thanks!

Ria-yu:
> Looks good. Can you review the copy with @Ria-yu?
> 
> Thanks!

we landed on "Unable to open link in external app. Check if the app is installed on this device."

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -488,18 +488,19 @@ extension BrowserViewController: WKNavigationDelegate {
 
     // Use for links, that do not show a confirmation before opening.
     fileprivate func showOverlay(forExternalUrl url: URL) {
-        // NOTE: This only considers schemes included in our LSApplicationQueriesSchemes
```
darinwf, line 491: Should we delete the `LSApplicationQueriesSchemes` entry as well?

darinwf (approved):
Future improvements might include giving users the ability to remember these choices so they don't have to constantly accept these prompts. We might wait until we observe frequent usage patterns like that or get user feedback.

PR #3280, opened 2022-03-31
Make Spaces detail UI more accessible

### LINKED ISSUE #1717: Spaces Detail UI is not fully accessible
#1716 addresses the lowest hanging fruit here, but I think the toolbar at the top needs a bit more work.

Right now, the individual space items, share button, more button and the items inside and the title for the header is accessible.

1. It is not possible to access the "Back" button inside the header. One of the reasons for this is the GridPicker below this UI is still accessible and captures touches. The other is, this back button doesnt have a label.
2. It is not possible to scroll down. We might want to audit ScrollViews we have in general and see if there are others we have missed.
3. The  tab switcher toolbar below this UI is still accessible. So can pick up touches, we should set it hidden while detail view is showing.

jonathangarelick:
Adds accessibility to the Spaces detail UI. 

Issue Pt 1 -- ✅ 
Issue Pt 2 -- I think the UI is already scrollable, you just have to tap in the main section of the Space before scrolling. See screen recording for demo.
Issue Pt 3 -- I think I addressed this (see code change), but I didn't have any issues with it when I was using VoiceOver.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

(turn sound on)

https://user-images.githubusercontent.com/41495718/161363328-4b289f96-af74-4462-b79f-3a8ffd182bd5.MP4

## Issues addressed
* Fixes #1717


Client/Frontend/Browser/Spaces/SpaceTopView.swift
```diff
@@ -39,10 +39,10 @@ struct SpaceTopView: View {
                     self.gridModel.closeDetailView()
                 },
                 label: {
-                    Symbol(decorative: .arrowBackward)
+                    Image(systemName: "arrow.backward")
```
evancrow, line 42: I wouldn't do this, should try and keep it consistent with the other views in the app. What you might try is modifying the `Symbol` label parameter or adding `. accessibilityLabel(label: Text(""))` directly to the symbol.

jonathangarelick: Yeah, so strangely, it doesn't work when I use `Symbol` and either apply the label directly to the symbol or the `label` argument.

jonathangarelick: It also works when I use `Image(systemSymbol: .arrowBackward)`.

Client/Frontend/Browser/Spaces/SpaceTopView.swift
```diff
@@ -39,7 +39,8 @@ struct SpaceTopView: View {
                     self.gridModel.closeDetailView()
                 },
                 label: {
-                    Symbol(decorative: .arrowBackward)
+                    Label("Return to all Spaces view", systemSymbol: .arrowBackward)
+                        .labelStyle(.iconOnly)
```
jonathangarelick, line 43: @evancrow How's this for a compromise? It's working with VoiceOver, and it uses an enum.

jonathangarelick:
Looks like the "expand description" button is not accessible either.

jonathangarelick:
@evancrow Okay, apologies if you got a bunch of emails from this. I think I got what you're looking for. I also fixed the accessibility of the expand/collapse button. Let me know if you need anything else.

evancrow (approved):
LGTM

PR #3281, opened 2022-03-31
Add Cookie Cutter UI

evancrow:
Adds a base for the Cookie Cutter UI behind the flag. Doesn't actually do anything yet, will follow up with an issue of next steps. Also, renames Hall of Shame to Who's Tracking You.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

cc @jlynch630 @hwiatneeva


evancrow:
Next steps issue #3282

Client/Frontend/Browser/CookieCutter/CookieCutterModel.swift
```diff
@@ -0,0 +1,53 @@
...
+class CookieCutterModel: ObservableObject {
+    @Published var cookieNotices: CookieNotices {
+        didSet {
+            guard cookieNotices != oldValue else {
+                return
+            }
+
+            Defaults[.cookieNotices] = cookieNotices
+
+            if cookieNotices == .declineNonEssential {
```
darinwf, line 33: nit: you could write this like so:

```
                let allowAll = cookieNotices != .declineNonEssential
                marketingCookiesAllowed = allowAll
                analyticCookiesAllowed = allowAll
                socialCookiesAllowed = allowAll
```

But actually maybe we don't want to force them to `..Allowed = true` in the case where the values are user selected?

darinwf (approved):

Client/Frontend/Browser/CookieCutter/CookieCutterModel.swift
evancrow: Hmm that's what was shown in the mocks. If the user options are all set to false it should then revert to `declineNonEssential`.

Issue #3278, opened 2022-03-30
Web page is clipped on the left hand side when in landscape mode


darinwf:
This is with 1.34.8:

https://user-images.githubusercontent.com/79867437/160940591-a40dac6b-6cd7-48b4-bb20-78331a7cbb50.MP4

darinwf:
@chassu was also seeing this. I wasn't able to repro until I tapped on a link on techmeme.com while it was in landscape mode and hit the back button.

evancrow:
Might be an issue with the SimulatedBackSwipeController... investigating

PR #3283, opened 2022-03-31
Fix Clipped WebPage

evancrow:
Offset the X with the leading and some of the trailing (too much and you can't interact with it) offset.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #3278


Client/Frontend/Browser/Content/TabContainerHost.swift
```diff
@@ -180,7 +182,9 @@ struct TabContainerContent: View {
                         )
                         .opacity(!simulatedSwipeModel.hidden ? 1 : 0)
                         .offset(
-                            x: -geom.size.width + simulatedSwipeModel.overlayOffset,
+                            x: -geom.size.width - geom.safeAreaInsets.leading
+                                - (geom.safeAreaInsets.trailing / 4)
```
darinwf, line 186: why divide trailing by 4?
why does this work?

evancrow: I added a comment explaining, forgot to push it. One sec...

darinwf (approved):
OK, this makes sense to me now. Thanks!

Issue #3272, opened 2022-03-30
Bottom bar stuck in tab switcher mode while web page and URL bar is showing


darinwf:
I saw this happen while using the browser on an iPhone 6s.

Video here:

https://user-images.githubusercontent.com/79867437/160880561-b966ccea-538e-48a6-b17e-bb7815b1d328.MOV

darinwf:
It seems like somehow the bottom bar did not respond properly to `showContent` being set to true. Unfortunately, I didn't test if I was able to tap on any links or interact with the URL bar. I suspect I would have been able to do so since the page scrolled properly and the URL bar was hiding on scroll. The bottom bar was just in some broken state, so we should probably focus the investigation on how that could be the case.

darinwf:
This is with latest TestFlight: 1.34.5.

And I'm able to repro by just switching tabs a bunch on the device...

darinwf:
Tapping on a link in the page fixes it. Could be the navigation events help refresh the UI somehow. Tapping on the overflow button shows the menu, but dismissing that menu, then the bottom bar is shown properly.

evancrow:
I've also been able to repro by rotating my device and then opening the tab switcher... might have something to do with the `ChromeModel`?

darinwf:
Unfortunately the fix caused a bigger regression, so had to revert. Re-opening this bug.

evancrow:
@darinwf Should we try using the original fix I had?

darinwf:
I'm seeing this bug show up in the simulator now:

https://user-images.githubusercontent.com/79867437/162484515-9a19d20d-b4d2-49e5-953a-db179a9d931e.mp4

The repro is just tapping on the tab card. @evancrow - I think it is worth understanding this bug better. I can't explain why the extra `objectWillChange.send()` call would help so it is making me nervous about just shipping that fix. We should try to understand what is really going on so we can be confident about the fix.

darinwf:
And now I cannot repro ... :-(

evancrow:
Yeah this bug is quite elusive, have not seen it in a bit

darinwf:
I went back to my iPhone 6s, and I can repro (sometimes).

evancrow:
So maybe this is a performance-related bug?

PR #3284, opened 2022-03-31
Fix Wrong BottomBar State

evancrow:
@darinwf Think I found a fix to the bug you filed, haven't been able to recreate it with this added line. Can you try it out as well, just to make sure?

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #3272


Client/Frontend/Browser/Card/CardGrid.swift
```diff
@@ -51,8 +51,10 @@ struct CardGridBackground: View {
                             withAnimation(CardTransitionUX.animation) {
                                 let showGrid =
                                     (cardTransitionModel.state == .visibleForTrayShow)
+
                                 if browserModel.showGrid != showGrid {
                                     browserModel.showGrid = showGrid
+                                    browserModel.objectWillChange.send()
```
darinwf, line 57: Hmm... `showGrid` is a `@Published` variable so this call to `objectWillChange.send()` should be redundant, right? If this helps though maybe that is a clue as to what is really going on...

evancrow: In my debugging I saw that that the value of `showGrid` was being set to false, but it was not changing in `BottomBarView` (listened with an `onChange` modifier).

darinwf: hmm

darinwf: Note, SwiftUI `onChange` runs before the new value is written to the property that is changing. By calling `objectWillChange.send()` after modifying `showGrid`, what you are essentially doing is generating an update after `showGrid` has been given its new value.

Since this fix seems to work, it makes me wonder if SwiftUI is performing an update synchronously (instead of asynchronously) when `showGrid` has not yet been updated. That would explain the bug we are seeing.

A fix could be to change `showGrid` to not be a published variable, and then to instead call `objectWillChange.send()` directly from the `didSet` on `showGrid`.

darinwf (commented):
I added a commit that does what I was thinking. This seems to work as well.

evancrow:
@darinwf Your change LGTM, feel free to stamp since I cannot.

darinwf (approved):

PR #3287, opened 2022-03-31
Resolve attributionToken before logging

darinwf:
### How was this tested?
- [x] I tested this manually


Client/Logging/ConversionLogger.swift
```diff
@@ -24,8 +26,49 @@ class ConversionLogger {
...
+            request.setValue("text/plain", forHTTPHeaderField: "Content-Type")
+
+            URLSession.shared.dataTask(with: request) { data, response, error in
+                guard error == nil else {
+                    Logger.browser.error(
+                        "Failed to resolve attributionToken: \(error!.localizedDescription)")
+                    return
+                }
+                if let data = data, let json = try? JSON(data: data) {
+                    let campaignId = json["campaignId"].stringValue
```
chung-lai, line 50: Looking at the doc, there are many attributes in the payload. Since the attributes are free formed, wonder if we should just loop through the json and log the entire payload? Unless we only care about these three fields

chung-lai (approved):
Meant to approve it :)

Client/Logging/ConversionLogger.swift
```diff
@@ -24,8 +26,49 @@ class ConversionLogger {
...
+            logAttributionData()
         } else {
             SKAdNetwork.updateConversionValue(event.rawValue)
         }
     }
+
+    private static func logAttributionData() {
+        if #available(iOS 14.3, *), let token = try? AAAttribution.attributionToken() {
+            // Kick-off a POST request to resolve the token.
+            var request = URLRequest(url: URL(string: "https://api-adservices.apple.com/api/v1/")!)
```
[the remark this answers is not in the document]
darinwf: Yeah, good idea.

Client/Logging/ConversionLogger.swift
```diff
@@ -24,8 +26,49 @@ class ConversionLogger {
...
+                    attributes.append(
+                        ClientLogCounterAttribute(
+                            key: LogConfig.AdServiceAttribute.AdGroupID,
+                            value: adGroupId))
+                    attributes.append(
+                        ClientLogCounterAttribute(
+                            key: LogConfig.AdServiceAttribute.KeywordID,
+                            value: keywordId))
+
+                    ClientLogger.shared.logCounter(
```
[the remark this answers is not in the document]
chung-lai: Without that, I think I don't see the attribute resolve event all the time 😢

Client/Logging/ConversionLogger.swift
darinwf: Oh, I could do that. This was just the set of fields that Sridhar requested. Just logging them all would mean we don't have to encode the names here. This code can be ignorant of those details. On the flip side, we could end up logging too much data if some of the other fields become long or we might worry about some of the other fields being data that we don't want to log??

Client/Logging/ConversionLogger.swift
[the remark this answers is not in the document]
darinwf: Oh good catch! I forgot that dataTask can run on a background thread.

Client/Logging/ConversionLogger.swift
chung-lai: That's true, may be we should only extract the keys defined in the documentation so we don't end up logging too much or if Apple had any server side issue later passing too big of a payload :D Think we should start with this list and add more as needed.

darinwf: I coded it up to just enumerate and report all. Maybe that is fine. Hmm...

PR #3290, opened 2022-03-31
tab switcher new design behind flag

chassu:
Not working yet, but running into an issue and would like some feedback. On onDataUpdated(), The old SingleLevelTabCardsView doesn't seem to be invalidated, which is also making the card transition animation broken. Not sure which part of the change is causing this. Also, scrolling the view also seems to update the switcher to the correct state. Not sure why that is happening either.

https://user-images.githubusercontent.com/87332917/161158612-9d27a943-7152-43cd-a2e4-5e15e133842c.mp4


chassu:
Closing a tab looks fine though. I'm aware that the section header should be gone when there's no tab to show for today. Will fix that.

https://user-images.githubusercontent.com/87332917/161165149-002effa0-2156-4372-9a67-b1daa3b124cd.mp4

chassu:
I'm trying to separate two issues I'm seeing:

1. Can't find the new tab I added when I go to the switcher
2. SingleLevelTabCardsView is getting rendered on the first scroll back to switcher (Just a slight scroll will do it). This existing on main build too.

1 is breaking the animation for now, but 2 is interesting, and I don't know where the update comes yet.

chassu:
Or actually, problem 2 isn't actually a problem if problem 1 never happens.

chassu:
Ok, I ran into some wax-on effect when fixing `selectedRowId` (@darinwf fun times 😉). Gonna do a bit refactoring and make sure the timeLineCardsView is not wrapped by some other view because that will seem to mess up the scrolling (a hard lesson learned about ScrollView a while back). I guess this problem turns out a bit tricky 😓

chassu:
I kind of spot the problem... with this new design, a selectTab event will could potentially change the rows. Needed to add a sink from selectedTabPublisher to CardModel.

For example, imagine a case where a user clicks on a tab from last week, the tab is going to get promoted to today's section.

darinwf:
> For example, imagine a case where a user clicks on a tab from last week, the tab is going to get promoted to today's section.

Makes sense. Maybe that change should be delayed until after the tab is fully visible. Then when the card grid is hidden, we can update the rows. It makes me wonder if `tabsUpdatedPublisher` is too coarse grained of a notification.

chassu:
I'm already trying to achieve this effect. Ignore the "Today" header for both section, the blue bordered section is meant to be Last Week's.

https://user-images.githubusercontent.com/87332917/161368069-6346c128-758b-4d56-94e2-a8e1540cf610.mp4

chassu:
by doing something like this....

```
        manager.selectedTabPublisher.sink { [weak self] _ in
            DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
                self?.updateRows()
            }
        }.store(in: &subscription)
```

so far it looks WAI

darinwf:
> by doing something like this....
> 
> ```
>         manager.selectedTabPublisher.sink { [weak self] _ in
>             DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
>                 self?.updateRows()
>             }
>         }.store(in: &subscription)
> ```
> 
> so far it looks WAI

Yes.... but without hardcoded delays (asyncAfter is almost always going to result in subtle bugs). Instead we can wait for `ContentVisibilityModel.showContent` to be true.

chassu:
so you're thinking a subscription that signals to CardModel after `ContentVisibilityModel.showContent` is changed?

darinwf:
> so you're thinking a subscription that signals to CardModel after `ContentVisibilityModel.showContent` is changed?

Yeah... it is adding some complexity to things of course. We might want to be smart about only calling updateRows when needed. We figure out how to let `TabCardModel` know when `showContent` changed, but maybe `TabCardModel` needs to consult some extra state it has that tells it that it should do work.

You might consider adding a flag on `TabCardModel` named `needsUpdateRows`. You could then set this to true when you see that an "old" tab gets selected. Then whenever the notification comes in about `showContent` being set to true, you could call `updateRows` if `needsUpdateRows` is true and then clear the flag.

chassu:
Investigating this animation glitch:

https://user-images.githubusercontent.com/87332917/162550040-0825d9cf-780d-4679-9986-2f438f66ff05.mp4

chassu:
One more thing I'll fix: I should initialize the last used time for tabs during restoration, if the value is nil then I'll initialize with `Date.nowMilliseconds`.

Client/Frontend/Browser/BrowserModel.swift
```diff
@@ -41,15 +41,25 @@ class BrowserModel: ObservableObject {
...
     }
 
     func showWithNoAnimation() {
+        if FeatureFlag[.enableTimeBasedSwitcher] {
+            gridModel.tabCardModel.contentVisibilityPublisher.send()
+        }
+        gridModel.scrollToSelectedTab()
         cardTransitionModel.update(to: .hidden)
         contentVisibilityModel.update(showContent: false)
+        gridModel.tabCardModel.contentVisibilityPublisher.send()
```
evancrow, line 62: Should this be here? Looks like you make the same call behind a flag right before this.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -43,16 +50,39 @@ class TabCardModel: CardModel {
...
 
     @Default(.tabGroupExpanded) private var tabGroupExpanded: Set<String>
+    var contentVisibilityPublisher = PassthroughSubject<Void, Never>()
+    var needsUpdateRows: Bool = false
 
     func updateRows() {
-        incognitoRows = buildRows(incognito: true)
-        normalRows = buildRows(incognito: false)
+        if FeatureFlag[.enableTimeBasedSwitcher] {
+            incognitoRows = buildRows(incognito: true, byTime: TimeFilter(rawValue: "Today"))
```
evancrow, line 58: Instead of using `rawValue`, you can just do `byTime: .today` or `byTime: .lastWeek`

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -24,12 +24,19 @@ protocol CardModel: ThumbnailModel {
...
+    case today = "Today"
+    case lastWeek = "Last Week"
+}
+
 class TabCardModel: CardModel {
     private var subscription: Set<AnyCancellable> = Set()
 
     private(set) var manager: TabManager
     private(set) var normalRows: [Row] = []
+    private(set) var normalRowsLastWeek: [Row] = []
```
evancrow, line 37: This might be fine for now, but I worry that if we add more time frames that this won't be scalable. What you might want to do is make `normalRows` and `incognitoRows` into a dictionary like so:

**init:**
`private(set) var normalRows: [TimeFilter: [Row]] = [:]`

**reading**:
`// get rows for today`
`normalRows[.today]`

**writing**:
`// write rows for today`
`normalRows[.today] = [// rows]`

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -43,16 +50,39 @@ class TabCardModel: CardModel {
...
+            normalRows = buildRows(incognito: false)
+        }
 
         // Defer signaling until after we have finished updating. This way our state is
         // completely consistent with TabManager prior to accessing allDetails, etc.
         self.objectWillChange.send()
     }
 
+    func getRows(incognito: Bool) -> [Row] {
+        var retVal: [Row] = []
```
evancrow, line 75: No need to shorten the name, took me a minute to understand what you meant. Should just name this `returnValue`.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -102,6 +152,10 @@ class TabCardModel: CardModel {
                     return details.id
                 case .tabGroupGridRow(let details, let range):
                     return details.allDetails[range].reduce("") { $0 + $1.id + ":" }
+                case .sectionHeader(let timeFilter):
+                    return timeFilter.rawValue == "Last Week"
```
evancrow, line 156: Ids don't necessarily need to be a hash like this. Can just make it `"today-header"` as long as it's unique.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -113,6 +167,8 @@ class TabCardModel: CardModel {
                     return details.isSelected
                 case .tabGroupGridRow(let details, let range):
                     return details.allDetails[range].contains { $0.isSelected }
+                case .sectionHeader(_):
```
evancrow, line 170: nit:
```suggestion
                case .sectionHeader:
```

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -124,6 +180,8 @@ class TabCardModel: CardModel {
                     return details.allDetails.count
                 case .tabGroupGridRow(_, let range):
                     return range.count
+                case .sectionHeader(_):
```
evancrow, line 183: nit:
```suggestion
                case .sectionHeader:
```

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -132,16 +190,49 @@ class TabCardModel: CardModel {
         var multipleCellTypes: Bool = false
     }
 
-    func buildRows(incognito: Bool) -> [Row] {
+    func filterTabByTime(tab: Tab, byTime: TimeFilter?) -> Bool {
```
evancrow, line 193: Should `byTime` be required? Would make sense since the whole point of this function is to get tabs for a time.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -132,16 +190,49 @@ class TabCardModel: CardModel {
...
+        let lastExecutedTime = tab.lastExecutedTime ?? Date.nowMilliseconds()
+        let minusOneDayToCurrentDate = Calendar.current.date(
+            byAdding: .day, value: -1, to: Date())
+        guard let startOfOneDayAgo = minusOneDayToCurrentDate else {
+            return true
+        }
+
+        switch byTime {
+        case .today:
+            return lastExecutedTime > Int64(startOfOneDayAgo.timeIntervalSince1970 * 1000)
```
evancrow, line 208: Where is this `1000` coming from? Be nice to have a comment and a variable that can be easily reused and changed.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -132,16 +190,49 @@ class TabCardModel: CardModel {
...
+                byAdding: .day, value: -7, to: Date())
+            guard let startOfLastWeek = minusOneWeekToCurrentDate else {
+                return true
+            }
+            return lastExecutedTime < Int64(startOfOneDayAgo.timeIntervalSince1970 * 1000)
+                && lastExecutedTime > Int64(startOfLastWeek.timeIntervalSince1970 * 1000)
+        }
+    }
+
+    func buildRows(incognito: Bool, byTime: TimeFilter? = nil) -> [Row] {
```
evancrow, line 221: nit: remove extra line

Client/Frontend/Browser/Card/CardsContainer.swift
```diff
@@ -40,17 +40,35 @@ struct TabGridContainer: View {
...
+            return isIncognito
+                ? tabModel.incognitoRows.first { $0.cells.contains(where: \.isSelected) }?.id
+                : tabModel.normalRows.first { $0.cells.contains(where: \.isSelected) }?.id
+        } else {
+            if isIncognito {
+                if let row = tabModel.incognitoRows.first(where: { row in
+                    row.cells.contains(where: \.isSelected)
+                }) {
+                    if row.index == 2 {
+                        //scroll to today header
```
evancrow, line 54: nit: space after `//`

Client/Frontend/Browser/Card/CardsContainer.swift
```diff
@@ -40,17 +40,35 @@ struct TabGridContainer: View {
...
+                    } else {
+                        return row.id
+                    }
+                }
+            } else {
+                if let row = tabModel.normalRows.first(where: { row in
+                    row.cells.contains(where: \.isSelected)
+                }) {
+                    if row.index == 2 {
+                        //scroll to today header
```
evancrow, line 65: ditto

Client/Frontend/Browser/Card/GridModel.swift
```diff
@@ -14,15 +14,23 @@ class GridModel: ObservableObject {
     @Published private(set) var pickerHeight: CGFloat = UIConstants
         .TopToolbarHeightWithToolbarButtonsShowing
     @Published var switcherState: SwitcherViews = .tabs {
+        willSet {
+            gridCanAnimate = true
+        }
+
         didSet {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
```
evancrow, line 22: I know I added this but I wonder if this really is the best way... don't really like using `DispatchQueue.main.asyncAfter`, especially for setting state variables, although the worst that might happen is there won't be an animation for a tenth of a second. Hm...

evancrow (commented):
Good work!

Client/Frontend/Browser/BrowserModel.swift
chassu: removed

Client/Frontend/Browser/Card/CardModels.swift
chassu: fixed

Client/Frontend/Browser/Card/CardModels.swift
chassu: fixed. Created the dictionary and it's only used when the feature flag is enabled.

Client/Frontend/Browser/Card/CardModels.swift
chassu: fixed

Client/Frontend/Browser/Card/CardModels.swift
chassu: fixed

Client/Frontend/Browser/Card/CardModels.swift
chassu: fixed

Client/Frontend/Browser/Card/CardsContainer.swift
chassu: fixed

Client/Frontend/Browser/Card/CardsContainer.swift
chassu: fixed

Client/Frontend/Browser/Card/CardModels.swift
chassu: it won't compile without the _ because the sectionHeader case takes in an argument

Client/Frontend/Browser/Card/CardModels.swift
chassu: ditto

Client/Frontend/Browser/Card/CardModels.swift
chassu: I thought of the same thing... but I couldn't think of a way to guard it in buildRows(). Because inside buildRows(), it only conditionally guards it when the feature flag is enabled:

```
        if FeatureFlag[.enableTimeBasedSwitcher] && if let byTime = byTime {
            return rows
        }

```
But I'm having problem to pass it here: 

```
(FeatureFlag[.enableTimeBasedSwitcher]
                    ? filterTabByTime(tab: tab, byTime: byTime) : true)
```

chassu: I think it should be fine to force unwrap here

Client/Frontend/Browser/Card/CardModels.swift
evancrow: Also remove the parentheses

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -229,7 +225,6 @@ class TabCardModel: CardModel {
         }
     }
```
evancrow, line 226: final nit: should add a new line back here (separate the functions, before line 227)

evancrow (approved):
LGTM - final nit

Issue #3282, opened 2022-03-31
Cookie Cutter Next Steps


evancrow:
- [x] Link `learn more` button (Cookie Cutter Settings)
- [x] Add `Toast` when Accept Non-essential Cookies options are all set to `false`
- [x] Update Clear Browsing Data UI
- [x] Support clearing Cookie Cutter Exclusion
- [x] Open Cookie Cutter settings from menu UI
- [x] Add the blocking script
- [x] Display the actual number of cookies blocked
- [x] Pass user preferences to script
- [x] Logging
- [x] Tests :)
- [x] Flip flag!

evancrow:
[Figma](https://www.figma.com/file/x32M9Egvsdvug1mHJvXuA8/Cookie-Cutter---mobile?node-id=0%3A1)

evancrow:
@hwiatneeva or @jlynch630 What is the link for the `learn more` button?

jlynch630:
The current link is https://help.neeva.com/hc/en-us/articles/4486326606355

Can probably update the language to include iOS, too

PR #3294, opened 2022-04-01
Reorganize Toast Structure + Cookie Cutter Stuff

evancrow:
A rather large PR for a rather small change:
1. Reworked the plumbing of `ToastViewManager` so you don't have to go through `SceneDelegate`.
2. Made it possible to show an `OverlayView` from `SettingsView` (needed a new view since the `SettingsView` is in an overlayed `UIWindow`)
3. Added the reset `Toast` when all allowed cookie options are set to false.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/37356984/161179804-06469043-9d9c-4e28-963b-72f47c2e89d9.mp4

## Issues addressed
Issue: #3282


Client/Application/SceneDelegate.swift
```diff
@@ -61,14 +56,13 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
         let profile = getAppDelegate().profile
 
-        self.bvc = BrowserViewController(profile: profile, scene: scene)
+        self.bvc = BrowserViewController(profile: profile, window: window!, scene: scene)
 
         bvc.edgesForExtendedLayout = []
         bvc.restorationIdentifier = NSStringFromClass(BrowserViewController.self)
         bvc.restorationClass = AppDelegate.self
 
         window!.rootViewController = bvc
```
darinwf, line 65: Should we be worried about creating a circular reference between `window` and `bvc`? Not an issue on phones since there is only one, but on tablets, we can have many `BVC` instances, right? This would result in memory leaking, I think.

darinwf (approved):
You might need to make sure the references to UIWindow are `weak`

Client/Application/SceneDelegate.swift
evancrow: Well on iPads we would also have multiple `SceneDelegate`s each with one `window` and one `bvc`, so no?

evancrow:
> You might need to make sure the references to UIWindow are `weak`

We don't store any reference to `window` actually, it's just passed and then de-initialized

PR #3300, opened 2022-04-04
Add condition for import WalletCore

buracules:
- Add #if XYZ control for import WalletCore
- Removed unnecessary files from Client

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Client/Frontend/Browser/BrowserViewController/BrowserViewController+ServerDelegate.swift
```diff
@@ -22,24 +25,30 @@ protocol WalletConnectPresenter: ModalPresenter {
 
 extension BrowserViewController: ServerDelegate, WalletConnectPresenter {
```
yozuysal, line 26: The actual long term fix for this should be to remove `BrowserViewController: ServerDelegate, WalletConnectPresenter` and make a new abstract class that is instantiated on construction of BVC. Then we can define the interface for that class and call it all across BVC and only if/def the actual implementation file.

yozuysal: So configureWallet, connectWallet(to...) etc would be all inside that interface

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -710,22 +710,24 @@ extension BrowserViewController: WKNavigationDelegate {
...
-                    // If this is only for invoking a wallet app with no params, cancel the navigation.
-                    decisionHandler(.cancel)
-                } else if let components = URLComponents(string: url.absoluteString),
-                    let uri = components.valueForQuery("uri"),
-                    let wcURL = WCURL(uri.removingPercentEncoding ?? ""),
-                    SceneDelegate.getBVC(with: tabManager.scene).connectWallet(to: wcURL)
-                {
-                    // If we can establish a connection through existing wallet, cancel the navigation.
-                    decisionHandler(.cancel)
+            #if XYZ
```
yozuysal, line 713: See above, if connectWallet is safe to call, we wouldnt need these

Client/Frontend/Browser/Card/CardGrid.swift
```diff
@@ -168,12 +168,14 @@ struct CardGrid: View {
...
-                    .environment(\.cardSize, cardSize)
-                    .environment(\.aspectRatio, CardUX.DefaultTabCardRatio),
-                isActive: $walletDetailsModel.showingWalletDetails
-            ) {}
+            #if XYZ
+                NavigationLink(
+                    destination: WalletDetailView()
+                        .environment(\.cardSize, cardSize)
+                        .environment(\.aspectRatio, CardUX.DefaultTabCardRatio),
+                    isActive: $walletDetailsModel.showingWalletDetails
```
yozuysal, line 176: For now we wont be showing wallet details anywhere, so we can just delete all this (WalletDetailsModel itself and the state as well)

Client/Frontend/TabChrome/TopBar/LocationView/TabLocationView.swift
```diff
@@ -127,49 +129,53 @@ struct TabLocationView: View {
...
-                                        showReaderModeSettings = false
-                                    })
-                                    .environmentObject(readerModeModel)
-                                    .environmentObject(BrightnessModel())
-                                    .environmentObject(
-                                        TextSizeModel(
-                                            webView: (readerModeModel.tabManager
-                                                .selectedTab?
-                                                .webView)!))
+                            #else
```
yozuysal, line 136: this is not an if/def I believe. xyz uses reader mode too. As I mentioned, our goal should be to make it OK to include @EnvironmentObject private var web3Model: Web3Model without any worries since that class is in Frontend/

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -118,10 +118,13 @@ class ZeroQueryModel: ObservableObject {
...
         if !Defaults[.signedInOnce] {
-            if NeevaConstants.currentTarget == .xyz && Defaults[.cryptoPublicKey].isEmpty {
-                promoCard = .walletPromo {
-                    self.bvc.web3Model.showWalletPanel()
-                }
+
+            if Defaults[.cryptoPublicKey].isEmpty {
+                #if XYZ
+                    promoCard = .walletPromo {
+                        self.bvc.web3Model.showWalletPanel()
```
yozuysal, line 125: since showWalletPanel is if/def guarded, this shouldn't be needed?

content-blocker-lib-ios/src/TabContentBlocker+ContentScript.swift
```diff
@@ -34,9 +37,9 @@ extension TabContentBlocker {
             components.scheme = "http"
             guard let url = components.url else { return }
 
-            if NeevaConstants.currentTarget == .xyz {
+            #if XYZ
                 WalletConnectDetector.shared.detectWalletConnect(for: url, in: mainDocumentUrl)
```
yozuysal, line 41: I think it is better to push the if/def to WalletConnectDetector. As a guiding principle, if we can get out of this with all if/def inside Neevawallet, that is a win.

buracules: WalletConnectDetector is inside WalletCore. I can't call function without if/def.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -492,12 +497,14 @@ class BrowserViewController: UIViewController, ModalPresenter {
             if Self.createNewTabOnStartForTesting {
                 self.tabManager.select(self.tabManager.addTab())
             } else if self.tabManager.normalTabs.isEmpty {
-                if NeevaConstants.currentTarget == .xyz {
+                #if XYZ
```
yozuysal, line 500: these should stay as 'if NeevaConstants.currentTarget == .xyz {' maybe? Since they can compile fine

Client/Frontend/Browser/BrowserViewController/BrowserViewController+OverflowMenu.swift
```diff
@@ -168,7 +168,9 @@ extension BrowserViewController {
             )
             showFeedbackPanel(bvc: self, screenshot: image ?? self.feedbackImage)
         case .cryptoWallet:
-            web3Model.showWalletPanel()
+            #if XYZ
+                web3Model.showWalletPanel()
```
yozuysal, line 172: I think this if/def is also unnecessary? (showWalletPanel compiles as is)

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -724,6 +724,7 @@ extension BrowserViewController: WKNavigationDelegate {
 
                 }
             }
+
```
yozuysal, line 727: extra line

Client/Frontend/Browser/Card/CardListViews.swift
```diff
@@ -4,7 +4,10 @@
 
 import Shared
 import SwiftUI
-import WalletCore
```
yozuysal, line 7: is this needed? what happens if you delete this completely?

Client/Frontend/Browser/Card/SwitcherToolbarView.swift
```diff
@@ -5,7 +5,10 @@
 import Defaults
 import Shared
 import SwiftUI
-import WalletCore
+
+#if XYZ
+    import WalletCore
```
yozuysal, line 10: since defaultBackgroundOrTheme is defined in Frontend/ I think this one also isnt needed

Client/Frontend/NeevaWallet/Web3Model.swift
```diff
@@ -540,3 +562,15 @@ class Web3Model: ObservableObject, ResponseRelay {
...
     @Published var showingWalletDetails = false
 }
+
+#if XYZ
+    extension Web3Model: WalletServerManagerDelegate {
+        func shouldShowToast(for message: LocalizedStringKey) {
+            self.toastDelegate?.shouldShowToast(for: message)
+        }
+
+        func getWeb3Model() -> Web3Model {
```
yozuysal, line 572: instead of giving access to the whole Web3Model, lets do 

1. WalletServerManagerDelegate: ResponseRelay so that we can pass the delegate itself as relay
2. Add updateCurrentSession() to the WalletServerManagerDelegate interface
3. Add updateCurrentSequence(sequence) to the WalletServerManagerDelegate interface

Client/Frontend/ZeroQuery/ZeroQueryView.swift
```diff
@@ -6,7 +6,10 @@ import Defaults
 import Shared
 import Storage
 import SwiftUI
-import WalletCore
+
+#if XYZ
```
yozuysal, line 10: is this needed? I think you can remove this completely

Client/Frontend/NeevaWallet/Web3Model.swift
```diff
@@ -57,7 +57,18 @@ class Web3Model: ObservableObject, ResponseRelay {
...
 
-    let server: Server?
+    #if XYZ
+        var serverManager: WalletServerManager?
+        weak var toastDelegate: ToastDelegate?
+    #endif
+
+    var server: Server? {
+        #if XYZ
+            return server?
```
yozuysal, line 67: typo?

buracules: Yes it is. It must be `serverManager?.server`.

Client/Frontend/NeevaWallet/Web3Model.swift
buracules: I also have to access wallet somehow. So I will use `func getWallet() -> WalletAccessor?`. I only need 2 properties so I can create 2 functions and get publicAdress and walletMeta as well.

yozuysal (approved):

PR #3305, opened 2022-04-05
Cookie Cutter UI P2

evancrow:
- Added the ability to open the Cookie Cutter settings from the tracking menu
- Link the `learn more` button
- Update the `Tracking UI`

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Issue: #3282

### DIFF
--- a/Client/Frontend/Settings/Privacy/DataManagementView.swift
+++ b/Client/Frontend/Settings/Privacy/DataManagementView.swift
@@ -13,6 +13,7 @@ enum ClearableDataType: String, Identifiable, Codable, CaseIterable {
     case cookies = "Cookies"
     case trackingProtection = "Tracking Protection"
     case dapps = "Connected dApps"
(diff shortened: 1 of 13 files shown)


Client/Frontend/Settings/Privacy/DataManagementView.swift
```diff
@@ -83,7 +89,16 @@ struct DataManagementView: View {
             Section(header: Text("Data on This Device")) {
                 ForEach(
                     ClearableDataType.allCases.filter {
-                        $0 == .dapps ? NeevaConstants.currentTarget == .xyz : true
+                        switch $0 {
+                        case .trackingProtection:
```
darinwf, line 93: I see, so the idea here is to deprecate the `trackingProtection` option in favor of `cookieCutterExclusions`. Makes sense.

darinwf (approved):

PR #3308, opened 2022-04-05
Add XYZClient to CircleCI

buracules:
- Create a new step for building XYZClient 
- Add new step to the jobs


buracules:
I forgot to update step name. I will update it as well when i figure our how to fix this.

darinwf (dismissed):
LGTM

Want to get @chung-lai's eyes on this too just at least as a heads up about changes to the CI config

buracules:
@darinwf it's failing on CircleCI right now. Am I missing something?

darinwf:
Hmm.. I'm not sure why it is failing.

chung-lai:
@buracules, i think `unit test/ui test` build is failing on XYZClient. Don't think we have run unit test on XYZClient yet? @yozuysal.

May be test build has been failing the whole time and we need to fix that. Also, could we run XYZClient as it's own job so it doesn't fail the entire build in the case XYZClient fails to complete.

buracules:
FYI: I canceled the Client build it's not failing

chung-lai:
> Right now I am getting the following error:
> 
> `./Client/APIConfig.xcconfig:1:1: error: unable to open file (in target "XYZClient" in project "Client")`
> 
> How is Client building without APIConfig file? it is not there for Client as well. @chung-lai

Chatted offline and we will add a pre-action to the xyz scheme

bootstrap.sh
```diff
@@ -35,4 +35,4 @@ swift build -c release
 popd
 
 # Link githooks
-./Scripts/link-git-hooks.sh
\ No newline at end of file
+./Scripts/link-git-hooks.sh
```
chung-lai, line 38: nit: undo this change? may be just extra space

chung-lai (approved):
LGTM!

PR #3315, opened 2022-04-06
Fix Memory Issue from Indexing Space Entities

### LINKED ISSUE #3155: Large memory usage when we index space items for spotlight

EdwardJTL:
We had previously omitted adding images to indexed spotlight entities due to a memory issue.

The issue is coming from the routine to download the page level favicon as a fall back; for some reason, calls to those when the app re-enters foreground ends up increasing the app's total memory size without bounds.

This change adds a lazy global UIImage that is the same as the polar bookmark image as a fallback, replacing the favicon.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested manually by observing the app's memory usage while going between foreground and background.

### Screenshot
Screen Shot 2022-04-06 at 3 08 41 PM

The memory spikes when the app enters foreground and it pulls all the PNG data for space entities.
Notice how the app always returns to around the same level after it enters background.

## Issues addressed
closes #3155


darinwf:
Are we fetching all of those images over the network each time the app starts up?

EdwardJTL:
> Are we fetching all of those images over the network each time the app starts up?

Yes I believe so.

for entities with a url for the thumbnail, the images are fetched with the shared SDWebImageManager, which should share the cache with the rest of the app. On `main`, these fetches only happen if the space items need to be displayed.

we could do a permanent on disk cache?

darinwf:
Yeah, I'm wondering about persistent caching... I believe SDWebImage should have a mode to enable that. Maybe we should be using that. Can be a separate issue for sure. Seems inefficient to fetch these at startup each time.

darinwf (approved):

darinwf (approved):
fromCacheOnly, cool!

EdwardJTL:
> Yeah, I'm wondering about persistent caching... I believe SDWebImage should have a mode to enable that. Maybe we should be using that. Can be a separate issue for sure. Seems inefficient to fetch these at startup each time.

I just tested it, and it looks like there might be some cache that persists across app launches; the network usage at start up is the same with or without the fetching code in spotlight store. In any case, i added the `fromCacheOnly` flag here so that we only query the cache.

I also observed that there is always a call to `https://neeva.com/graphql/GetSpacesData` at app launch, and the response is 14MB (cuz i have a lot of spaces); i suspect this is because for some images the data is in the response. I suppose this is another area for future work

darinwf:
I think we fetch spaces data on startup so we can properly fill-in the "add to spaces" button on the bottom bar. We also do not have a persistent store of spaces on the client. Adding one may be a good future optimization.

Issue #3292, opened 2022-04-01
Rebuild Recently Closed Menu in SwiftUI


evancrow:
One of the last remaining `UIKit` views, should modernize the `RecentlyClosedTabsPanel` using `SwiftUI` and also make it conform to the rest of the app style.

evancrow:
This should also include the `HistoryPanel`

PR #3322, opened 2022-04-07
SwiftUI History

evancrow:
(behind flag)

Rebuilds the history panel as a SwiftUI view and creates a separate view model.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases

## Issues addressed
Fixed: #3292


EdwardJTL:
The menu that pops up from "Clear Recent History..." had a text explaining what it will clear. can we keep that?
also can we have an arrow on the recently closed button since it should navigate to another view?

Client/Frontend/Browser/Tabs/TabMenu.swift
```diff
@@ -124,4 +126,65 @@ struct TabMenu {
...
+                        .renderingMode(.template)
+                }
+            }
+        }
+    }
+
+    func swiftUIOpenInNewTabMenu(_ url: URL) -> some View {
+        return Group {
+            Button {
+                let tab = self.tabManager.addTabsForURLs(
```
darinwf, line 163: nit: Can you factor the guts of the action handler out into a local function that is just parameterized by `incognito`? That way you avoid code duplication between incognito and non-incognito cases.

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
```diff
@@ -0,0 +1,143 @@
...
+            return deferMaybe(FetchInProgressError())
+        }
+
+        isFetchInProgress = true
+
+        return profile.history.getSitesByLastVisit(
+            limit: queryFetchLimit, offset: currentFetchOffset) >>== { result in
+                // Force 100ms delay between resolution of the last batch of results
+                // and the next time `fetchData()` can be called.
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
```
darinwf, line 66: This is very complicated. Is there a way to achieve this without using `asyncAfter`?

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
```diff
@@ -0,0 +1,143 @@
...
+    private let profile: Profile
+    let tabManager: TabManager
+
+    private let queryFetchLimit = 100
+    private var currentFetchOffset = 0
+    private var isFetchInProgress = false
+
+    @Published var groupedSites = DateGroupedTableData<Site>()
+
+    var recentlyClosedTabs: [SavedTab] {
```
darinwf, line 39: Awesome switching to storing `SavedTab` objects here. Unifies behavior with the other forms of undo-close-tab. Love it!

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
```diff
@@ -0,0 +1,143 @@
...
+        currentFetchOffset = 0
+
+        loadSiteData().uponQueue(.main) { result in
+            self.addSiteDataToGroupedSites(result)
+        }
+    }
+
+    func loadSiteData() -> Deferred<Maybe<Cursor<Site?>>> {
+        guard !isFetchInProgress else {
+            return deferMaybe(FetchInProgressError())
```
darinwf, line 57: So what happens if this fails? It means `reloadData` might fail right? What are the callers of `reloadData` supposed to do in that case? I don't see any handling of the error caused by `isFetchInProgress` being true when `loadSiteData` is called.

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
```diff
@@ -0,0 +1,143 @@
...
+                    // we don't keep persistent identifiers to the activities so we can only delete all
+                    UserActivityHandler.clearIndexedItems()
+                    self.reloadData()
+                }
+            }
+        } else {
+            self.profile.history.clearHistory().uponQueue(.main) { _ in
+                UserActivityHandler.clearIndexedItems()
+
+                self.reloadData()
```
darinwf, line 117: this call to `reloadData` can fail. what should we do if that happens?

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
```diff
@@ -0,0 +1,143 @@
...
+        if let hours = timeFrame.hours {
+            if let date = Calendar.current.date(byAdding: .hour, value: -hours, to: Date()) {
+                let types = WKWebsiteDataStore.allWebsiteDataTypes()
+                WKWebsiteDataStore.default().removeData(
+                    ofTypes: types, modifiedSince: date, completionHandler: {})
+
+                self.profile.history.removeHistoryFromDate(date).uponQueue(.main) { _ in
+                    // we don't keep persistent identifiers to the activities so we can only delete all
+                    UserActivityHandler.clearIndexedItems()
+                    self.reloadData()
```
darinwf, line 110: this call to `reloadData` can fail. what should we do if that happens?

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
```diff
@@ -0,0 +1,143 @@
...
+
+                self.reloadData()
+                self.tabManager.recentlyClosedTabs.removeAll()
+            }
+        }
+    }
+
+    func removeHistoryForURLAtIndexPath(site: Site) {
+        profile.history.removeHistoryForURL(site.url).uponQueue(.main) { result in
+            self.reloadData()
```
darinwf, line 125: ditto

Client/Frontend/Site/SiteListView.swift
```diff
@@ -0,0 +1,54 @@
...
+        GroupedCell.Decoration {
+            VStack(spacing: 0) {
+                if let sites = sites {
+                    ForEach(
+                        Array(sites.enumerated()), id: \.element
+                    ) { index, site in
+                        SiteRowView(tabManager: tabManager, site: site) {
+                            tappedItemAtIndex(index)
+                        }.onAppear {
+                            itemAtIndexAppeared(index)
```
darinwf, line 29: This strategy of loading data incrementally looks really solid. Nice!

darinwf (commented):
Great work on this! Just one thing that worries me...

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
evancrow: Maybe? Really all of this logic is from the Firefox code so I understand just as much as you.

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
evancrow: We could show an error Toast? Again, this logic is from Firefox which didn't have a fallback method.

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
evancrow: If `isFetchInProgress` is `true`, then that means `loadData` is already active so the data will already be loaded, just by a different caller.

darinwf: My worry was that it might be stale data that gets shown. Maybe we should instead set a flag indicating that it is necessary to load the data again after the current load completes.

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
darinwf: I don't think we want to tell the user about it. I just want to make sure we handle the edge cases gracefully.

Client/Frontend/History/SwiftUI/History/HistoryPanelModel.swift
darinwf: Got it. I didn't realize this was copied from the Firefox code. I am mostly worried about showing stale state to the user if a mutation occurs and we call reloadData but that gets ignored because it was already in progress. The in progress reload could still only be showing old data. So it could make sense to set a flag to indicate that a reload is needed again after the current one continues.

darinwf (approved):
LGTM, thanks!

PR #3323, opened 2022-04-07
Removed WalletCore from Client

buracules:
- WalletCore removed completely from Client.

Web3Model is still accessible from Client. But only with one property. I tried to use compiler flags as little as possible inside BrowserViewController. 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Client/Frontend/Settings/Clearables.swift
```diff
@@ -87,15 +87,17 @@ class TrackingProtectionClearable: Clearable {
     }
 }
 
-class ConnectedDAppsClearable: Clearable {
-    func clear() -> Success {
-        for session in Defaults[.sessionsPeerIDs] {
-            Defaults[.dAppsSession(session)] = nil
+#if XYZ
```
yozuysal, line 90: this can be target == .xyz I think?

Client/Frontend/NeevaWallet/Web3Model.swift
```diff
@@ -2,364 +2,366 @@
...
+    import web3swift
+    import WalletConnectSwift
+    import BigInt
+    import WalletCore
+#endif
 
-    var publicAddress: String {
-        wallet?.publicAddress ?? ""
-    }
+class WalletDetailsModel: ObservableObject {
```
yozuysal, line 19: this is unused now I think we can remove it

Client/Frontend/NeevaWallet/Web3Model.swift
```diff
@@ -2,364 +2,366 @@
...
 import SwiftUI
-import WalletConnectSwift
-import WalletCore
 import XCGLogger
-import web3swift
 
-class Web3Model: ObservableObject, ResponseRelay {
-    let openURLForSpace: (URL, String) -> Void
-    let openURL: (URL) -> Void
+#if XYZ
```
yozuysal, line 12: what would happen if instead of the giant ifdef here, we had 

Web3Model as everything outside of the if def and Web3ModelImpl:Web3Model as everything inside. Then inside bvc we create Web3Model() and web3ModelImpl() inside the if def again. And Web3ModelImpl is not included inside the client target. wdyt? (This just makes reading and making changes on web3model easier as I am sure we are gonna do quite a bit)

buracules: Oh now i get it. Yes that will clean the code a lot. After fixing this Test fails i will do that.

yozuysal:
As we discussed offline, let's land this as is with a TODO to revisit Web3Model setup. I think this is a concrete win for Client target and we should take it. We can improve Web3Model code development with a few changes as a followup

yozuysal (approved):

PR #3325, opened 2022-04-07
Fix BottomBar Not Hiding 100% When Scrolling

evancrow:
Caused by a weird bug where `safeArea` would be 0. Just passed the `safeArea` from inside the `GeometryReader` instead of using a separate modifier.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

After:

https://user-images.githubusercontent.com/37356984/162333609-bc87d50c-df1c-4125-89c9-0ff01b01de6c.mp4

Before:

https://user-images.githubusercontent.com/37356984/162333637-0b2504a3-0666-4453-9989-454204b30ddd.mp4


darinwf:
Awesome, thanks for trying a better fix on this one. I'll test it out on my iPhone 6s later tonight to see if I can repro.

chung-lai:
Downloaded last week's release and looks like the issue was already  there. If we are confident with the fix, we can cherry pick this along with PR #3327 and do a hot fix.

darinwf:
I think we should give more bake time for this one just in case.

darinwf (approved):
This seems to work well!

darinwf:
Can mark this as fixing https://github.com/neevaco/neeva-ios/issues/3331

darinwf:
After testing this some more, I'm comfortable taking this on the release branch.

PR #3329, opened 2022-04-08
Avoid putting all search tabs in the same tab group

### LINKED ISSUE #3330: All of my searches are ending up in a single tab group
@michellew7 - I think what you showed me was a real bug, and I've seen it too now. Will get this fixed.

darinwf:
The issue here is that normalizedHostAndPathForDisplay strips out the
query parameters, making it so that all neeva.com/search URLs will
look identical.

Instead, we should use the equals(:with) function to perform the
comparison, which is the same function used by TabManager.getTabFor.

### How was this tested?
- [x] I tested this manually

I plan to write an automated test as a follow-up.

Fixes #3330


chung-lai (approved):

darinwf:
The test I added is not perfect. It doesn't quite cover this scenario, but it is still a good test, so I added it. Will think more about how to write a test that would have failed given the repro I had for this bug.

Issue #3267, opened 2022-03-29
When neeva.com redirects to /signin, we should display the native sign-in UI


darinwf:
Currently, we just have the user see the web sign-in screen. We currently only intercept when starting a navigation to /signin, but the web client can issue a pushState (or equivalent) to /signin which WebKit would not see as starting a new navigation. We probably just need to extend our code in WKNavigationDelegate.didFinish.

darinwf:
neeva.com may decide to redirect a user to `/signin` if it determines that the user hasn't used Neeva recently enough. Requiring re-auth like this is a security measure. Therefore, the app needs to be able to handle this case properly and show the native sign-in flow.

darinwf:
This is pretty relevant for users who have been using the product for ~6 months or if they have been inactive for a period of time. Would be nice to make it lower friction to re-auth.

jedireza:
Moved to ClickUp 🔒 https://app.clickup.com/t/3w33cmx

PR #3336, opened 2022-04-08
Move Neeva /signin Redirect Code

evancrow:
Makes sure the redirect code is called for all navigation paths.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #3267


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -817,6 +790,30 @@ extension BrowserViewController: WKNavigationDelegate {
             }
         }
 
+        // for Neeva signup or signin page, open native auth panel
+        if let url = responseURL, NeevaConstants.isAppHost(url.host), request?.httpMethod == "GET" {
```
darinwf, line 794: I'm not sure if this covers all of the cases. I believe javascript in the page will do the equivalent of `history.replaceState("/signin")` to change the URL. In other words, navigations to `/signin` won't always correspond to a network response. I wonder if we need this logic to live in `BVC.navigateInTab` instead as that is the choke point for all ways in which the URL can change.

evancrow: Turns out `BVC.navigateInTab` is called by the same function that I moved the code to: `webView(_ webView: WKWebView, didFinish navigation: WKNavigation!)`. Also, not a simple way to move it out of this function, since we need to be able to cancel the navigation

darinwf: `BVC.navigateInTab` is called in other cases too. I'm worried we might miss some of those cases if we only handle this inside `didFinish`. Maybe we can do some testing to be sure of that. We'd want to configure a bad `httpd~login` cookie locally and see how the server redirects the client to `/signin` and make sure that we handle that case well.

evancrow: Sorry, I meant that `navigateInTab` is called _only_ by `navigationDidFinish`. (At least that's what Xcode told me :))

darinwf: See here in `BrowserViewController.swift`:

Screen Shot 2022-04-17 at 10 40 27 PM

evancrow: Thank you for finding that. Will move it now.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1138,6 +1138,45 @@ class BrowserViewController: UIViewController, ModalPresenter {
                 }
             }
         }
+
+        // if on neeva search result path and never signed in, user is in preview
```
darinwf, line 1142: I don't think we want to move this code. I think the goal was to move the code that observes a navigation to `/signin` and have the app show the native sign in experience instead of the web `/signin` experience. The code below is what we use to show the account creation overlay when the user hits preview mode limits. I think that code is appropriate to have in `didFinish`.

evancrow: I really just copy/pasted the wrong thing 🤩

jedireza:
I found this PR going through the onboarding issue; #3267 in particular. I noticed this PR is about a month old now. Luckily there aren't any merge conflicts. I didn't go through the conversation in detail. Are there any blockers on this?

evancrow:
I think we should close this actually. @darinwf and I discussed and moving this code to `navigatedInTab` wouldn't work because we don't have the parameters to pass to the redirect code.

PR #3337, opened 2022-04-08
Determine new value of showGrid synchronously

### LINKED ISSUE #3272: Bottom bar stuck in tab switcher mode while web page and URL bar is showing

darinwf:
This should resolve the bugs we've seen with the bottom bar appearing to be
stuck as the tab switcher bottom bar when viewing a page, or the less common
issue with a tab appearing to not be scrollable but just jiggle around as you
try to scroll it w/ the card grid scrolling below it.

The issue is that when the DispatchQueue.main.async callback runs, the value of
cardTransitionModel.state could have already had its value changed to a value
other than visibleForTrayHidden, which would cause us to not set showGrid to
false.

BrowserModel.onCompletedCardTransition would never run if showGrid did not get a
new value, and the bottom bar would also continue to think that it should keep
showing the bottom bar for the grid. These are the bugs we were observing, and
it is explained by the above.

Fixes #3272

### How was this tested?
- [x] I tested this manually


evancrow (approved):
Makes sense!

darinwf:
I think there may still be some additional bugs related to this code. The `DispatchQueue.main.async` creates the potential for other ordering issues. For example, there are probably bugs if some code calls `showSpaces` while this callback is pending. `showSpaces` sets `showGrid` to true, but this callback could run after it and then set it back to false. Maybe we need to have some kind of sequence number associated with attempts to set showGrid to a new value so we can ensure proper sequencing.

darinwf:
> I think there may still be some additional bugs related to this code. The `DispatchQueue.main.async` creates the potential for other ordering issues. For example, there are probably bugs if some code calls `showSpaces` while this callback is pending. `showSpaces` sets `showGrid` to true, but this callback could run after it and then set it back to false. Maybe we need to have some kind of sequence number associated with attempts to set showGrid to a new value so we can ensure proper sequencing.

Another choice could be to always only change `showGrid` via `DispatchQueue.main.async`. That would also guarantee ordering. Just not sure that it wouldn't create other issues :-/

PR #3342, opened 2022-04-09
Prototype moving scrollTo calls earlier

darinwf:
@chassu - Here's an exploration of moving the scrollTo calls earlier and delaying other steps, involved in transitioning between content and cardgrid, until after the scrollTo has happened. This uses a dreaded `asyncAfter` so I'm not happy about it. The consequential change here is moving scrollTo ahead of other work related to the transition. I'm curious how this would impact the issue we were seeing on your PR.

### DIFF
--- a/Client/Frontend/Browser/BrowserModel.swift
+++ b/Client/Frontend/Browser/BrowserModel.swift
@@ -41,13 +41,16 @@ class BrowserModel: ObservableObject {
         if gridModel.tabCardModel.allDetails.isEmpty {
             showWithNoAnimation()
         } else {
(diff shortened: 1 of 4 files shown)


darinwf:
@chassu - I assume something like this PR would still be helpful for the new sectioned tab switcher design. I have some more ideas for this PR that I want to try out.

chassu:
Hmm, I actually included them as part of #3290 already. What are the ideas that you're thinking about?

darinwf:
Closing this PR since it is already in the tree.

PR #3350, opened 2022-04-11
Fix CardTransition iPad

evancrow:
We were always using the full width of the device, even if on `iPad`, the window was not. This just changes how the width was calculated. I was trying to re-create the issue the comment laid out, but it doesn't seem to occur. 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

**Before:**

https://user-images.githubusercontent.com/37356984/162842932-ca6a449b-dc23-4f1e-92b0-f81d4efb1810.mp4

**After:**

https://user-images.githubusercontent.com/37356984/162842942-2aa5483e-d9c6-45e8-8e0d-2b5ba0afc9b2.mp4

iPhone:

https://user-images.githubusercontent.com/37356984/162842945-105fc7e0-3511-4497-aae4-79031a865317.mp4


chung-lai:
The fix makes sense to me.  @chassu left the original comment, would love a look from him

darinwf:
Be sure that landscape card transitions still look right. Make sure the
card is scaled properly. I believe there was a problem before when the code
was originally as this PR makes it.

evancrow:
@darinwf Oh right, now I remember the original issue. I'll set it up to only use the `containerFrame` on iPad, since that's where the issue I was trying to fix occurred anyways...

chung-lai (approved):

PR #3351, opened 2022-04-11
Fix Logging Crash When Opening NeevaScope

evancrow:
Logging when opening NeevaScope for the first time would sometimes crash. This unwraps the values to make sure that it doesn't crash.

FYI @chassu and @EdwardJTL 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### DIFF
--- a/Client/Logging/ClientLogger.swift
+++ b/Client/Logging/ClientLogger.swift
@@ -69,7 +69,9 @@ public class ClientLogger {
 
         #if DEBUG
             if !Defaults[.forceProdGraphQLLogger] {
-                let attributes = loggingAttributes.map { "\($0.key! ?? "" ): \($0.value! ?? "")" }
+                let attributes = loggingAttributes.compactMap {
+                    "\(String(describing: $0.key)): \(String(describing: $0.value))"
+                }
                 let path = path.rawValue
                 debugLoggerHistory.insert(
                     DebugLog(


EdwardJTL:
@chung-lai this happens when we call logCounter with a `ClientLogCounterAttribute` initialized without a value. I used to just remove the force unwrapping locally. Not sure if you had other plans for how graphql logging should work in debug

Client/Logging/ClientLogger.swift
```diff
@@ -69,7 +69,9 @@ public class ClientLogger {
 
         #if DEBUG
             if !Defaults[.forceProdGraphQLLogger] {
-                let attributes = loggingAttributes.map { "\($0.key! ?? "" ): \($0.value! ?? "")" }
+                let attributes = loggingAttributes.compactMap {
+                    "\(String(describing: $0.key)): \(String(describing: $0.value))"
```
chung-lai, line 73: thanks for fixing it! Totally slipped my mind. I think it would be great if we do assertion here as we should not have empty value imo.

EdwardJTL: if we want to assert that attributes cannot have empty values, then we should fix the initializer to not allow that either

evancrow: We can't change the `init` since it's declared in `API.swift`. Is that right @chung-lai?

evancrow: I think an `assert` might be fine in this case

evancrow:
@chung-lai @EdwardJTL Think I might just close this issue actually. Seems like the intended action is just for it to crash if the value is nil, and it's only on debug builds.

chung-lai:
> @chung-lai @EdwardJTL Think I might just close this issue actually. Seems like the intended action is just for it to crash if the value is nil, and it's only on debug builds.

Makes sense, thank you!

PR #3354, opened 2022-04-11
Use work item release freq for spotlight images

EdwardJTL:
This makes the queue release memory as soon as each "block" is done, instead of holding onto them when there are consecutive work items, which is what caused memory to pile up after repeated requests to index images.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
spam the tab switcher button and watch memory

Screen Shot 2022-04-11 at 6 44 36 PM


chung-lai (approved):
the change seems reasonable and the memory pattern looks much better! Thanks.

(Removed cherry pick label as we haven't branched yet)

PR #3357, opened 2022-04-12
Reconnect Download to DownloadQueue

EdwardJTL:
We override the `Download.delegate` property in `ToastDefaults`, which breaks the delegate chain. As a result, none of the delegate methods in `DownloadQueue` and the queue delegate get called. 

Added a separate property to reconnect this.


Client/Frontend/Browser/DownloadQueue.swift
```diff
@@ -4,13 +4,14 @@
...
 
-protocol DownloadDelegate {
+protocol DownloadDelegate: AnyObject {
     func download(_ download: Download, didCompleteWithError error: Error?)
     func download(_ download: Download, didDownloadBytes bytesDownloaded: Int64)
     func download(_ download: Download, didFinishDownloadingTo location: URL)
 }
 
 class Download: NSObject, ObservableObject {
+    fileprivate(set) weak var parentQueue: DownloadDelegate?
```
evancrow, line 14: Can you add a quick help comment (three dashes `///`) to each of those so someone else reading this knows what each one does. Thanks!

evancrow (approved):
LGTM, just one suggestion before you merge

PR #3365, opened 2022-04-12
Fix openSettings EnvironmentKey

evancrow:
Needed to be possible to pass page as a parameter or else only one page would be opened.


chassu (approved):
LGTM!

PR #3369, opened 2022-04-13
Gas Fee Indicator for Wallet

buracules:
- Added Gas Fee functionality to wallet under `Show Gas Fee UI` featureflag

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

Simulator Screen Shot - iPhone 13 Pro - 2022-04-13 at 21 46 22
Simulator Screen Shot - iPhone 13 Pro - 2022-04-13 at 21 46 18


Client/Frontend/NeevaWallet/Web3Toolbar.swift
```diff
@@ -14,6 +14,7 @@ struct Web3Toolbar: View {
     private let overFlowMenuAction: () -> Void
     private let showTabsAction: () -> Void
     private let openLazyTabAction: () -> Void
+    @EnvironmentObject var model: Web3Model
```
yozuysal, line 17: hmm, might be good to find a different way of getting this model (Now web3toolbar will start listening on all web3model changes). maybe we pass it in the constructor and keep it as var rather than @EnvironmentObject var?

WalletCore/GasFeeModel.swift
```diff
@@ -0,0 +1,70 @@
...
+            self = .medium
+        default:
+            self = .high
+        }
+    }
+
+    public var tintColor: Color {
+        switch self {
+        case .low:
+            return .brand.green
```
yozuysal, line 28: Color(light: .brand.green, dark: .brand.adaptive.green) and so on for the ones below.

yozuysal (commented):
looks mostly ready to go. a few nits.

Client/Frontend/NeevaWallet/GasFeeView.swift
```diff
@@ -0,0 +1,118 @@
...
+
+    var text: String {
+        switch self {
+        case .never: return "Never"
+        case .gwei40: return "<40 gwei"
+        case .gwei60: return "<60 gwei"
+        }
+    }
+
+    var selectedText: String {
```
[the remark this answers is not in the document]
buracules: It was for title actually but I couldn't find a way make title different from menu item. I am justgonna delete this.

Client/Frontend/NeevaWallet/Web3Toolbar.swift
buracules: TabToolbarHost and TabToolbarView doesn't have web3model that's why i implemented here. But it is just a color code for tabBar I can also use defaults for this?

WalletCore/GasFeeModel.swift
buracules: There is no .brand.adaptive.green. Can i get this color codes maybe i can create this colors?

yozuysal: sorry brand.variant is what I meant . All of these are defined in NeevaColors.swift in Shared

Client/Frontend/NeevaWallet/Web3Toolbar.swift
yozuysal: let's keep this for now.

yozuysal (approved):
lgtm % the brand.variant changes.

I am assuming that these new files are not included in Client target etc etc right?

buracules:
It should not be included. I will double check now.

Issue #3367, opened 2022-04-13
Cookie is not set after user login


chung-lai:
With the change not to navigate to Neeva home (#3309) after login, the cookie does not get set properly. Our current logic is to set login cookie only after user navigates to a Neeva page for certain authentication flow (login with Google)

Reprod:

1. Reinstall app
2. Navigate to an url (e.g., yahoo.com)
3. Go to settings and login with Google
4. Go back to settings

Note the login is not set

cc @evancrow @darinwf

chung-lai:
Given release is happening soon, planning to go back to previous UX and navigate to neeva.com upon login. And we can revisit this for next release.

darinwf:
Should we close this bug and open a new bug about making it so that sign-in does not create a neeva home navigation? Or should we just re-purpose this bug?

chung-lai:
We can file a new one to skip neeva home navigation by looking into how to set token for okta flow.

chung-lai:
Filed #3614. Closing this now

PR #3375, opened 2022-04-14
Correctly Pass Auth Cookie

evancrow:
Instead of setting the token via the `WebView`, we can pass it directly to `UserInfo`.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #3367

### DIFF
--- a/Shared/API/NeevaUserInfo.swift
+++ b/Shared/API/NeevaUserInfo.swift
@@ -115,6 +115,10 @@ public class NeevaUserInfo: ObservableObject {
         }
     }
 
+    public func updateLoginCookie(with token: String) {
+        self.setLoginCookie(token)
+    }
+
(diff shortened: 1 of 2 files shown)


Shared/API/NeevaUserInfo.swift
```diff
@@ -115,6 +115,10 @@ public class NeevaUserInfo: ObservableObject {
         }
     }
 
+    public func updateLoginCookie(with token: String) {
```
chung-lai, line 118: Why do we need this function?

chung-lai (commented):
Think this may fix part of the issue but this probably won't work with `oktaSignin` which we don't get the token right away but until we navigate to a neeva url.

evancrow:
Closing this pending further investigation and discussion

PR #3376, opened 2022-04-14
Make Sure OverlayView is Allowed to be Visible

evancrow:
Wrap the content in an if statement that checks if `canDisplay` is `true`

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


chung-lai (approved):
Looks like this only affect setting screen and it does not cause any functionality issue. The change LGTM but would feel better to soak this change a bit on `main` instead of merging to release.

Issue #3324, opened 2022-04-07
Create Zombie Tabs When Low-Memory Warning Received


evancrow:
If we receive a low memory warning, we should take the tabs (except the 10 more recently used or whatever number we decide on) and make them into zombies.

A Zombie Tab is a tab whose `WebView`, which is very memory expensive, is `nil`.

cc @darinwf

PR #3378, opened 2022-04-14
Performance Zombie Tabs

evancrow:
(behind a flag)

When enabled, if a low memory warning is received we will only allow ten tabs (or whatever we set it to) to have a live `WebView`.  All others will have their `WebView` closed.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #3324


Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -679,8 +679,8 @@ private class TabContentScriptManager: NSObject, WKScriptMessageHandler {
     }
 
     func addContentScript(_ helper: TabContentScript, name: String, forTab tab: Tab) {
-        if helpers[name] != nil {
-            assertionFailure("Duplicate helper added: \(name)")
+        guard helpers[name] == nil else {
+            return
```
chung-lai, line 683: Should we still assert with the code change? Or it's on purpose?

Client/Application/AppDelegate.swift
```diff
@@ -317,6 +317,25 @@ extension AppDelegate {
...
         ClientLogger.shared.logCounter(.LowMemoryWarning, attributes: attributes)
+
+        // Turn all but the newest x Tabs into Zombie Tabs.
+        if FeatureFlag[.lowMemoryZombieTabs] {
+            let maximumNumberOfAliveTabs = 10
+
+            for sceneDelegate in SceneDelegate.getAllSceneDelegates() {
+                // Filter tabs for each Scene
+                let tabManager = sceneDelegate.bvc.tabManager
+                let tabs = tabManager.tabs.sorted {
```
chung-lai, line 328: Pretty sure it does, does `tabManager.tabs` include all the tabs (even the ones in tab group) and incognito tabs as well?

Client/Application/AppDelegate.swift
```diff
@@ -317,6 +317,25 @@ extension AppDelegate {
...
+                    $0.lastExecutedTime ?? Timestamp() > $1.lastExecutedTime ?? Timestamp()
+                }
+
+                tabs.enumerated().forEach { index, tab in
+                    if index >= maximumNumberOfAliveTabs {
+                        tab.close()
```
chung-lai, line 334: Was looking at the usage of `close()` in the code and currently it's called in `RemoveTab` only. In one of the cases, it posts a `didClose` event. Not an expert on tabManager code, just worried about any side effects by calling `close` without removing the actual tab.

Also, when user selects the zombie tab, does it undo what we do in `close()`.

chung-lai (commented):
It's a good change, just left a few questions.

Also, 10 seems like a good place to start and we can tune it later. Wonder how much memory each tab consumes. May be we can add attributes to the current memory warning logging and see what # of tabs cause most memory warnings :)

Client/Frontend/Browser/Tabs/Tab.swift
evancrow: This is on purpose... sometimes when restoring the zombie tab it would crash but really it's just better to use a guard statement.

Client/Application/AppDelegate.swift
evancrow: Yes, tabs is all tabs

Client/Application/AppDelegate.swift
evancrow: Yeah. All the code in `close` is related to deleting the `WebView`, that's it. All of that is reset in `createWebView` which is called when a user selects a tab.

Client/Frontend/Browser/Tabs/Tab.swift
chung-lai: interesting, wonder if we should try to understand more why it fails sometimes like if something didn't clean up properly. May be change to a log statement?

chung-lai (approved):
Thanks for doing this!

PR #3380, opened 2022-04-15
Fix Suggestion Back Button Edge Case

evancrow:
Found an edge case where sometimes when using the suggestion back button, tapping a new result would do nothing as it would be canceled out by a `WebView` back navigation. This prevents the back navigation if the user tapped a result.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


chassu (approved):
LGTM

PR #3385, opened 2022-04-15
Fix AllowedView OverlayView Logic

evancrow:
This fixes an issue where some views would not be shown when closing the `SettingsView` (i.e. the `IntroView`). This was caused by the `OverlayView` in the `SettingsView` dismissing the `IntroView` while it was still visible. I decided to remove the logic for `isVisible` and close an overlay, instead of just hiding the view until it dismisses itself, or is allowed to be shown.

### How was this tested?
- [ ] I tested this manually
- [x] I added automated tests


darinwf:
Is there a github issue that this resolves?

darinwf (approved):

evancrow:
> Is there a github issue that this resolves?

Nope!

Issue #3343, opened 2022-04-09
iPad Card Grid Messed Up After Resizing Window/WebView


evancrow:
When using iPad in split view, and general usage (zooming in a tab, changing the window frame, opening/closing tabs) I sometimes get into a state where the Cards overflow and I can't interact with anything until I resize the window.

(Please ignore my essay haha)
https://user-images.githubusercontent.com/37356984/162590532-9a3f0e5f-4e27-45aa-8db1-15796f3baf7a.MOV

FYI @chassu, could be related to the bug where the tab column count is messed up.

Also, by window, I mean the split screen view the Neeva app is in (UIWindow).

evancrow:
I also found that opening an external URL can cause this bug too.

Issue #2804, opened 2022-02-05
CardGrid Sometimes Only Shows Two Columns When It Usually Shows Four


evancrow:
On my iPad, sometimes I can get into a state where only two columns are shown, when usually there are four. Restarting the app usually fixes the bug. I noticed that this sometimes occurs after signing out and back in but that doesn't always recreate the bug. FYI @chassu.

![image](https://user-images.githubusercontent.com/37356984/152653529-5f226c64-d36f-43e9-983d-ba712ec8ab14.png)

PR #3387, opened 2022-04-17
Fix CardGrid Resize Issues

evancrow:
Caused by a bug where the `GeomReader` would update with the wrong value while the app was in the background. This limits the updates to when the `BVC` is active.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #2804
Fixed: #3343


Client/Frontend/Browser/Card/CardGrid.swift
```diff
@@ -125,15 +125,22 @@ struct CardGrid: View {
     }
 
     func updateCardSize(width: CGFloat, topToolbar: Bool) {
+        guard SceneDelegate.getBVC(with: tabCardModel.manager.scene).isActive else {
```
darinwf, line 128: nit: Let's avoid adding more dependencies on BVC. Avoiding circular dependencies between modules helps keep code easier to understand and ideally unit test. How about changing `BVC.isActive` to be something like `GridModel.shouldUpdateCardSize`?

darinwf (approved):

PR #3392, opened 2022-04-18
Just call a method on TabCardModel directly

darinwf:
No need to go through a Combine publisher for what could just be a method.


chassu (approved):
LGTM

PR #3393, opened 2022-04-18
Eliminate the asyncAfter in CardsContainer

darinwf:
Use a new set of state variables in `GridModel` to signal when the grid has been vertically or horizontally scrolled to reveal the selected `Card`. Use this as the signal to drive the `scrollCompletion` that initiates the card transition animation.

This still involves some potentially precarious `DispatchQueue.main.async` calls, but those feel more acceptable than the `DispatchQueue.main.asyncAfter` that we have today.

### How was this tested?
- [x] I tested this manually using an iPhone 6s and simulator. Both seem to work well with this change. I cannot repro any "wax on" effect or any other quirks. Tested with a variety of cases, including focusing a selected tab that is far out of view as well as one that is within a collapsed card grid that requires both vertical and horizontal scrolling to reveal.


darinwf:
@chassu - Not intending just yet to land this. Would be great if you could test it out when you have a chance to see if it feels just as solid as the current solution. Thanks!

darinwf:
@evancrow - I wonder if this PR will help resolve the issues you are seeing with not being able to interact with cards in the card grid sometimes.

chassu:
re 1: oh, whichever comes first has nil'ed out the completion... I think

darinwf:
> Tested it out and it looks good. I have two questions though:
> 
> 1. when we perform a horizontal scroll and a vertical scroll, doesn't that queue up two completion calls? Are we fine with that?

Right. What you wrote. It keeps the logic simple to allow possibly two useEffect callbacks. Whichever one runs first will do the work of running the completion callback. It seems like it would be more ideal to wait for the second, when we know we should be getting a second call, before we run the completion callback. I assume it is not creating an issue since the completion callback is run through a `DispatchQueue.main.async`, which presumably allows time for the horizontal scroll to complete.


> 2. I noticed that the vertical scroll always come before the horizontal scroll. Do you have an explanation of why it always comes in that order?

Two thoughts:

a)  When vertical scrolling is required, the `CollapsedCardGroupView` may not yet be created. It is only when the `CollapsedCardGroupView` is created that it can respond to the `useEffect` for `needsScrollToSelectedTab`.

a)  When the `CollapsedCardGroupView` does exist, maybe SwiftUI has anyways decided to run the vertical scroll first since `CardsContainer` probably registered its `useEffect` for `needsScrollToSelectedTab` before `CollapsedCardGroupView` would have done so. This is because `CardsContainer` is the parent `View`.

chassu (approved):

Issue #3388, opened 2022-04-17
ios-log: scheme URLs should be ignored


jonathangarelick:
Reported by @EdwardJTL 

<img src="https://user-images.githubusercontent.com/41495718/163726791-cab266f3-2a13-45be-b705-fa13318a559d.png" height="500">

To reproduce:
-Go to movieguide.org
-Click around on some internal links, and this eventually should trigger

This functionality is used by Movie Guide's iOS app to record JavaScript console logs, but I don't think this is relevant to the user.
Related StackOverflow explanation: https://stackoverflow.com/a/6508343

darinwf:
I see. So is the idea that we should maintain a list of URL schemes that we should just silently drop on the floor?

evancrow:
So now it appears that we have the opposite issue as before (needing to maintain a list of URLs to not allow instead of allow). Curious if there's some sort of middle ground, such as a rule list that the scheme must follow.

jonathangarelick:
@darinwf Yes.
@evancrow You're right, but this seems like an improvement. I think it is reasonable to assume most custom scheme queries are okay, there are just going to be some weird situations that need to be handled, like this.

darinwf:
I imagine the set of URL schemes we'd want to block would be relatively low compared to the number we'd want to allow, and it makes sense to by default allow all. So I think our strategy of selectively blocking those that are problematic makes sense.

By the way, how do other browsers handle this case? (Always good to consider what users experience in other browsers when deciding on a policy for ours.)

jonathangarelick:
Safari and Chrome iOS ignore it. Firefox iOS has the same problem that we do.

darinwf:
Interesting... I don't see any hits for "ios-log" in the Chromium codebase: https://source.chromium.org/search?q=ios-log&sq=&ss=chromium

I wonder how / why it ignores this.

jonathangarelick:
Did some digging and it looks like here is where Chrome handles external urls. Strangely, `IsAppSpecificURL` seems to always return false, so I must be missing something

https://source.chromium.org/chromium/chromium/src/+/main:ios/chrome/browser/app_launcher/app_launcher_tab_helper.mm;l=100?q=RequestToLaunchApp&ss=chromium%2Fchromium%2Fsrc:ios%2F

darinwf:
[...]

PR #3394, opened 2022-04-18
Exclude Schemes from OpenInAppOverlay

evancrow:
Creates a list of schemes to ignore in WebViewDelegate

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Issue: #3388


Client/Frontend/Browser/URLFixup.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 import Shared
 
-class URIFixup {
```
EdwardJTL, line 8: i wonder if there is a reason this was named `URI` and not `URL` in the original repo. It seems like `getURL` for example may add a scheme to the input. So maybe we should leave the name as is because the inputs to these utils may not be URLs? @darinwf may know more?

darinwf: The mozilla codebase is/was pedantic about the distinction between URI and URL, when in practice it doesn't matter :-) I recommend standardizing on URL for everything since that avoids confusion and eliminates this topic :-D

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -729,7 +729,10 @@ extension BrowserViewController: WKNavigationDelegate {
             return
         }
 
-        if !(url.scheme?.contains("neeva") ?? true) {
+        // Only show OpenInAppOverlay for schemes other than Neeva and from the main frame.
+        if !(url.scheme?.contains("neeva") ?? true)
+            && (navigationAction.targetFrame?.isMainFrame ?? false)
```
EdwardJTL, line 734: nit: can we separate them with a comma since they are already on two lines?
```
if !(url.scheme...),
   navigationAction...
```
also this way we can have `// Disallow custom app scheme requests via iframes` comment directly on that second line

EdwardJTL (approved):

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
evancrow: Think I got it... Swift Format kinda freaked out a bit.

PR #3396, opened 2022-04-19
Add Debug Code for ZombieTabs

evancrow:


Client/Application/AppDelegate.swift
```diff
@@ -316,31 +316,30 @@ extension AppDelegate {
...
-                let tabs = tabManager.tabs.sorted {
-                    $0.lastExecutedTime ?? Timestamp() > $1.lastExecutedTime ?? Timestamp()
-                }
-
-                tabs.enumerated().forEach { index, tab in
-                    if index >= maximumNumberOfAliveTabs {
-                        tab.close()
-                    }
-                }
+                sceneDelegate.bvc.tabManager.makeTabsIntoZombies()
```
darinwf, line 342: You know what... maybe we should be more aggressive here. Maybe in low memory situations we should just kill everything but the selected tab.

darinwf (approved):

Client/Application/AppDelegate.swift
evancrow: I think we should at least keep 1-3 tabs around. In case the user is trying to multitask.

darinwf: Yeah, that's a good point. I was thinking about the scenario of a low memory warning coming in when you have 10 or fewer tabs already. Maybe in that case, we should be taking more drastic measures. One idea would be to use a ratio instead of a hard limit. So instead of saying keep 10 tabs, we could say keep half of the tabs (always including the selected tab). Just thinking out loud here.... this PR is good to go.

evancrow: Let's merge this for now, and hopefully the logging will give us some info on how many tabs is too much.

PR #3379, opened 2022-04-15
time based switcher pt 2

chassu:
add archived tab settings view and fix for cookie cutter settings

### How was this tested?
- [x] Existing automated tests are enough to cover my changes

### Screenshots and screen recordings

Uploading Simulator Screen Recording - iPhone 13 Pro - 2022-04-14 at 18.14.52.mp4…

PR #3397, opened 2022-04-19
create a state where no tab is selected in time based tab switcher

chassu:
Add a state where none of the tab is selected. This happens when the last tab in the today section is closed.

### How was this tested?
- [x] I tested this manually

### Manual test cases

1. close last tab in today section using close button
2. close last tab in today section using card detail's long press menu
3. close last tab in today section using switcher tool bar's long press menu
4. close a tab in last week's section when no tab is selected

https://user-images.githubusercontent.com/87332917/163908555-f109987d-4840-41ea-8d9f-762e8456a1f6.mp4


Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -396,6 +380,15 @@ class TabManager: NSObject {
         }
     }
 
+    func shouldSelectNewTabAfterRemovalOf(tab: Tab) -> Bool {
+        let tabsFiltered = tabs.filter {
+            return tab.isIncognito == $0.isIncognito && filterTabByTime(tab: $0, byTime: .today)
+        }
+        return
+            !(tabsFiltered.count == 1 && tabsFiltered[0].tabUUID == tab.tabUUID
```
evancrow, line 388: I'm concerned that if `tabsFiltered` is empty then this will crash with an `index not in range` error. Can you try using `.first` and unwrapping that?

(this also applies to other places as well)

chassu: but if the first condition evaluates to false, the rest of the conditions won't be evaluated. Doesn't that avoid the index problem?

evancrow: If you unwrap` .first` it will be a cleaner, "swiftier" solution. Relying on a specific count can sometimes lead to bugs later on.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -190,4 +196,16 @@ extension BrowserViewController: ToolbarDelegate {
 
         return UIMenu(children: actions)
     }
+
+    func isLastSelectedTabTodaySection(tab: Tab) -> Bool {
```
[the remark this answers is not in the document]
darinwf: nit: suggest renaming to `isLastSelectedTabInTodaySection` <- note the "In" addition ... actually, we shouldn't encode "Today" here since the first section may not contain just tabs from today, right? This the recent tabs section. Should choose a name to reflect that reality.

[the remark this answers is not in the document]
darinwf: Why isn't this method just the same as `shouldSelectNewTabAfterRemovalOf`? Why is the logic different...

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -463,3 +459,51 @@ class TabManager: NSObject {
...
+            : Calendar.current.date(
+                byAdding: .day, value: -1, to: Date())
+        guard let startOfOneDayAgo = minusOneDayToCurrentDate else {
+            return true
+        }
+        // timeIntervalSince1970 returns the number of seconds. It is converted
+        // to milliseconds by multiplying by 1000 to compare with lastExecutedTime
+        // which is stored in milliseconds.
+        switch byTime {
+        case .today:
```
darinwf, line 497: I'm confused about the time segmentation here. You have tabs from today and tabs from last week? What about tabs from the days in between? Oh... lastWeek seems to be inclusive of the tabs used yesterday? I would think tabs from last week should not include tabs from yesterday.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -42,22 +42,6 @@ class TabManager: NSObject {
 
     let navDelegate: TabManagerNavDelegate
 
-    public static func makeWebViewConfig(isIncognito: Bool) -> WKWebViewConfiguration {
```
darinwf, line 45: What was the purpose of moving this function?

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -12,9 +12,13 @@ extension TabManager {
         guard let index = tabs.firstIndex(where: { $0 === tab }) else { return }
         addTabsToRecentlyClosed([tab], showToast: showToast)
         removeTab(tab, flushToDisk: true, notify: true)
+        let bvc = SceneDelegate.getBVC(with: scene)
```
darinwf, line 15: Can we please avoid adding dependencies on BVC from `TabManager`? If we want to have the tab tray show when there is no selected tab, then BVC should just do so via a sink attached to selectedTabPublisher, right?

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -463,3 +459,51 @@ class TabManager: NSObject {
...
+        configuration.ignoresViewportScaleLimits = true
+        if isIncognito {
+            configuration.websiteDataStore = WKWebsiteDataStore.nonPersistent()
+        }
+        configuration.setURLSchemeHandler(InternalSchemeHandler(), forURLScheme: InternalURL.scheme)
+
+        return configuration
+    }
+
+    func filterTabByTime(tab: Tab, byTime: TimeFilter) -> Bool {
```
darinwf, line 480: Maybe this should be a method on `Tab` instead? It doesn't seem to use any state from `TabManager`. The name could also be improved, something like: `tab.wasLastExecuted(.today)` which reads nicely and aligns to the `lastExecutedTime` property of `Tab`.

The definition of `TimeFilter` should also move to `Tab.swift` accordingly.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -190,4 +196,16 @@ extension BrowserViewController: ToolbarDelegate {
...
+        let allDetailsFiltered = self.tabCardModel.allDetails.filter { tabCard in
+            return
+                (self.tabCardModel.representativeTabs.contains(tab)
+                || self.tabCardModel.allDetailsWithExclusionList.contains { $0.id == tabCard.id })
```
darinwf, line 204: I think you should avoid using IDs for equality checks. See https://github.com/neevaco/neeva-ios/pull/3403 for reference. You should just compare the `TabCardDetails` to another `TabCardDetails` directly. `TabCardDetails` conforms to `Equatable` afterall.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -190,4 +196,16 @@ extension BrowserViewController: ToolbarDelegate {
...
+                && tab.isIncognito
```
darinwf, line 205: Why are you checking incognito here?

darinwf (changes requested):

Client/Frontend/Browser/BrowserViewController/BrowserViewController+TabToolbarDelegate.swift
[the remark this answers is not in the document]
chassu: I forgot to change this part. I was going to use `shouldSelectNewTabAfterRemovalOf` instead and remove this function.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: Hm, I thought we had this discussion before. I thought the definition for the sections are:
today: {set of tabs whose last used time is < 24 hours}
last week: {set of tabs whose last used time is < 7 days - {today}}

If tabs from yesterday don't belong to last week, then where do they go? It feels weird to be to include them in "today"'s section.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: I didn't mean to modify this part but it happened when I cherry-picked the very first commit into this branch. I end up place it in where it lives on the main branch. I think this should be fine once #3379 is merged.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+TabToolbarDelegate.swift
chassu: This function is removed. But `shouldSelectNewTabAfterRemovalOf` uses incognito as a filter so I'll explain - it's because we want to make sure that no tab is selected when the last tab in today's section is closed. The tabs array contains all tabs including both incognito and normal tabs. So I need the filter to make sure we don't include incognito tabs.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
darinwf: I thought we made a decision to call this section "Recent" instead of "Today". Please confirm with Ria, but see also: https://www.figma.com/file/t1jW7cahkClVHAIlnHJuv8/Tab-Manager-(Spaces-%26-Tab-Groups)?node-id=3264%3A12992

chassu: I looked up the meeting notes before and can confirm we decided to start with recent. Do you think it's better to include yesterday's tabs to recent section?

chassu:
can confirm the issue above has been resolved (also made a minor change to check if any card is selected instead of isEmpty)

Client/Frontend/Browser/BrowserViewController/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -164,7 +164,14 @@ extension BrowserViewController: ToolbarDelegate {
                 title: Strings.CloseTabTitle, image: UIImage(systemSymbol: .xmark)
             ) { _ in
                 if let tab = self.tabManager.selectedTab {
-                    self.tabManager.removeTab(tab, showToast: false)
+                    if FeatureFlag[.enableTimeBasedSwitcher] {
+                        self.tabManager.removeTab(
+                            tab, showToast: false,
+                            updateSelectedTab: self.tabManager.shouldSelectNewTabAfterRemovalOf(
```
darinwf, line 170: nit: It seems a bit awkward to have `tabManager.removeTab`'s `updateSelectedTab` parameter determined by another method of `TabManager`. Why doesn't `TabManager` just call its own method directly? I guess the answer is that you don't want to run this logic in all cases. Perhaps that difference should be expressed by inventing a new variant of `removeTab` that is named differently, appropriately setting it apart from normal `removeTab`?

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -40,6 +40,11 @@ protocol TabDelegate {
     @objc optional func tab(_ tab: Tab, willDeleteWebView webView: WKWebView)
 }
 
+enum TimeFilter: String {
+    case today = "Today"
```
darinwf, line 44: It's a separate matter from this PR, but we should revisit the definition of these categories. Our History UI has Today, Yesterday and Last Week. Perhaps we should standardize on those as well.

I don't think it makes sense to define Today as within the last 24 hours as that includes yesterday, but we can discuss separately.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -15,6 +15,8 @@ extension TabManager {
 
         if updateSelectedTab {
             updateSelectedTabAfterRemovalOf(tab, deletedIndex: index, notify: true)
+        } else {
+            self.selectTab(nil, notify: true)
```
darinwf, line 19: It doesn't make sense that in the `!updateSelectedTab` case we are calling `selectTab`. That's updating the selected tab. Perhaps the parameter is now misnamed?

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -43,6 +43,7 @@ class TabManager: NSObject {
     // Use `selectedTabPublisher` to observe changes to `selectedTab`.
     private(set) var selectedTab: Tab?
     private(set) var selectedTabPublisher = CurrentValueSubject<Tab?, Never>(nil)
+    private(set) var noSelectedTabStatePublisher = PassthroughSubject<Void, Never>()
```
darinwf, line 46: This is stale.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -399,6 +398,18 @@ class TabManager: NSObject {
         }
     }
 
+    func shouldSelectNewTabAfterRemovalOf(tab: Tab) -> Bool {
+        let tabsFiltered = tabs.filter {
```
darinwf, line 402: nit: I think you can optimize this function a bit to just count the number of tabs executed today. There's no need to generate an array of all tabs from today to implement this function.

That said, another approach here could be to revise updateSelectedTabAfterRemovalOf to handle this logic. After the tab has been removed, you can just scan the array to see if there is any remaining tab in the today section. Wouldn't that be simpler?

chassu: I removed `shouldSelectNewTabAfterRemovalOf` and handle the logic inside `updateSelectedTabAfterRemovalOf`. Let me know what you think.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -85,7 +85,10 @@ extension TabManager {
         let viableTabs: [Tab] = tab.isIncognito ? incognitoTabs : normalTabs
         let bvc = SceneDelegate.getBVC(with: scene)
 
-        if closedLastNormalTab || closedLastIncognitoTab {
+        if closedLastNormalTab || closedLastIncognitoTab
+            || (FeatureFlag[.enableTimeBasedSwitcher]
+                ? !viableTabs.contains(where: { $0.wasLastExecuted(.today) }) : false)
```
darinwf, line 90: So much simpler, right?!

darinwf (approved):
LGTM

PR #3403, opened 2022-04-19
Fix bug where restored tabs do not load and cannot be closed

### LINKED ISSUE #3401: Regression: Tab restore does not work; just results in a blank tab
@evancrow FYI

There seems to be no way to get the tab to reopen properly without force quitting and restarting the app. Also notice that you cannot even close one of these tabs, and when trying to edit the URL / query, it won't actually leave the suggest experience when hitting enter.

https://user-images.githubusercontent.com/79867437/164042986-1f2a97ff-a809-49e2-8708-ccf3a009b885.mp4

darinwf:
Fixes #3401

### How was this tested?
- [x] I tested this manually
- [x] I added automated tests


darinwf:
@evancrow @chassu FYI

yozuysal (approved):

PR #3406, opened 2022-04-19
hide grid after selecting a tab from history view

chassu:
### How was this tested?
- [x] I tested this manually

before:

https://user-images.githubusercontent.com/87332917/164108913-575826df-33e4-4e38-abdb-54605aa4e1ba.mp4

after:

https://user-images.githubusercontent.com/87332917/164108937-c8f5773a-ad86-4298-ac98-2270def9572d.mp4


darinwf:
Hmm... tapping on an item in History should use createOrSwitchToTab and **bring the selected tab to the foreground**. It is perhaps a regression from the new SwiftUI-based History UI that that isn't happening now.

darinwf:
Something like this:
```
diff --git a/Client/Frontend/Browser/BrowserViewController.swift b/Client/Frontend/Browser/BrowserViewController.swift
index d9fd2f0e9..6a4da31ed 100644
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1494,6 +1494,10 @@ extension BrowserViewController {
         }
 
         updateInZeroQuery(selected?.url as URL?)
+
+        if selected != nil {
+            hideCardGrid(withAnimation: false)
+        }
     }
 
     func getSceneDelegate() -> SceneDelegate? {
```

darinwf (changes requested):
See comment above on suggested change.

chassu:
https://user-images.githubusercontent.com/87332917/165200003-8227051f-2c76-413c-982f-8e487d66a3b1.mp4

chassu:
looks the that fix alone is causing the normal card transition animation to fail. will investigate it.

Client/Frontend/Site/SiteRowView.swift
```diff
@@ -32,6 +33,7 @@ struct SiteRowView: View {
             Button {
                 action()
                 openURL(site.url)
+                browserModel.hideGridWithNoAnimation()
```
darinwf, line 36: Can this be moved within the `openURL` handler? I feel like it is a lot for `SiteRowView` to have to know about hiding the grid.

darinwf: Yeah, `HistoryPanelViewController` seems like a better spot for this. WDYT?

chassu: makes sense to me!

darinwf (approved):

darinwf:
When you commit be sure to update the commit log entry accordingly since this PR evolved a lot since the start :)

PR #3407, opened 2022-04-20
Adjust Copy For EmptyCardGrid When TopToolbar

evancrow:
Changes the copy for EmptyCardGrid to say above instead of below when the top toolbar is active. Also, probably need a similar UI for spaces.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Client/Frontend/Browser/Card/EmptyCardGrid.swift
```diff
@@ -6,13 +6,14 @@ import SwiftUI
 
 struct EmptyCardGrid: View {
     let isIncognito: Bool
+    var isTopBar: Bool = false
```
EdwardJTL, line 9: can this be a let?

EdwardJTL (commented):
could you add some screenshots for inline vs not inline after the change?

Client/Frontend/Browser/Card/EmptyCardGrid.swift
evancrow: Well that depends... do we want false to be the default value or do we not want one?

evancrow:
Here's the screenshot (the incognito version has the top toolbar copy):

Screen Shot 2022-04-19 at 9 11 24 PM

EdwardJTL (approved):

PR #3411, opened 2022-04-20
Add image support for xyz suggestions

buracules:
- Add thumbnail support for xyz suggestions

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases

Simulator Screen Shot - iPhone 13 Pro - 2022-04-21 at 02 17 39


Client/Frontend/Suggestions/SearchSuggestionViews.swift
```diff
@@ -713,25 +713,37 @@ private struct FindInPageSuggestionView: View {
...
-        Text(query)
+    private var label: some View {
+        Text(suggestion.displayText)
             .withFont(.bodyLarge)
             .lineLimit(1)
     }
 
+    @ViewBuilder
+    private var icon: some View {
+        if let urlString = suggestion.image, let imageUrl = URL(string: urlString) {
```
yozuysal, line 729: this will probably give you a warning since you are not using urlString.

I would directly do let imageUrl = URL(string: suggestion.image ?? "")

Client/Frontend/Suggestions/SuggestionModel.swift
```diff
@@ -664,7 +664,7 @@ class SuggestionModel: ObservableObject {
             bvc.updateFindInPageVisibility(visible: true, query: query)
         case .xyzQuery(let query):
```
yozuysal, line 665: nit: suggestion

yozuysal (approved):
Two nits and lgtm

Client/Frontend/Suggestions/SearchSuggestionViews.swift
buracules: I am using inside the constructor of imageURL.  But yes i dont need to do that i can just use directly like you said.

PR #3415, opened 2022-04-21
Remove Toast Redundancy 

evancrow:
Removes the need to enqueue a Toast, now it's done by default.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Client/Frontend/Toast/ToastViewManager.swift
```diff
@@ -11,24 +11,37 @@ class ToastViewManager: QueuedViewManager<ToastView> {
...
+        }
+
+        return toast
     }
 
     public func makeToast(
         content: ToastViewContent,
         toastProgressViewModel: ToastProgressViewModel? = nil,
-        displayTime: Double = 4.5, autoDismiss: Bool = true
+        displayTime: Double = 4.5, autoDismiss: Bool = true, enqueue: Bool = true
```
darinwf, line 34: nit: doesn't look like any callers set `enqueue` to false, so why have it as a parameter?

darinwf (approved):

Client/Frontend/Toast/ToastViewManager.swift
evancrow: That's true, just thought it might be a nice future thing, just in case?

darinwf: I recommend we leave it out until we actually need it. Keeps the code simpler :)

darinwf (approved):

PR #3416, opened 2022-04-21
TabStore and import flow fixes

yozuysal:
- Fixes an issue with tab restore for xyz target by adding a new group
- Updates the flow for public address only so that instead of a toast, we
show a branded UI with clarification about why the user needs to import credentials
- Small fix to gas fee model not getting a value immediately.

https://user-images.githubusercontent.com/83613310/164510542-d9f3975d-bd37-4ea2-b4b8-9ae80f331f73.mp4


WalletCore/GasFeeModel.swift
```diff
@@ -46,19 +46,24 @@ public class GasFeeModel: ObservableObject {
     }
 
     public func configureTimer(with wallet: WalletAccessor?, chain: EthNode = EthNode.Ethereum) {
+        guard let wallet = wallet else { return }
+        updateGasPrice(with: wallet, chain: chain)
```
buracules, line 50: Sorry I missed that. It wasn't clear while interval is 2.

yozuysal: no worries. I also didn't notice it until now because of it updating frequently.

buracules (approved):

PR #3417, opened 2022-04-21
Add loading indicator support for spaces without follow

buracules:
- Moved space details function to SpaceCardModel
- Added loading indicator 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/15333612/164538663-a62cc9e5-5dd6-463f-885c-b60235599966.mp4


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -776,6 +776,20 @@ class SpaceCardModel: CardModel {
                 Defaults[.seenBlackFridayNotifyPromo] = true
             })
     }
+    
+    func updateSpaceDetails(with id: String, manager: SpaceStore) {
+        manager.setState(.refreshing)
```
yozuysal, line 781: I feel like this should be the job of the SpaceStore since it is internal state. Anyway we can set the right state in the openSpaceWithNoFollow implementation?

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -511,8 +512,13 @@ class SpaceCardDetails: CardDetails, AccessingManagerProvider, ThumbnailModel {
     init(space: Space, manager: SpaceStore) {
         self.id = space.id.id
         self.manager = manager
-        self.spaceRef = space
-        updateDetails()
+        setSpace(space)
+    }
+
+    init(id: String, manager: SpaceStore) {
```
yozuysal, line 518: hmm, so technically this can be called for a space that is in SpaceStore that we are following as well. So maybe isFollowing should not be false by default, but should be more like manager.allSpaces.contains {$0.id == id}?

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -776,6 +776,13 @@ class SpaceCardModel: CardModel {
                 Defaults[.seenBlackFridayNotifyPromo] = true
             })
     }
+
+    func updateSpaceDetails(with id: String, manager: SpaceStore) {
```
yozuysal, line 780: I think let's name this to be the same updateSpaceWithNoFollow

yozuysal (approved):
let's make sure to test with

fresh install (not signed out after sign in, but unistall and reinstall)
signed in user

otherwise a few nits and lgtm

PR #3418, opened 2022-04-21
Handle webui message to show sign up/in flow in preview mode for site preferences

jedireza:
Fixes a bug where a preview mode "Sign Up" button click/tap does nothing (site preferences). Depends on a `webui` PR to land. That PR sends a message over the native bridge and this PR handles that message.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Tested in the simulator with my local instance of `webui` running with the fix.


Client/Frontend/Prompt/WebUIMessageHelper.swift
```diff
@@ -37,30 +54,23 @@ class WebUIMessageHelper: TabContentScript {
...
-                id: id, stepName: tourStep, webView: self.webView! as WKWebView)
-            bvc.showQuestNeevaMenuPrompt()
-        case .openFeedbackPanelWithInputFieldHighlight:
-            TourManager.shared.setActiveStep(
-                id: id, stepName: tourStep, webView: self.webView! as WKWebView)
-            showFeedbackPanel(bvc: bvc)
+        switch message.name {
+        case .showSignUp:
+            ClientLogger.shared.logCounter(
+                .PreviewPromptSignIn, attributes: EnvironmentHelper.shared.getFirstRunAttributes())
```
jedireza, line 71: @darinwf I chose the `.PreviewPromptSignIn` interaction. Let me know if you think it should be something else.

Client/Frontend/Prompt/WebUIMessageHelper.swift
```diff
@@ -5,6 +5,23 @@ import WebKit
 
 let messageHandlerName = "webui"
 
+private enum WebUIMessageName: String {
+    case showSignUp = "showSignUp"
```
darinwf, line 9: `String`-valued enum cases already have an implicit raw value matching the name of the case, so no need for the `= "showSignUp"` part. You can just drop that and it should all still work.

Client/Frontend/Prompt/WebUIMessageHelper.swift
darinwf: Actually, we should define a new interaction for this. It may be a good idea also to make `showSignUp` more specific to preferred providers configuration or to have another parameter that indicates the origin of the button in the UI of the web page. That way the logging will more precisely indicate what UI element is driving sign ups, etc.

Client/Frontend/Prompt/WebUIMessageHelper.swift
```diff
@@ -5,6 +5,23 @@ import WebKit
...
+}
+
+private struct WebUIMessage {
+    let id: String
+    let name: WebUIMessageName
+    let data: [String: Any]
+
+    init(id: String, name: String, data: [String: Any]) {
+        self.id = id
+        self.name = WebUIMessageName(rawValue: name) ?? WebUIMessageName.unknown
```
darinwf, line 20: By the way, might also be possible to write this as:

```
self.name = .init(rawValue: name) ?? .unknown
```

Swift!

Trimming off `WebUIMessageName` before `.unknown` is the one that is super commonplace. The `.init` is less so but I see it sometimes too.

Client/Frontend/Prompt/WebUIMessageHelper.swift
```diff
@@ -37,30 +54,31 @@ class WebUIMessageHelper: TabContentScript {
...
+                case "previewPreferredProvider":
+                    ClientLogger.shared.logCounter(
+                        .PreviewPreferredProviderSignIn,
+                        attributes: EnvironmentHelper.shared.getFirstRunAttributes())
+                default:
+                    break
+                }
+            }
+
+            bvc.presentIntroViewController(true, completion: { bvc.hideZeroQuery() })
```
darinwf, line 81: I'm not sure if `hideZeroQuery` is needed here. I suspect it shouldn't be since the zero query UI shouldn't be showing when the user clicks on this part of the page.

darinwf (approved):

Client/Frontend/Prompt/WebUIMessageHelper.swift
jedireza: Cool! Updated.

Client/Frontend/Prompt/WebUIMessageHelper.swift
jedireza: Roger that. Removed.

PR #3419, opened 2022-04-21
don't change selected tab when switching modes

chassu:
### How was this tested?
- [x] I tested this manually
- [x] I added automated tests

https://user-images.githubusercontent.com/87332917/164555052-a8a03dac-0431-46ac-a78e-044e5fc3bd3b.mp4


evancrow:
@darinwf @chassu Was just thinking about this, but with this PR we could probably change the logic of enabling the done button in `CardGrid` to check if there is a selected tab rather than if the `CardGrid` `isEmpty`. Thoughts?

darinwf (commented):
This still needs the `animation(nil)` changes we discussed, right?

chassu:
I also noticed this behavior: 

https://user-images.githubusercontent.com/87332917/165190199-00ae716d-9772-4c3b-9c29-8c9b2f0fc4d1.mp4

I'm going to fix it assuming that we prevent removeTab from selecting a new tab if the removed tab is in different mode from the selected tab. But let me know if you have different opinion.

and this:

https://user-images.githubusercontent.com/87332917/165190516-1de0a7c5-ab1f-411c-ab56-57dc2802141d.mp4

@darinwf do you like how the space transition to the selected tab without animation? Or do you think that transition should also have the same animation (spaces -> normal tabs -> card transition animation)?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -121,7 +121,8 @@ extension BrowserViewController: ToolbarDelegate {
         }
 
         let switchPrivacyMode = { [self] (_: UIAction) in
-            tabManager.toggleIncognitoMode(fromTabTray: false)
+            tabManager.toggleIncognitoMode(
+                fromTabTray: false, clearSelectedTab: false, selectNewTab: true)
```
chassu, line 125: explanation for adding `clearSelectedTab: false` - I found a bug without settings this argument. Imagine a case where a normal tab is selected while there is no incognito tab, we click the "open incognito mode" from the switcherToolBarMenu, and then click "cancel". Without settings this argument, the selectedTab (which is the original normal tab we are at) is going to be changed to nil. And once we hit cancel, the switcher button won't work anymore because the selectedTab is nil. (if selectedTab is nil, scrollToSelectedTab won't work, and all the subsequent animation is not executed)

Client/Frontend/Browser/Card/GridModel.swift
```diff
@@ -31,6 +31,7 @@ class GridModel: ObservableObject {
         }
     }
     @Published var gridCanAnimate = false
+    var switchModeWithoutAnimation = false
```
darinwf, line 34: Doesn't this need to be an `@Published` var? Otherwise, the `.animation` properties that depend on it won't necessarily use the latest value of this var. Am I missing something?

Client/Frontend/IncognitoModel.swift
```diff
@@ -22,4 +22,6 @@ class IncognitoModel: ObservableObject {
     func toggle() {
         self.isIncognito.toggle()
     }
+
+    var completion: (() -> Void)?
```
darinwf, line 26: Unused. Leftover from previous approach right?

darinwf (approved):
LG, just a couple nits

Client/Frontend/Browser/Card/GridModel.swift
chassu: I was thinking the same thing, but it was working when I didn't add `@Published`. I think it makes sense to have it.

Issue #3420, opened 2022-04-22
Text size button sometimes does not trigger second overlay


jonathangarelick:
in the app store build:

https://user-images.githubusercontent.com/41495718/164758172-2e25b831-f29a-4d9b-bfb9-ab4ef3d29f15.MP4

in the simulator on main a2139914469

https://user-images.githubusercontent.com/41495718/164758223-7839ab91-2590-4701-9806-5e40e20f0cfa.mp4

cc @evancrow

PR #3425, opened 2022-04-22
Showing/Hiding OverlayView Fixes

evancrow:
Fixed an issue where the OverlaySheets wouldn't animate to hide.

Also fixed an issue where a different overlay wouldn't show from another overlay. This was caused by a lot of extra calls to `hideOverlay()`. (Also, existing tests should work for this?)

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #3420


Client/Frontend/Overlay/Sheet/OverlaySheetView.swift
```diff
@@ -262,7 +262,12 @@ struct OverlaySheetView<Content: View, HeaderContent: View>: View, KeyboardReada
         GeometryReader { outerGeometry in
             DismissBackgroundView(
                 opacity: model.backdropOpacity, position: model.position,
-                onDismiss: style.nonDismissible ? {} : onDismiss)
+                onDismiss: style.nonDismissible
+                    ? {}
+                    : {
+                        onDismiss()
+                        model.overlayManager.hideCurrentOverlay()
```
darinwf, line 269: I wonder if `hideCurrentOverlay` is always risky? Maybe we should have a way to refer to the specific overlay instance we want to hide?

darinwf (approved):

Client/Frontend/Overlay/Sheet/OverlaySheetView.swift
evancrow: Yes? The problem is that the only instance of the overlay is `currentOverlay` which would always match what is visible, and hide that.

darinwf: Right, and I assume the issue that we are trying to fix is that the next overlay has become `currentOverlay` but the code is still trying to hide the previous overlay, and the only tool it has to do so is calling `hideCurrentOverlay`. The more robust solution then would be to let code refer to a specific overlay and for their to be no concept of "current overlay" that code can ask to hide.

darinwf (approved):
Nicely done with the latest commit!

Issue #3409, opened 2022-04-20
Text Size for Tab not serialized as part of a SavedTab


darinwf:
This means if you background the app and it gets killed, when you come back to it, the tab will switch back to the default text size. It will appear to the user as if just switching apps results in the text size getting reset.

@jonathangarelick FYI

jonathangarelick:
It appears that there are two cases to be handled here: (1) when a tab is closed, then restored and (2) when the app is killed and restores from cold start.

darinwf:
I think those are essentially the same code path: thru `SavedTab`

PR #3427, opened 2022-04-22
Parameterize TextSizeModel with Tab instead of WKWebView

jonathangarelick:

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases

  and dark mode

## Issues addressed
* Second step (of 3) to complete #3409


Client/Frontend/Overlay/Content/TextSize/TextSizeModel.swift
```diff
@@ -8,38 +8,33 @@ import WebKit
...
 
-    var objectWillChange = ObservableObjectPublisher()
-    private var subscription: AnyCancellable?
-
-    init(webView: WKWebView) {
-        self.webView = webView
-        self.subscription =
-            webView
-            .publisher(for: \.pageZoom, options: .new)
-            .sink { [weak self] _ in self?.objectWillChange.send() }
```
jonathangarelick, line 21: This subscription appears to be unused, so I removed it.

Client/Frontend/Overlay/Content/TextSize/TextSizeView.swift
```diff
@@ -30,17 +30,3 @@ struct TextSizeView: View {
...
-        var body: some View {
-            TextSizeView(model: model, onDismiss: {})
-                .overlay(Text(model.label))
-        }
-    }
-
-    static var previews: some View {
-        Preview(model: TextSizeModel(webView: WKWebView()))
-    }
-}
```
jonathangarelick, line 46: This needed to be removed because TextSizeModel now requires a Tab as a parameter, which would then need a BrowserViewController reference.

darinwf (commented):
In addition this change, and could be a follow-up PR, I think it makes sense to introduce a `Tab.pageZoom` dynamic property. I think all of the `neeva_zoomAmount` manipulation should be behind that property, so `TextSizeModel` isn't directly messing with `WKWebView` either.

Client/Frontend/Overlay/Content/TextSize/TextSizeModel.swift
```diff
@@ -8,38 +8,33 @@ import WebKit
...
 
+    // If tab.webView is nil here, something is very wrong
     private var observer: AnyCancellable?
     var pageZoom: CGFloat {
         get {
-            webView.neeva_zoomAmount
+            tab.webView?.neeva_zoomAmount ?? 1.0
         }
         set {
-            objectWillChange.send()
```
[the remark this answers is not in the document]
jonathangarelick: Done, thanks.

darinwf (approved):

PR #3428, opened 2022-04-22
Fix Other Suggestion Back Button Edge Case

evancrow:
Similar to #3380, just needed the same fix when using the enter key to return from the URL bar.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
- Go to any website from URL bar query search
- Hit the back button
- Type in a new query and tap enter
- Should be on SRP (expected and correct)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -607,12 +607,12 @@ class BrowserViewController: UIViewController, ModalPresenter {
         }
     }
 
-    public func hideZeroQuery(suggestionTapped: Bool = false) {
+    public func hideZeroQuery(performedNavigation: Bool = false) {
```
darinwf, line 610: Hmm, I would imagine that pretty much all call-sites need to pass `true` except for the one that corresponds to cancelling the zero-query UI.

Maybe we should flip the sense of this boolean and call it `wasCancelled` instead? We might also consider making it non-optional to force people to think about what value it should have.

darinwf (approved):

Client/Frontend/Browser/BrowserViewController.swift
darinwf: Actually, after a quick inspection, I see we have a bunch of spots that just call `hideZeroQuery`... not sure if those should really correspond to navigating back. Maybe only if the user hits the "cancel" button?

evancrow: I think what makes the most sense is to make the parameter `wasCancelled`, will make the change

PR #3430, opened 2022-04-23
Refactor and persist text size modification

jonathangarelick:

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases

  and dark mode

## Issues addressed


Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -97,6 +97,28 @@ class Tab: NSObject, ObservableObject {
     /// For security reasons, the URL may differ from the web view’s URL.
     @Published private(set) var url: URL?
 
+    private var observer: AnyCancellable?
+    var pageZoom: CGFloat = 1.0 {
+        didSet {
+            if let webView = webView {
+                webView.setValue(pageZoom, forKey: "viewScale")
```
jonathangarelick, line 104: This is a departure from how we were doing things, but I think this is worth a shot. Firefox is currently managing zoom this way, and Brave is planning to as well.

Client/Frontend/Browser/Tabs/TabManager/TabManager+WKNavigationDelegate.swift
```diff
@@ -48,6 +48,25 @@ extension TabManager: WKNavigationDelegate {
...
+                x: originalOffset.x, y: originalOffset.y * pageZoom)
+            observer = webView.scrollView.publisher(for: \.contentOffset)
+                .sink { offset in
+                    if offset != originalOffset {
+                        webView.scrollView.contentOffset = newOffset
+                        self.observer = nil
+                    }
+                }
+        }
+
```
jonathangarelick, line 69: You'll notice that this code is duplicated from `Tab.swift`. There's probably a better way to do this -- would appreciate your thoughts.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -72,6 +72,9 @@ class TabManager: NSObject {
     var timerToTabsToast: Timer?
     var closedTabsToShowToastFor = [SavedTab]()
 
+    // This is used to manage scrolling in the WKNavigationDelegate extension
+    var observer: AnyCancellable?
+
```
jonathangarelick, line 77: This is not in `TabManager+WKNavigationDelegate.swift` because "Extensions must not contain stored properties."

Client/Frontend/Browser/Tabs/TabManager/TabManager+WKNavigationDelegate.swift
darinwf: Here's a few ideas:

1. You could create a method on `Tab` called `applyPageZoom`, and then use that in both places.
2. You could create a method on `Tab` called `applyPageZoomIfNeeded` and then use that in both places. (Maybe it is appropriate even in the `didSet` case to test if modifying `viewScale` is necessary?)
3. Alternatively, you could do something simple like this:

```
if let pageZoom = selectedTab?.pageZoom,
    webView.value(forKey: "viewScale") as? CGFloat != pageZoom
    {
        selectedTab?.pageZoom = pageZoom
    }
```

That should trigger the `didSet` handler too.

Client/Frontend/Browser/Tabs/Tab.swift
darinwf: This looks good to me. However, we might consider obscuring "viewScale" in a similar manner to how @j-f1 did it using `componentsJoinedByString`.

j-f1: Also, I’m curious what happened to the `_` here — I definitely remember it being required (i.e. `_viewScale`). (see the source: https://github.com/WebKit/WebKit/blob/dfda84308f7e222f3cbb966641f26b3ca6b72379/Source/WebKit/UIProcess/API/Cocoa/WKWebView.mm#L3627-L3650)

j-f1: My rationale at the time was that Apple allegedly runs automated scanners that look for private API usage like this. (That’s why I also included a build-time switch to instantly disable all the private API usage, in case App Review complained). But if Firefox is shipping the API without any obfuscation, then it‘s probably fine to do the same here.

And hopefully iOS 16 will add a public API for this!

darinwf: The "_" prefix was probably needed when using Obj-C because the C ABI adds those prefixes to exported symbols. Not surprising that Swift would hide that detail.

darinwf (approved):
LGTM

Issue #3431, opened 2022-04-24
Selecting an image breaks Neeva


MatejaSrejic:
Upon selecting an attachment (in my case from the Photo Library) on any website the Neeva client turns all black. 

https://user-images.githubusercontent.com/75619038/164965446-89f8e7d0-d271-4635-beb1-7f4dbeaee046.MOV

darinwf:
Thanks for the bug report. @evancrow - any thoughts on this one?

evancrow:
Not right away... this is something WebKit handles and not something we implemented. Did a quick look in the logs though, and whenever the photo picker pops up, the `GeomFrame` is set to 10x10. Likely that controller is messing with something in SwiftUI.

PR #3433, opened 2022-04-24
Fix WebKit ImagePicker

evancrow:
Don't know why, but when the `ImagePickerController` was presented by the `WebView`, it would mess up the `BrowserView` content. A workaround was to present the controller as a `pageSheet` instead, which wouldn't interfere with any content.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/37356984/164991104-f395f4a3-cdcb-410c-9f00-8bfc2f40ed96.mp4

## Issues addressed
Fixed: #3431


darinwf (approved):
Interesting! LGTM

Issue #3437, opened 2022-04-25
Restored tabs have a blank screenshot


darinwf:
When writing out the tabs state, we do a pass to prune out any screenshots from the image cache that are no longer needed. We do that by looking only at TabManager.tabs. We should include the set of SavedTab objects that make up the undo tab close stack.

@evancrow FYI

evancrow:
@darinwf Where does this filter occur in the code?

darinwf:
In `TabManagerStore.prepareSavedTabs`

PR #3447, opened 2022-04-25
Persist Screenshot UUIDs Keys for RecentlyClosedTabs

evancrow:
To keep the screenshot images just needed to make sure the UUIDs for the screenshot were not erased.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed:  #3437


Client/Frontend/Browser/Tabs/TabManagerStore.swift
```diff
@@ -69,7 +71,11 @@ class TabManagerStore {
                 screenshots.append(.init(key: uuid.uuidString, image: image))
             }
         }
-        imageStore?.updateAll(screenshots)
+
+        imageStore?.updateAll(
```
darinwf, line 75: Hmm, you had plumb through this extra param because `SavedTab` does not hold onto the screenshot for the tab. This makes sense. We could add a reference to the screenshot from `SavedTab`, but that would potentially waste memory. OK, makes sense.

nit: `exisitingKeysToKeep` has a typo. Actually how about calling this parameter `extraKeysToKeep`?

darinwf (approved):
LG!

darinwf (approved):

PR #3448, opened 2022-04-26
Prevent visual jumping when expanding tab group

jedireza:
When expanding a tab group, I noticed the space between the header and first row of cards jumped a bit. This PR adds the same top spacing to the expanded tab group `HStack`s to match a collapsed tab group's top padding.

Side note; I noticed this comment and padding to collapsed tab groups.

https://github.com/neevaco/neeva-ios/blob/c12dc3d7af8b80e729715ffba94d633ee21cb8d5/Client/Frontend/Browser/Card/CollapsableCardGroupView.swift#L80-L81

I tried removing this and that change would also fix the card jumping. Without this line, I did notice the top shadow of a card was clipped at one point. However after I saw it once, I was unable to get it to clip again.

Also before any changes I've noticed top card shadow clipping for the second row of cards in an expanded tab group, but again, not 100% consistent. Though I think with the change in this PR, it might actually fix that.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

| Before | After |
| --- | --- |
| tab-top-spacing-before | tab-top-spacing-after |

## Issues addressed
N/A


darinwf (approved):
LGTM

Would be good for Charles to also review since there could be something subtle here that I'm not thinking of.

chassu (approved):
I did remember the top card shadow issue wasn't very consistent. I might have forgot to think about the expanded case when I was fixing it... this fix should be good.

PR #3450, opened 2022-04-26
disable scrolling when card is animating

chassu:
Just played around with this change, I'm realizing two things:
1. not related to this change, but I found the card shrinks when you tap on it duration animation, which I think is not good. You can even see the switcher behind it.
2. there's a delay when the animation "looks like" it's finished, but the cardTransitionModel.state hasn't been reset to .hidden yet. It's a bit significant. During that time, vertical swiping could end up being detected as horizontal movement (seen in the video) 

I repro'ed both issue in the video. Thoughts?

The disabling is working as intended. I just realized 1, and then 2 is a side effect of this temporary change.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/87332917/165255848-6c5cc5a6-740b-483c-b9aa-a4e958f6542d.MP4


chassu:
Could it be our animation taking too much time to run?

chassu:
note to self: fix visiblefortrayhidden (one way)

darinwf (approved):

PR #3455, opened 2022-04-26
Add ability to reveal password when signing up

jedireza:
Adds the ability to reveal the password when creating a new account.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

## Issues addressed
🔒 https://app.clickup.com/t/270u299


Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -67,19 +68,38 @@ public struct EmailForm: View {
...
-                    SecureField("Password (required)", text: $password)
+                    ZStack(alignment: .trailing) {
+                        Group {
+                            if showPassword == true {
+                                TextField("Password (required)", text: $password)
+                            } else {
+                                SecureField("Password (required)", text: $password)
+                            }
+                        }
+                        .frame(height: 24.666667)
```
jedireza, line 79: I don't love this, but it was the only way I found to keep the height of the `TextField` and `SecureField` the same. I derived that number by using a `GeometryReader` and showing the default height of each of these fields.

A text field on the left and secure field on the right:
image

jedireza: For sure. I just added one.

chung-lai (approved):
LGTM. Would just suggest to double check this on other devices (smaller screen and iPad)

jedireza:
Thanks @chung-lai. I ran the simulator for both iPhone Mini and iPad.

| iPhone Mini | iPad 11 in. |
| --- | --- |
| Simulator Screen Shot - iPhone 13 mini - 2022-04-27 at 17 09 38 | Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-04-27 at 17 10 58 |

PR #3456, opened 2022-04-26
Escape HTML Encoding in Cheatsheet Recipe Title

EdwardJTL:
See screenshot for `'`

- Added extension in `Shared` module for decoding the response from GraphQL
- Only applied to cheatsheet title

@darinwf maybe we would want to apply this everywhere in the graphql decoding?

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
tried on that webpage, see screenshot

Simulator Screen Shot - iPhone 13 - 2022-04-26 at 15 54 00


darinwf:
Had you considered using an existing library for this like https://github.com/alexisakers/htmlstring ?

darinwf:
Here's another one: https://github.com/sonsongithub/htmlspecialcharacters

darinwf (approved):
Awesome, glad that worked out!

PR #3458, opened 2022-04-26
turn on drag and drop in switcher

chassu:
I was a bit worried about dragging a tab across several tab groups in a row. But so far it looks good.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/87332917/165401864-c4cf3cdf-3a48-48b8-a1b5-31711ff03d48.MP4


evancrow (approved):
This LGTM! Still might want @darinwf to stamp.

evancrow (changes requested):
Found a bug: If you reorder tabs and then close the app, the new order is not saved and when you reopen the app it goes back to the original position.

chassu:
Interesting. IIRC the operation modifies the `tabManager.tabs` array, but maybe that order needs to be stored into memory. I've always thought `tabManager.storeChanges` achieves that already.

chassu:
After running into many tricky problems with `dropEntered` (could get to states that are tricky to handle), I've decided to call `manager.rearrangeTabs` inside it. This could be better - we were getting to two view updates when we made separate calls on `tabCardModel.rearrangeAllDetails` and `manager.rearrangeTabs`. I also disabled dropping on pinned tabs and let pinned time do the ordering.

evancrow (approved):
LGTM!!

PR #3459, opened 2022-04-26
disable context menu and shrink during animation

chassu:
There used to be ways for the animation card to scale down (if tapped) or trigger the context menu (if long pressed).

### How was this tested?
- [x] I tested this manually

Before:

https://user-images.githubusercontent.com/87332917/165405636-df0761f2-bdde-494e-bfe4-513681781033.mp4

After:
Not doable after the fix.


evancrow (approved):
LGTM

PR #3462, opened 2022-04-27
Remove fetching recipe on recipe sites

EdwardJTL:
remove fetching recipe and only check domain
also removed `RecipeViewModel` which we no longer need

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
manually tested that the check works and the existing flows are not broken


darinwf (approved):
Thanks!

PR #3464, opened 2022-04-27
Basic CookieCutter Integration

evancrow:
This PR adds:

- The Cookie Cutter script, works!
- A basic model which can receive events
- Some minor UI updates
- Based on Tracking Prevention permission

I also had to update yarn/npm + webpack

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/37356984/165620412-58469b68-ba3c-4008-a0bf-41c99a930277.mp4

## Issues addressed
Issue: #3282


Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/CookieCutterHelper.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
+
+import { CookieEngine } from 'cookie-cutter';
+
+// Not used by iOS.
+CookieEngine.flagSite(async () => {});
```
jlynch630, line 6: Are we waiting on a second cut to implement this? Cookie cutter can cause reload loops without it

jlynch630 (approved):
Looks awesome!

Client/Frontend/Browser/CookieCutter/CookieCutterModel.swift
```diff
@@ -56,6 +58,11 @@ class CookieCutterModel: ObservableObject {
         }
     }
 
+    // MARK: - Methods
+    func cookieWasHandled() {
+
```
darinwf, line 63: Would be good to add a TODO comment here about what is needed.

Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/CookieCutterHelper.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
+
+import { CookieEngine } from 'cookie-cutter';
+
```
darinwf, line 4: Would be good to add TODO comments about the work that is still needed.

darinwf (commented):
This looks great! I think we may want to keep package-lock.json though so that our builds will be stable (release builds at branch point will be repeatable).

Client/Frontend/Browser/CookieCutter/CookieCutterModel.swift
evancrow: Will add a comment, but I don't think we really know what we wanna do here yet. I think this will be for education content later though right?

evancrow:
It appears that updating `webpack` has broken `FindInPage` :(

Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/CookieCutterHelper.js
evancrow: Didn't we discuss that this didn't need to be implemented? Thought the `getHostname` stopped the loop? If I do need to implement this, what exactly do I need to store? I would assume a dictionary of some kind.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -210,6 +210,8 @@ class BrowserViewController: UIViewController, ModalPresenter {
 
         super.init(nibName: nil, bundle: nil)
 
+        self.tabManager.cookieCutterModel = browserModel.cookieCutterModel
```
darinwf, line 213: Seems like it would be nice if `TabManager` did not need to know about cookie cutter. I see the `TabEventHandler` code though, so I get why this is plumbed through.

darinwf (approved):

Client/Frontend/Browser/BrowserViewController.swift
evancrow: Yes this is not ideal. In my next PR, I want to try and attach the `CookieCutterModel` directly to the tracking protection code, instead of plumbing through `TabManager`.

PR #3465, opened 2022-04-28
AB experiment for promo card

chung-lai:
A/B experiment for zero query promo card

| Preview Sign up | Default Browser |
| - | - |
| Screen Shot 2022-04-27 at 5 27 51 PM | Screen Shot 2022-04-27 at 5 52 01 PM |

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


hwiatneeva (commented):
From design -- LGTM. Thanks Chung!

hwiatneeva:
Oh one thing actually, the DB's X location has different padding. Let's match the paddings of the Sign UP promo's X. TY!
image

chung-lai:
> Oh one thing actually, the DB's X location has different padding. Let's match the paddings of the Sign UP promo's X. TY! image

Good catch @hwiatneeva!

darinwf (approved):

darinwf:
By the way, I noticed after dismissing both of these default browser and account creation promo cards, that I then started seeing this one:



I wonder if this will skew our data.

chung-lai:
> By the way, I noticed after dismissing both of these default browser and account creation promo cards, that I then started seeing this one:
> 
> 
> 
> I wonder if this will skew our data.

Hmm...let's only show it for users who have signed in before and never dismiss the preview user sign up promo card. Definitely need some clean up on the promo card logic soon :)

PR #3467, opened 2022-04-28
PopoverView Crash HotFix

evancrow:
Some Spaces views use the `OverlaySheetModel` as an `EnvironmentObject`, which was not passed down from the `PopoverView` and would cause the app to crash.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


darinwf (approved):
Thanks for the quick fix!

evancrow:
Skipping tests to land this for tonight's release, and since it only adds an `EnvironmentObject`.

PR #3476, opened 2022-04-28
Refresh user info to check verification status for spaces feature

jedireza:
There's currently a bug where OKTA signup email verification state doesn't update until after relaunch of the app. This happens because we only refresh the user info state during app launch.

In this PR we're re-fetching the `NeevaUserInfo` state when the user is interacting with the spaces feature and is logged in but not in a non-verified state.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Tested in the simulator.

The semi-dead time in the middle, where you can kind of see my task switcher, is where I go check email and click on the verification link.

https://user-images.githubusercontent.com/979929/165860854-ff7d70fc-404e-4e0d-b9d5-f2984d10cc36.mp4

## Issues addressed
🔒 https://app.clickup.com/t/22ygzbt


jedireza:
In the first video, because of how SwiftUI updates views `init` was being called when dismissing the verification prompt, which also triggered a user info reload.

Here is another test run, but this time I didn't click on spaces first. In this case we can see that the dialog will correct itself after the user info is reloaded, since we're observing that object.

https://user-images.githubusercontent.com/979929/165880699-b603d206-54ee-4da3-b93f-0b07a8e8b4b1.mp4

yozuysal (approved):
thanks Reza!

Client/Frontend/Overlay/Content/Spaces/AddToSpaceOverlayContent.swift
```diff
@@ -12,20 +12,33 @@ struct AddToSpaceOverlayContent: View {
...
     }
 
+    init(request: AddToSpaceRequest, bvc: BrowserViewController, importData: SpaceImportHandler?) {
+        self.request = request
+        self.bvc = bvc
+        self.importData = importData
+        self.userInfo = NeevaUserInfo.shared
+
+        // check if the user has verified their email since the last session update
+        if self.userInfo.isUserLoggedIn, !self.userInfo.isVerified, !self.userInfo.isLoading {
```
darinwf, line 33: As mentioned over zoom, might be more robust to put this in an `.onAppear` modifier. That way we are foiled by SwiftUI somehow caching this `View` and deciding not to run our `init` again.

PR #3478, opened 2022-04-29
Pass CookieCutter Prefs Down

evancrow:
### How was this tested?
- [x] I tested this manually

## Issues addressed
Issue: #3282


Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/CookieCutterHelper.js
```diff
@@ -1,41 +1,73 @@
...
+        switch (type) {
+        case 2:
+            return cookiePreferences.marketing;
+        case 3:
+            return cookiePreferences.analytic;
+        case 4:
+            return cookiePreferences.social;
+        default:
+            return true;
+        }
```
jlynch630, line 61: The engine categories are actually a little more granular, but we can switch over them and just generalize to the three categories we set user preferences for. Also probably a good idea to decline cookies if our preferences don't explicitly allow them.
```suggestion
        switch (type) {
        case CookieCategoryType.Marketing:
        case CookieCategoryType.DoNotSell:
            return cookiePreferences.marketing;
        case CookieCategoryType.Analytics:
        case CookieCategoryType.Preferences:
            return cookiePreferences.analytic;
        case CookieCategoryType.Social:
        case CookieCategoryType.Unknown:
            return cookiePreferences.social;
        default:
            return false;
        }
```

jlynch630 (approved):
Looks good, just the one note on category preferences.

Issue #3463, opened 2022-04-27
Tab cards appearing on top of spaces grid


darinwf:
Observed using 1.38.3

Steps to repro:

1) Visit web page
2) Add to spaces
3) Open spaces from "add to spaces" flow
4) Hit the back button on the spaces detail view

FYI @yozuysal @buracules - any ideas on this one? This is on today's release branch

darinwf:
<img src="https://user-images.githubusercontent.com/79867437/165606656-2397ecf9-9339-448a-a734-afb07ed9ee0e.jpg" width="300">

darinwf:
Hmm, these phantom cards appear stuck on top of the view. Switching between normal tabs and spaces doesn't get rid of them. Unclear if the repro steps above are really what leads to the issue.

Oh, these are collapsed (single row) tab groups that are spilling over into the spaces part of the grid.

@evancrow FYI -- could this be due to the card grid hotfix somehow?  513122872e0f79a146467f194bb80b1779e2edc2

darinwf:
If i scroll my tab groups so the right most card is visible and then switch to spaces, I no longer see the phantom tabs.

evancrow:
I doub it has to do with my change since that affected the width of the `CardScrollContainer` instead of specific tabs and tab groups. Maybe @chassu's recent changes are related?

chassu:
I don't think I have recent changes that touched on this. Question for @darinwf - when you scroll to the right most tab in your tab group and then switch to incognito mode, does it spill over?

darinwf:
I can no longer repro :-(

evancrow:
@darinwf Going to remove the release blocker label for now since since you can't repro

evancrow:
@darinwf I'm seeing this on iPad now... taking a look

PR #3480, opened 2022-04-29
Misc. TabGroup UI Fixes

evancrow:
Lots of tiny bugs I've wanted to fix for a while and the one big one!

Included:

- Fixed an issue where cards in a collapsed tab group would overlay a different section (caused by clipping being disabled, not sure why)
- Made the rounded corners of a collapsed tab group change to which end was visible (see video)
- Fixed a bug where the top of cards would be clipped
- Fixed a bug where a tab group would expand to the full width on iPad even if only 2 cards were present

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

iPad:

https://user-images.githubusercontent.com/37356984/165890774-acd52e0e-1e65-4422-a020-648360c21ef6.mp4

iPhone:

https://user-images.githubusercontent.com/37356984/165890907-487679c2-d56f-4928-87a0-310aae246674.mp4

## Issues addressed
Fixed: #3463


Client/Frontend/Browser/Card/CollapsableCardGroupView.swift
```diff
@@ -93,7 +123,6 @@ struct CollapsedCardGroupView: View {
                 // instantiated. This works by referencing an input parameter (groupDetails), which causes
                 // SwiftUI to think that this ViewModifier needs to be evaluated again.
                 let _ = groupDetails
-                scrollView.clipsToBounds = false
```
chassu, line 96: removing this is causing the card transition animation to get clipped

evancrow: Ohh okay, I'll set this to occur if the animation is happening. This is the cause of the bug where the tabs would overflow to another view.

chassu (approved):
LGTM!

Issue #3479, opened 2022-04-29
Add Option To Open All Items In Space


evancrow:
Similar to the web version of spaces, we should add an option from the three-dot menu in the SpacesDetailView to open all the links in a Space.

PR #3482, opened 2022-04-29
Add Open All Space URLs Option

evancrow:
Adds the option to open all Space URLs in new tabs. Also fixes a bug where ZombieTabs wouldn't load a URL because we never called the request, not sure how it's been working this whole time.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/37356984/165996973-2597f29e-d61d-4277-87ef-11a972146380.mp4

## Issues addressed
Fixed: #3479


evancrow:
With TabGroup change:

https://user-images.githubusercontent.com/37356984/166000780-e1de6f77-a40f-4b87-a488-68e055347901.mp4

Client/Frontend/Browser/Carousel/SimulatedSwipeController.swift
```diff
@@ -133,14 +133,16 @@ class SimulatedSwipeController:
...
 
-        if tab.canGoBack {
-            model.hidden = true
-        } else if let _ = tab.parent {
-            model.hidden = false
-            model.chromeModel.canGoBack = true
-        } else if let id = tab.parentSpaceID, !id.isEmpty {
-            model.hidden = false
-            model.chromeModel.canGoBack = true
+        DispatchQueue.main.async { [self] in
```
darinwf, line 136: Can you add a comment explaining why this DispatchQueue.main.async is needed?

Client/Frontend/Browser/Tabs/SavedTab+ConfigureExtension.swift
```diff
@@ -46,8 +46,15 @@ extension SavedTab {
     }
 
     func configureTab(_ tab: Tab, imageStore: DiskImageStore? = nil) {
-        // Since this is a restored tab, reset the URL to be loaded as that will be handled by the SessionRestoreHandler
-        tab.setURL(nil)
+        if sessionData == nil {
```
darinwf, line 49: Good catch on this. I think it would be better ideally to factor this bug fix out into its own PR, but guessing it is needed for opening all spaces to work? Can you also write some tests for this bug fix? Can add to `TabRestoreTests` using the same technique as the existing test to restore a closed tab and then confirm that it loads properly.

Client/Frontend/Browser/Carousel/SimulatedSwipeController.swift
```diff
@@ -133,14 +133,18 @@ class SimulatedSwipeController:
...
-        if tab.canGoBack {
-            model.hidden = true
-        } else if let _ = tab.parent {
-            model.hidden = false
-            model.chromeModel.canGoBack = true
-        } else if let id = tab.parentSpaceID, !id.isEmpty {
-            model.hidden = false
-            model.chromeModel.canGoBack = true
+        // DispatchQueue required to prevent a runtime crash,
+        // caused by updating the UI from not the main thread.
```
darinwf, line 137: This doesn't make sense to me. The code is not run on a background thread.

evancrow:
Marking as cherry-pick to get the ZombieTab bug fix in the release. The feature part of this PR is pretty small and more power user-focused, unlikely to be noticed.

Client/Frontend/Browser/Spaces/SpaceTopView.swift
```diff
@@ -197,6 +197,23 @@ struct SpaceTopView: View {
...
+    @ViewBuilder var openAllURLsButton: some View {
+        if let space = space, let urls = space.contentURLs {
+            Button {
+                let tabs = self.tabModel.manager.addTabsForURLs(Array(urls), shouldSelectTab: false)
+                for tab in tabs {
+                    tab.parentSpaceID = space.id.id
+                    tab.rootUUID = space.id.id
+                }
+
+                self.tabModel.manager.updateTabGroupsAndSendNotifications(notify: true)
```
darinwf, line 209: I think we should avoid calling this function from outside of `TabManager`. We can do that easily by adding a `rootUUID` parameter to `addTabsForURLs` method. I don't think we want to set `parentSpaceID` on all of the tabs. That attribute is used to make it so that hitting the back button on one of those tabs takes you back to the space, but we aren't navigating the user to each of the tabs in the newly created tab group. We also aren't selecting any of the tabs in that tab group. Maybe we should be? If we do, then only the tab that is selected would have the option to go back to the space.

ClientTests/ZombieTabTests.swift
```diff
@@ -0,0 +1,35 @@
...
+    var manager: TabManager!
+
+    override func setUp() {
+        super.setUp()
+
+        let profile = TabManagerMockProfile()
+        manager = TabManager(profile: profile, imageStore: nil)
+    }
+
+    func testZombieTabRestoredProperly() {
```
darinwf, line 20: nit: Let's just have this be part of the existing `TabRestoreTests` since it is also about tab restoration.

evancrow: This has to be a `ClientTest` as opposed to a `XCUITest` because there isn't a good UI flow to test this.

darinwf: Oh! I overlooked the fact that this is a ClientTest. D'oh! Good to prefer unit tests over integration tests when possible anyways, so LGTM

Client/Frontend/Browser/Spaces/SpaceTopView.swift
evancrow: I don't think we should select a Tab, I'm treating this more like open in background or restore tab, and letting the user choose what they want to open.

Client/Frontend/Browser/Carousel/SimulatedSwipeController.swift
evancrow: I took away the `DispatchQueue`, and I don't see the crash anymore anyways. I think it was a one-time fluke, so I removed it. Or my other code cleanup fixed it.

Client/Frontend/Browser/Spaces/SpaceTopView.swift
darinwf: OK in that case, we shouldn't set `parentSpaceID`.

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -29,6 +30,14 @@ extension TabManager {
             selectTab(newTabs.last, notify: true)
         }
 
+        if let rootUUID = rootUUID {
+            for tab in newTabs {
+                tab.rootUUID = rootUUID
+            }
+
+            self.updateTabGroupsAndSendNotifications(notify: true)
```
darinwf, line 38: This shouldn't be necessary. It is redundant with `tabsUpdatedPublisher.send()` below

darinwf: I think you should probably set the rootUUID properties before sending any notifications. Note, selectTab sends a notification, so setting rootUUID should happen before calling selectTab.

evancrow: `self.updateTabGroupsAndSendNotifications` is necessary, although we can set `notify` to `false`. Without it, the tab group is not created.

darinwf (approved):
LGTM, thanks!

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
darinwf: Oh got it. That's a good solution. Thanks!

PR #3483, opened 2022-04-29
Update Package.resolved

jedireza:
Xcode is making this change for me locally when I pull from `main`. I'm not sure if I should be committing this. It seems a little strange.

In #3456 there is a file `Neeva.xcworkspace/xcshareddata/swiftpm/Package.resolved` that was changed, but for me the file that Xcode is changing is `Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`.


jedireza:
Closing in favor of https://github.com/neevaco/neeva-ios/pull/3484

PR #3489, opened 2022-05-01
Make Text in NeevaScope Selectable

EdwardJTL:
Apply iOS 15 `textSelection` modifier to cheatsheet content to allow copy and share of plain text.

Downside: not highlightable.
I tried writing a wrapped UITextField view that supports highlighting and selection. It works, but the work required to make it conform to all the formatting we have (multi-line wrapping, trimming, font, etc.) is not worth it. This change is much simpler. We can pursue that later if we decide there's a specific view that really needs highlighting.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tried on both iOS 14 and 15.4 simulators

Simulator Screen Shot - iPhone 13 - 2022-05-01 at 16 36 53


Shared/UI/Util/SelectableTextModifier.swift
```diff
@@ -0,0 +1,33 @@
...
+    func body(content: Content) -> some View {
+        Group {
+            if #available(iOS 15, *) {
+                if allowSelection {
+                    content
+                        .textSelection(.enabled)
+                } else {
+                    content
+                        .textSelection(.disabled)
+                }
```
EdwardJTL, line 20: this cannot be inlined with a ternary expression because `.enabled` and `.disabled` are actually different types

darinwf:
What's the difference between "highlighting and selection" of text?

EdwardJTL:
> What's the difference between "highlighting and selection" of text?



> What's the difference between "highlighting and selection" of text?

ah by "select", i mean select all. The entire content in `Text` in "selected" and the "Copy" and "Share..." actions receive the whole string.
However, when the text is "selected", it is not highlighted (visually). With UITextField's, you can select a range, and range of text that is selected is also highlighted (in the accent colour and with two cursors on each end which you can move around to change the selection range).

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuView.swift
```diff
@@ -206,6 +206,7 @@ public struct CheatsheetMenuView: View {
                 }
             } else {
                 cheatsheetContent
+                    .selectableIfAvailable(allowSelection: true)
```
darinwf, line 209: minor nit: it seems like it would work to have an `init(_ allowSelection: Bool)` so you could just write `.selectableIfAvailable(true)` as that reads well already. The `allowSelection` doesn't really improve readability. It is just a bit redundant.

darinwf (approved):
LGTM

darinwf (approved):

PR #3494, opened 2022-05-02
move data related call from `init` to `onAppear`

jedireza:
Code hygiene change. This PR moves the data call introduced in https://github.com/neevaco/neeva-ios/pull/3476 out of `init` and into an `onAppear`.

Semi-related discussion on SO: 
- https://stackoverflow.com/questions/66523124/calling-functions-in-init-of-swiftui-view-and-observableobject

And while we did protect ourselves from spamming the request by checking if we're already mid-flight (`!isLoading`), `init` can be called a lot. So it seems reasonable to call this an anti-pattern.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

And the original bug fix is still in good shape.

https://user-images.githubusercontent.com/979929/166316640-45c7ef5b-e70c-485c-8055-fef0f9330d6c.mp4

## Issues addressed
- https://github.com/neevaco/neeva-ios/pull/3476#pullrequestreview-959379060


Client/Frontend/Overlay/Content/Spaces/AddToSpaceOverlayContent.swift
```diff
@@ -23,22 +23,13 @@ struct AddToSpaceOverlayContent: View {
...
-        if self.userInfo.isUserLoggedIn, !self.userInfo.isVerified, !self.userInfo.isLoading {
-            self.userInfo.reload()
-        }
-    }
-
     @ViewBuilder
     var content: some View {
         if userInfo.isUserLoggedIn, !userInfo.isVerified {
             EmailVerificationPrompt(email: userInfo.email ?? "", dismiss: hideOverlay)
+                .onAppear {
```
yozuysal, line 30: this lgtm! Just a note: We kind of get away with a more reliable onAppear here because of the @ViewBuilder and if else setup. if/else in a ViewBuilder recycles the outgoing View so we know that this onAppear will be called every time the condition changes.

But for a View body, onAppear can be not the most reliable thing to call these since swiftUI tends to keep them around (most of our Overlay content Views are like this for example). So much better to use onAppear on branch/leaf Views like you are doing here.

yozuysal (approved):
thanks!

Client/Frontend/Overlay/Content/Spaces/AddToSpaceOverlayContent.swift
jedireza: Thanks for mentioning that!

Shared/API/NeevaUserInfo.swift
```diff
@@ -97,7 +97,9 @@ public class NeevaUserInfo: ObservableObject {
     }
 
     public func reload() {
-        self.fetch()
+        if !self.isLoading {
```
darinwf, line 100: Nice, like moving the `isLoading` check here.

PR #3495, opened 2022-05-03
Cookie Cutter Education

evancrow:
Adds the Cookie Cutter onboarding UI, which will display the first time a cookie is blocked.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

Onboarding popup:

https://user-images.githubusercontent.com/37356984/166392133-1dd48915-6375-4b92-94da-85afacf0d245.mp4

Updated OverlaySheet Title/Buttons:

## Issues addressed
Issue #3282

cc @hwiatneeva @jonathangarelick


evancrow:
@darinwf Once this merges, should be ready for internal testing...

darinwf:
Nice work on this. From the first "Onboarding popup" video, it looks like the alignment of the close button is a bit off. Compare to the one for the add-to-spaces sheet. Maybe it is that the margins are different size? Best to check with @hwiatneeva on exact sizing of the margins / placement of the close button.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WalletConnectPresenter.swift
```diff
@@ -41,7 +41,6 @@ extension BrowserViewController: WalletConnectPresenter {
 
 extension BrowserViewController: ToastDelegate {
     func shouldShowToast(for message: LocalizedStringKey) {
-        print(message)
```
darinwf, line 44: Thanks!

Client/Frontend/Browser/CookieCutter/CookieCutterOnboardingView.swift
```diff
@@ -0,0 +1,79 @@
...
+                Button(action: onOpenMyCookieCutter) {
+                    Text("Open My Cookie Cutter")
+                        .withFont(.labelLarge)
+                        .foregroundColor(.brand.white)
+                        .frame(maxWidth: .infinity)
+                }
+                .buttonStyle(.neeva(.primary))
+
+                Button {
+                    NotificationPermissionHelper.shared.requestPermissionIfNeeded(
```
darinwf, line 51: Would be good to have @chung-lai review notifications permissions handling. Do we need to worry about the case of a user who already declined notifications permissions?

darinwf (approved):

Client/Frontend/Browser/CookieCutter/CookieCutterOnboardingView.swift
evancrow: > Nice work on this. From the first "Onboarding popup" video, it looks like the alignment of the close button is a bit off. Compare to the one for the add-to-spaces sheet. Maybe it is that the margins are different size? Best to check with @hwiatneeva on exact sizing of the margins / placement of the close button.

Should be 8px on the right side, will make the fix...

evancrow: I don't think so? In that case, probably best to not bother them if they already declined.

darinwf: I was wondering if we need to change the text of this button if they have already declined notifications.

evancrow: Oh, right :)... maybe it still works? Could keep it in our back pocket and show this again or something else at a later time.

evancrow:
Updated padding to match mocks (32 leading, 24 trailing):

Screen Shot 2022-05-03 at 11 18 16 AM

Client/Frontend/Browser/CookieCutter/CookieCutterOnboardingView.swift
chung-lai: If they already declined notification, should we take them to settings and turn on notification? Or show a dialog saying notification is off and ask if they want to navigate to settings?

chung-lai:
> @darinwf Once this merges, should be ready for internal testing...

I'll bring the new tracking protection setting to the cookie cutter settings screen so people can try out the whole experience. And we can finalize the tracking protection UI when @hwiatneeva has the design.

Client/Frontend/Browser/CookieCutter/CookieCutterOnboardingView.swift
evancrow: I think we could make a dialog, but I think that should be a separate PR/issue as that would apply to anywhere we request notifications...

PR #3505, opened 2022-05-04
Send logging error event for first run impression

chung-lai:
We observe there are some missing first run events on Mode and ad campaign data. We want to log an error for those cases to see how frequent this is happening. We want to capture the error for a couple key first run events and send the error message next time we successfully send any log events.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


darinwf (approved):
I see. So this will help us get a read on network errors contributing to the problem. Makes sense.

Issue #3503, opened 2022-05-04
History Tab Only Shows Up To Last Week


evancrow:
Might be an issue with sites not being loaded as the user scrolls.

PR #3508, opened 2022-05-05
Make HistoryList a LazyVStack

evancrow:
This was caused by `loadNextItemsIfNeeded` not being called since the views were already rendered, even if they were offscreen. Putting everything in a `LazyVStack` updates the model when it is scrolled.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
Fixed: #3503


darinwf (approved):
Of course! :-)

Good catch

PR #3509, opened 2022-05-05
Improve Restore Tab Performance

evancrow:
Makes sure tabs are restored as zombies, also I made it so that restoring tabs wouldn't override the selected tab (can change it back but it seemed to make more sense to me).

@chassu I wasn't able to recreate any of the bugs we saw :(

### How was this tested?
- [x] I tested this manually

- Open 500 tabs
- Closed 500 tabs
- Restore from Toast
- Almost **instantaneous**


chassu (approved):
Ooh, just realized that we didn't set the zombie argument. Maybe that was why I was experiencing lagginess from restoring tabs. LGTM

Issue #3502, opened 2022-05-04
Add Search to History UI


evancrow:
Saw this on an App Store review and it makes a lot of sense. Should add a search bar to History/Recently Closed to quickly find a site, might be hard to figure out searching through sites that have not been loaded yet though.

PR #3510, opened 2022-05-05
Add Search to History

evancrow:
Learned some SQL!

### How was this tested?
- [x] I tested this manually
- [x] I added automated tests

https://user-images.githubusercontent.com/37356984/166854968-5df26ccd-80c7-4a35-88b3-6e91d088c635.mp4

## Issues addressed
Fixed: #3502 

_adding automated tests_


Client/Frontend/History+RecentlyClosed/History/HistoryPanelModel.swift
```diff
@@ -47,6 +47,17 @@ class HistoryPanelModel: ObservableObject {
...
 
+    // History search
+    @Published var siteFilter = "" {
+        didSet {
+            searchForSites(with: siteFilter)
+        }
+    }
+    @Published var filteredSites = DateGroupedTableData<Site>()
+    var isSearching: Bool {
+        !siteFilter.isEmpty
```
EdwardJTL, line 58: should this be named as `useFilteredSites` or something? The current name suggests if searching is actively happening, but i don't see this Bool being cleared.

Client/Frontend/History+RecentlyClosed/History/HistoryPanelModel.swift
```diff
@@ -47,6 +47,17 @@ class HistoryPanelModel: ObservableObject {
 
     @Published var groupedSites = DateGroupedTableData<Site>()
 
+    // History search
+    @Published var siteFilter = "" {
+        didSet {
+            searchForSites(with: siteFilter)
+        }
+    }
+    @Published var filteredSites = DateGroupedTableData<Site>()
```
EdwardJTL, line 56: could we link these two together with a combine pipeline that also uses a debounce? 
also if searching can be performed on a background thread, we can insert that into the pipeline as well

Storage/SQL/SQLiteHistory.swift
```diff
@@ -682,6 +682,34 @@ extension SQLiteHistory: BrowserHistory {
         return db.runQueryConcurrently(
             sql, args: nil, factory: SQLiteHistory.iconHistoryColumnFactory)
     }
+
+    public func getSitesWithQuery(query: String) -> Deferred<Maybe<Cursor<Site?>>> {
+        let sql = """
+            SELECT
```
EdwardJTL, line 688: not an expert on running sql in swift. would it be possible to reuse some of this code with `getSitesByLastVisit`? for example, breaking it down to which attributes need to be selected and which schemas should be used.

EdwardJTL (commented):
this is pretty cool!

Storage/SQL/SQLiteHistory.swift
evancrow: Uh I am not an expert either (literally just learned what I needed to do this)... but I don't think so?

Client/Frontend/History+RecentlyClosed/History/HistoryPanelModel.swift
evancrow: Searching has to be done from the main thread I'm pretty sure, at least that's how the code we inherited from Firefox works. Adding the debounce object though, pretty neat, didn't know that was a thing till now!

Client/Frontend/History+RecentlyClosed/History/DebounceObject.swift
```diff
@@ -0,0 +1,23 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+// Based from https://onmyway133.com/posts/how-to-debounce-textfield-search-in-swiftui/
```
evancrow, line 4: @darinwf Is this the right way to attribute code usage?

Storage/SQL/SQLiteHistory.swift
EdwardJTL: I guess we can take the common part, like select, from, where, group by, and order by parts into substrings, then share them between the two functions and join them to build the final query we want. 
Ideally something like https://github.com/vapor/sql-kit might be better because it is more readable; we would only be using their expression class, but i'm not sure if we need write additional code to convert it to a string.

maybe for now just leave a comment above this query and the one in the other function to remind people they may need to update both if they change one?

EdwardJTL (approved):

Storage/SQL/SQLiteHistory.swift
evancrow: Created functions for large portions of shared code and added a comment.

Client/Frontend/History+RecentlyClosed/History/DebounceObject.swift
darinwf: Yeah, looks good. Couple nits: "from" -> "on" and I'd put this comment just above the `class DebounceObject..` line.

Client.xcodeproj/project.pbxproj
```diff
@@ -3157,6 +3160,7 @@
 		206954A927FF44A2000360D6 /* History */ = {
 			isa = PBXGroup;
 			children = (
+				20318F242823786A003B03C1 /* DebounceObject.swift */,
```
darinwf, line 3163: Perhaps `DebounceObject` should live in a more general location? Maybe `Client/Utils`?

Client/Frontend/History+RecentlyClosed/History/DebounceObject.swift
```diff
@@ -0,0 +1,23 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Combine
+import Foundation
+
+// Based on https://onmyway133.com/posts/how-to-debounce-textfield-search-in-swiftui/
```
darinwf, line 8: Might be good to also add a code sample here in the comments showing how to use this class.

darinwf (approved):
Test drove and it looks good to me. I didn't try with a large history db though. You should probably verify that it performs well in that case. We might find that it would be good to add a progress indicator / spinner while the SQL query is running in case it takes a long time to complete.

evancrow:
Ran some tests with 100s of tabs and I can't notice a difference in performance

Issue #3439, opened 2022-04-25
CardGrid Should Scroll After Closing Bottom Tab


evancrow:
https://user-images.githubusercontent.com/37356984/165156754-3461036d-da64-47fb-80cc-64bc5018aaca.mp4

After closing the tab in the bottom row, the rest of the `CardGrid` should move down and the empty space should disappear.

FYI @chassu

chassu:
hmmm, didn't see this on the app store build on my device, but saw it on my simulator on main... wondering if it's a regression

chassu:
I'm not able to repro on my testing device. Do you see this on a real device?

evancrow:
Both. I've found it easy to recreate when using a mix of tabs and tab groups and then having a single tab on the bottom.

darinwf:
I saw this bug this morning too w/ latest TestFlight.

chassu:
I've tried 
1. setting the maxWidth and maxHeight of the scroll view to the container size
2. add useEffect(deps: tabModel.allDetails){scrollToSelectedTab()} to CardScrollContainer. I think that broke some stuff and the scrollToSelectedTab didn't solve the issue.

Any ideas?

evancrow:
My guess is the `scrollToSelectedTab` didn't work because the next tab to select is already partially visible, right? Maybe we need a method to scroll to the tab above the selected tab?

evancrow:
@chassu I think this may be related to the `ArchivedTabsView` somehow. Works fine when that's enabled and the bug did not exist before that PR merged.

chassu:
just want to point out that this bug already exists before the `ArchivedTabsView` PR was merged 😢

chassu:
I think I've found a way to consistently repro it - make the bottom tab a zombie tab (can go to debug settings and turn tabs into zombie tabs), select the tab, go back to switcher, and then close the tab. I'm suspecting that the scrollView bug happens when webView of the bottom closed tab is still loading.

evancrow:
Fixed by: #3514

PR #3514, opened 2022-05-05
scrollView should be pushed down when bottom tab is closed

chassu:
I think the issue is caused by the animation being nil. It was set to nil to fix the stuttering bug. I introduced another state `gridModel.tabsDidChange` to change the animation to `.default` temporarily to work around this problem. This change provides animation when pushing down the scrollView as a benefit.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/87332917/166888335-03007b18-c182-4b44-bd8b-34e62e13b030.MP4

## Issues addressed
#3439


Client/Frontend/Browser/Card/GridModel.swift
```diff
@@ -32,6 +32,15 @@ class GridModel: ObservableObject {
     }
     @Published var gridCanAnimate = false
     @Published var switchModeWithoutAnimation = false
+    @Published var tabsDidChange = false {
+        didSet {
+            if tabsDidChange == true {
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
```
evancrow, line 38: This does not seem ideal but I don't know what else we could do... @darinwf any thoughts?

darinwf: At minimum, I think what's missing here is a detailed comment explaining why we need this code and importantly why we can't solve this without resorting to `asyncAfter`.

Client/Frontend/Browser/Card/CardsContainer.swift
```diff
@@ -144,8 +144,14 @@ struct CardScrollContainer<Content: View>: View {
...
             ScrollViewReader(content: content)
         }
-        // Fixes a bug where scrollView would stutter at the edge
-        .animation(gridModel.gridCanAnimate ? .interactiveSpring() : nil)
+        // Fixes two animation bugs:
+        // 1. scrollView would stutter at the edge without making the animation nil
+        // 2. scrollView wouldn't push down when the bottom tab is closed if the
+        // animation is nil
+        .animation(
+            (gridModel.gridCanAnimate
```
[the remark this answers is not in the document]
chassu: `.interactiveSpring()` was too fast for the push down animation. The default one looks a lot smoother to me.

https://user-images.githubusercontent.com/87332917/167031292-096f879b-e7b6-4963-a4c4-f75416285321.MP4

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -117,7 +119,14 @@ class TabCardModel: CardModel {
...
             self?.manager.didRestoreAllTabs ?? false
         }).sink { [weak self] in
+            self?.tabsDidChange = true
             self?.onDataUpdated()
+            // 'tabsDidChange' is used by CardScrollContainer to set its animation
+            // to .default. This is needed to handle a bug which the scroll view
+            // doesn't get pushed down when the bottom tab is closed.
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                self?.tabsDidChange = false
+            }
```
chassu, line 129: I purposely did not make `tabsDidChange` a published variable and leave the update to `self?.onDataUpdated()` because there will be a objectWillChange downstream. I was worried it will be too chatty if I make it a published variable. Do you think this makes sense? @darinwf

evancrow (approved):
Awesome work @chassu, LGTM but should wait for @darinwf final review.

Client/Frontend/Browser/Card/CardModels.swift
darinwf: Good! Just add a comment explaining it :)

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -49,6 +49,8 @@ class TabCardModel: CardModel {
     static let todayRowHeaderID: String = "today-header"
     static let lastweekRowHeaderID: String = "lastWeek-header"
 
+    var tabsDidChange = false
```
darinwf, line 52: nit: Can this be `private(set)`?

Client/Frontend/Browser/Card/CardsContainer.swift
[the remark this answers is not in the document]
darinwf: ^^^ Is this a video of using `interactiveSpring` or `default`?

Client/Frontend/Browser/Card/CardModels.swift
darinwf: Sounds good. That is consistent with all of the other state variables on `TabCardModel`.

Client/Frontend/Browser/Card/CardsContainer.swift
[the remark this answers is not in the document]
chassu: ^^^ this is `interactiveSpring`

It's a lot faster compared to the video using `default` at the top

darinwf (approved):

PR #3515, opened 2022-05-05
Update package-lock.json and yarn.lock

darinwf:
This is follow-up from 651eec0c06024e1c68aa3783367f3975e17e40a8


chung-lai (approved):
Thanks! Was about to fix it after today's build :P

PR #3520, opened 2022-05-05
Change How OverlaySheet Animations Are Managed

evancrow:
Moved control of animations from `OverlaySheetManager` to `OverlayManager`. Fixes use cases when `animate` was set to `false`.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/167031339-982bae49-53bd-45f0-ac5e-510b20af0b8e.mp4

No animation:

https://user-images.githubusercontent.com/37356984/167031352-1f1e1ae3-903c-4080-860a-a1a40fe26bcb.mp4


Client/Frontend/Overlay/Sheet/OverlaySheetView.swift
```diff
@@ -25,6 +25,8 @@ enum OverlaySheetUX {
 
     /// Bottom padding applied to the sheet content
     static let bottomPadding: CGFloat = 18
+
+    static let animationOffset: CGFloat = 500
```
chung-lai, line 29: i guess i'm not familiar with Overlay manager. Does all sheets have the same offset even they may have different height?

Client/Frontend/Overlay/Sheet/OverlaySheetView.swift
```diff
@@ -252,7 +254,10 @@ struct OverlaySheetView<Content: View, HeaderContent: View>: View, KeyboardReada
         GeometryReader { outerGeometry in
             DismissBackgroundView(
                 opacity: model.backdropOpacity, position: model.position,
-                onDismiss: style.nonDismissible ? {} : onDismiss)
+                onDismiss: style.nonDismissible ? {} : onDismiss
+            )
+            .animation(nil)
```
chung-lai, line 259: not sure I understand why we need the fade transition here? Is it so that we don't need `self.backdropOpacity = OverlaySheetUX.backdropMaxOpacity` in OverlaySheetModel?

Client/Frontend/Overlay/OverlayManager.swift
```diff
@@ -245,9 +247,15 @@ class OverlayManager: ObservableObject {
             }
 
             switch overlay {
-            case .backForwardList, .sheet:
+            case .backForwardList:
                 slideAndFadeOut(offset: 0)
             case .fullScreenModal, .popover:
+                if case .popover = overlay {
```
chung-lai, line 253: i guess same question here, is this change so that we don't need the animation in `OverlaySheetModel`?

chung-lai (commented):
Thanks for fixing this! Left some questions and look like this change the way how animation is done for various types of overlays. May worth trying out multiple overlay types and make sure there is no other side effect.

Client/Frontend/Overlay/Sheet/OverlaySheetView.swift
evancrow: The animate offset doesn't necessarily have to be the height, just has to be a large enough distance to make the slide in/out look right.

Client/Frontend/Overlay/Sheet/OverlaySheetView.swift
evancrow: That's exactly it.

Client/Frontend/Overlay/OverlayManager.swift
evancrow: Also yes.

evancrow:
Should only affect the `OverlaySheet` and their `Popover` variants.

chung-lai:
> 

Didn't verify it but this morning when I run my branch (which may have this change) and there was a popover display bug that it shows off screen when I tap on the shield icon. May want to make sure it's not due to this change.

evancrow:
Wasn't able to repro with this PR, but also those views are not affected by the `OverlayMananger` and `OverlayView`. Those are a different type of popover.

chung-lai (approved):

PR #3525, opened 2022-05-06
Skip preview home and enhance zero query UX in preview mode

jedireza:
This PR begins the enhancement plans to our zero query UX for preview mode customers. Specifically it addresses the following:
- Skip the preview home view (gradient style welcome screen) and show zero query instead (keyboard not showing by default; i.e. w/o cursor focus on the URL bar)
- We only show section header expand/collapse chevrons after the user has done a first navigation
- Query suggestions now evolve as the users searches and the three example queries eventually disappear.
- Suggested sites are now active for preview mode users (and isn't seeded with defaults anymore)
    - If the user has ever signed in, we include the Neeva home site
- Spaces; for preview mode users we show spaces from the maker community (From the Neeva Community).
- Spaces; for logged in users show their spaces

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

#### Neeva Client

https://user-images.githubusercontent.com/979929/167544536-14244fc9-a563-4548-8781-b214026c0024.mp4

#### XYZ Client

https://user-images.githubusercontent.com/979929/167544816-1317028f-e8ca-4c17-b3c2-dda7b1c0640f.mp4

## Issues addressed
🔒 https://app.clickup.com/t/2w2yg4r
🔒 https://app.clickup.com/t/2w2yg08
🔒 https://app.clickup.com/t/2w2yg81


darinwf:
Test drove. I think it would be nice to avoid the transition on the zero-query content when giving focus to the location bar:

https://user-images.githubusercontent.com/79867437/167264248-fe39d9b5-f35c-4ff6-ac7c-08614fc510e9.mp4

Ideally, the zero-query UI is unchanged when focus is given, and we just animate in the cancel button and nothing else.

jedireza:
Yep I noticed the same thing about the animation. I’ll look at correcting that.

Client/Frontend/Browser/BrowserBottomBarView.swift
```diff
@@ -14,7 +15,9 @@ struct BrowserBottomBarView: View {
     @EnvironmentObject var tabContainerModel: TabContainerModel
 
     @ViewBuilder var toolbar: some View {
-        if !browserModel.showGrid && !chromeModel.inlineToolbar && !chromeModel.isEditingLocation {
+        if !browserModel.showGrid && !chromeModel.inlineToolbar && !chromeModel.isEditingLocation
+            && Defaults[.didFirstNavigation]
```
yozuysal, line 19: rather than hiding the toolbar under this condition, can we move the Preview UI to be in front of the toolbar and have it cover it? (ie or in some other way move the business logic of Preview UI closer to Preview UI)

if that is gonna get involved, we will need to add `(Defaults[.didFirstNavigation] || NeevaConstants.currentTarget == .xyz) `here

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -404,10 +404,8 @@ class BrowserViewController: UIViewController, ModalPresenter {
             }
         }
 
-        if FeatureFlag[.enableSuggestedSpaces] {
-            DispatchQueue.main.async {
-                SpaceStore.suggested.refresh()
-            }
+        DispatchQueue.main.async {
```
yozuysal, line 407: we will need to check NeevaConstants.currentTarget == .client here

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -525,7 +523,7 @@ class BrowserViewController: UIViewController, ModalPresenter {
                     }
                 #else
                     if !Defaults[.didFirstNavigation] {
-                        self.showPreviewHome()
+                        self.showZeroQuery()
```
yozuysal, line 526: hmm, this we may want to check with @hwiatneeva , I think the current thought it more towards making a PreviewHome that uses components from ZeroQuery but looks slightly different (with a search box etc)

Client/Frontend/ZeroQuery/SuggestedSearchViews.swift
```diff
@@ -12,26 +12,28 @@ struct SuggestedSearchesView: View {
     @Environment(\.onOpenURL) private var openURL
     @Environment(\.setSearchInput) private var setSearchInput
 
+    let exampleQueries = ["best headphones", "lemon bar recipe", "react hooks"]
```
yozuysal, line 15: rather than being in the View here, should we move the business logic to the model itself? Then model.suggestedQueries comes in at 3 at all times and we dont need to duplicate view code below. So all the (3-...) etc logic happens inside the model.wdyt?

Client/Frontend/Browser/BrowserViewController.swift
darinwf: We met last Thursday and identified this UX. The idea of a search box w/ Neeva wordmark above is also interesting and probably worth building toward. But we thought this was a good intermediate step. Also, thinking ahead to such a design, we probably still want to have URL bar w/ zero-query UI below so that we can transition to the normal URL bar and zero-query UI in a smooth fashion. Current "preview home" -> "URL bar + zero-query" transition is not great.

Client/Frontend/Browser/BrowserBottomBarView.swift
jedireza: Yeah, preview home is gone and we're working with zero query here. I added a check for `.xyz` as you mentioned. I did notice that the XYZ client isn't showing the intro screens anymore so I need to look at that.

Client/Frontend/Browser/BrowserViewController.swift
jedireza: I wrapped this in a check for `.client`.

Client/Frontend/ZeroQuery/SuggestedSearchViews.swift
jedireza: Yep that's where I started. I wanted to get something working before making any larger refactor changes. But since the UX is looking good, I just went through and refactored this.

Client/Frontend/Browser/BrowserViewController.swift
hwiatneeva: Yup. This step is necessary, and likely, we'll keep this way for a while. This state(preview 1st zero query) lasts only briefly. 

// A better home: we need to allow ourselves some months at least to think of a better direction. We need to give our full attention to this incremental experiment(s) and other current key features first. Cheers!

Client/Frontend/Browser/BrowserTopBarView.swift
```diff
@@ -19,10 +20,6 @@ struct BrowserTopBarView: View {
     @Environment(\.horizontalSizeClass) private var horizontalSizeClass
     @Environment(\.verticalSizeClass) private var verticalSizeClass
 
-    private var isShowingPreviewHome: Bool {
-        tabContainerModel.currentContentUI == .previewHome
```
darinwf, line 23: looks like you can also remove the `@EnvironmentObject var tabContainerModel: TabContainerModel` line above, which should help a tad with performance (one less thing causing this View to be invalidated)

Client/Frontend/Browser/BrowserView.swift
```diff
@@ -76,10 +76,6 @@ struct BrowserView: View {
     @State var safeArea = EdgeInsets()
     @State var topBarHeight: CGFloat = .zero
 
-    private var isShowingPreviewHome: Bool {
-        tabContainerModel.currentContentUI == .previewHome
```
darinwf, line 80: ditto... can eliminate the `tabContainerModel` var now.

Client/Frontend/ZeroQuery/ZeroQueryHeader.swift
```diff
@@ -16,17 +17,19 @@ struct ZeroQueryHeader: View {
...
-                    // This button is not an accessibility element.
-                    Symbol(decorative: icon, size: ZeroQueryUX.ToggleIconSize, weight: .medium)
-                        .frame(
-                            width: ZeroQueryUX.ToggleButtonSize,
-                            height: ZeroQueryUX.ToggleButtonSize,
-                            alignment: .center
-                        )
-                        .background(Color(light: .ui.gray98, dark: .systemFill)).clipShape(
-                            Circle())
+                if Defaults[.didFirstNavigation] {
```
darinwf, line 20: nit: Another idea might be to simply hide the toggle UI when `action` is nil, and then pass `nil` when we want to hide the toggles, when `didFirstNavigation` is false. That way we keep configuration of the initial zero-query screen more localized to fewer files.

XCUITests/ZeroQueryTests.swift
```diff
@@ -5,42 +5,4 @@
 import Foundation
 
 class ZeroQueryTests: BaseTestCase {
-    func testZeroQueryHidesAfterLoadingURL() {
```
darinwf, line 8: Seems like there are probably a new set of tests to write though. Can be a follow-up PR, but would be good to have coverage for all the new behaviors introduced by this PR.  e.g., that navigating to a site results in the site being added to an otherwise empty suggested sites section.

darinwf (approved):

Storage/DefaultSuggestedSites.swift
```diff
@@ -34,48 +34,7 @@ open class DefaultSuggestedSites {
     ]
 
     public static let sites = [
```
yozuysal, line 36: I might have missed this convo, so just checking if it is intentional.

This will make us not show any of the current defaults even after first navigation.

yozuysal (approved):
lgtm % one question about default suggested sites. I know we talked about overhauling the default sites list but wanted to check in if removing them is intentional

Client/Frontend/ZeroQuery/ZeroQueryHeader.swift
jedireza: Yeah, I don't love the peppered checks everywhere. I changed this to have an explicit API by adding a property `hideToggle`.

XCUITests/ZeroQueryTests.swift
jedireza: Agreed. Issue filed: https://github.com/neevaco/neeva-ios/issues/3559

Storage/DefaultSuggestedSites.swift
jedireza: Yes it was intentional. They populate as the user visits sites. We hide this whole section until there are sites though, so there won't be a header floating all by itself.

yozuysal: But we check `viewModel.suggestedSitesViewModel.sites.count > 0` so it will show only one and then two etc.. With the default list it would start filled up and then swap to more personalized ones as the user visits.

Client/Frontend/Browser/BrowserView.swift
jedireza: Thanks. Fixed.

chung-lai:
@jedireza, If we want this to go out in this week's release, we need to cherry pick it. Or we want to have some bake time and release next week?

jedireza:
@chung-lai thanks. Let's see if folks are good with https://github.com/neevaco/neeva-ios/pull/3567 and then I think we'll be in good shape.

PR #3530, opened 2022-05-06
Remove TabManager.didRestoreAllTabs as it is unnecessary now

### LINKED ISSUE #3518: Eliminate `tabManager.didRestoreAllTabs`
We shouldn't need this. The `TabManager` should be smart enough to throttle tabsUpdated events properly without needing clients to check if tab restoration is happening.

darinwf:
Given the suppression of notifications that we already do during
tab restore, we don't need this flag. It is redundant.

Confirmed by adding code to the one spot in CardModels where
we were checking for a false value and saw that it is never false
at that spot. So it is safe to remove.

Fixes #3518


chassu (approved):
LGTM

PR #3533, opened 2022-05-06
History Scroll Fixes

evancrow:
The way the index was calculated was flawed because it was the index out of the current section, not all sites. Also, fixes a `ScrollView` jumping bug by unfortunately removing the LazyVStack and removing sticky headers.

### How was this tested?
- [x] I tested this manually

Uploading Simulator Screen Recording - iPad Pro (12.9-inch) (5th generation) - 2022-05-06 at 12.56.45.mp4…


darinwf:
@evancrow - How are you testing this? Do you have a way of generating a history DB for testing that has a lot of navigations, old and new?

darinwf (approved):

evancrow:
> @evancrow - How are you testing this? Do you have a way of generating a history DB for testing that has a lot of navigations, old and new?

A few ways:

- I have an old simulator that has a lot of past navigations
- Go to Wikipedia and tap Random Article a few hundred times

PR #3537, opened 2022-05-07
Default browser button changes

chung-lai:
Changes to default browser buttons for better UX

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

Open to settings:

https://user-images.githubusercontent.com/87332578/167239513-0d6c30d8-5a85-45f6-a22b-a0c629f8f42b.mov

Resume after background kill:

https://user-images.githubusercontent.com/87332578/167239520-64cbc408-35fb-46e4-bd79-7f3373e64d82.mov

| Promo card | Settings |
| --- | --- |
| Screen Shot 2022-05-06 at 10 01 50 PM | Screen Shot 2022-05-06 at 10 02 04 PM |


chung-lai:
@hwiatneeva, @darinwf, still WIP but post screenshots for any early feedback.

darinwf (approved):
LGTM...

One thing I notice is that in both the control and variant arms of this experiment, we will restore the DB interstitial. Perhaps we should avoid doing that in the control case so that we preserve existing behavior. Thoughts?

chung-lai:
> LGTM...
> 
> One thing I notice is that in both the control and variant arms of this experiment, we will restore the DB interstitial. Perhaps we should avoid doing that in the control case so that we preserve existing behavior. Thoughts?

Nice catch, will make sure control doesn't see the screen when user foreground the app

PR #3542, opened 2022-05-09
Updates to XYZ build IA

yozuysal:
Tune down the landing experience by removing some of the screens:
- First run is only launched through clicking the wallet button
- Remove NFT suggestions and relevant first run screen.
- Update zero query suggestions and title.
- Don't show Home and the NFT Views
- Prompt to import wallet if the user tries to connect to a dApp
- Update hexagons to be more reliable for light mode using a Path

### How was this tested?
- [X] I tested this manually
- [ ] I added automated tests


buracules (approved):

Client/Frontend/ZeroQuery/View+Extensions.swift
```diff
@@ -45,37 +45,89 @@ extension View {
...
-                .clipShape(Hexagon().rotation(.degrees(30)))
-                .blendMode(.multiply)
-        }
+    public func hexagonClip(with size: CGFloat) -> some View {
+        self
+            .frame(width: size * 4 / 5, height: size * 4 / 5)
+            .background(Color.white)
+            .clipShape(Hexagon())
+            .frame(width: size, height: size)
+            .background(WalletTheme.gradient.opacity(0.2))
```
buracules, line 54: I think we need XYZ control here. We need to import WalletCore for WalletTheme

buracules, line 54: I think we need XYZ control here. We need to import WalletCore for WalletTheme

buracules (changes requested):

Client/Frontend/ZeroQuery/View+Extensions.swift
yozuysal: thanks for catching this! fixed.

buracules (approved):

PR #3547, opened 2022-05-09
Log neeva token using a restful api

chung-lai:
Logging the attribution token using our server API so we can resolve all attributes server side instead of relying on Client side. Will keep both client side and server side resolver for a little while to make sure the migration is looking good.

Not test yet. Will merge once the server with new API is launched.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


darinwf (commented):
What is your thinking around deleting the existing reporting code?

chung-lai:
> What is your thinking around deleting the existing reporting code?

Do you mean the code hitting the apple server? I'm thinking to at least keeping it for one release to make sure the pipeline is working as intended. And then we should delete the existing client-side resolving code.

darinwf:
> Do you mean the code hitting the apple server? I'm thinking to at least keeping it for one release to make sure the pipeline is working as intended. And then we should delete the existing client-side resolving code.

Yes, that's what I was wondering about. SGTM

chung-lai:
> What is your thinking around deleting the existing reporting code?

Added TODO comment and tested on m1 where the rest api was deployed. PTAL

Client/Logging/ConversionLogger.swift
```diff
@@ -89,6 +89,29 @@ class ConversionLogger {
...
+                var neevaTokenRequest = URLRequest(url: NeevaConstants.neevaTokenApiURL)
+                neevaTokenRequest.httpMethod = "POST"
+                let neevaTokenData = "sessionUUID=\(Defaults[.sessionUUIDv2])&aaaToken=\(token)"
+                neevaTokenRequest.httpBody = Data(neevaTokenData.utf8)
+                neevaTokenRequest.setValue(
+                    "application/x-www-form-urlencoded", forHTTPHeaderField: "Content-Type")
+
+                URLSession.shared.dataTask(with: neevaTokenRequest) { data, response, error in
+                    guard error == nil else {
+                        DispatchQueue.main.async {
```
darinwf, line 104: No need for this `DispatchQueue.main.async` right? `logNeevaRequestError` already includes that.

Client/Logging/ConversionLogger.swift
```diff
@@ -168,4 +191,20 @@ class ConversionLogger {
             )
         }
     }
+
+    private static func logNeevaRequestError(
```
darinwf, line 195: nit: Add a comment indicating that this function can be called on a background thread.

darinwf (approved):

Issue #3534, opened 2022-05-06
Buggy tracking protection whitelist toggle 


chung-lai:
When you open the shield icon and toggle the tracking prevention button, it will whitelist the site and disable tracking protection for that site. When you navigate to another site, the shield icon updates the badge but the stats in the popover and the toggle do not reflect the right state which it stays off.

Upon restarting the app, it's back to normal but when you navigate to a whitelist site, the toggle in the popover stays ON even though the tracking protection is off at that time.

See video for detail

https://user-images.githubusercontent.com/87332578/167218051-c1a988ef-7042-4c2e-8c08-085f106ec610.mov

Issue #3521, opened 2022-05-05
Disabling cookie cutter from the shield menu should be a per-site action


darinwf:
What I notice on latest TestFlight is that disabling on one site from the shield appears to disable cookie cutter on other sites. But I also notice that it does not appear disabled in Settings, so I'm not quite sure what is broken.

This is with the "Cookie Cutter" feature flag enabled.

@evancrow FYI

evancrow:
I don't think there is a disabled state in settings. @hwiatneeva should there be a toggle in settings to disable Cookie Cutter? There's one for tracking protection so maybe that should just be replaced by Cookie Cutter (since the tracking protection state controls the Cookie Cutter state)?

PR #3549, opened 2022-05-09
Fix Incorrect Tracking Protection State

evancrow:
First, we weren't even updating the tracking state when the URL changed (main issue). The state setting was flakey as we store if tracking is allowed but set the foreground state as protection enabled (confusing).

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #3521
Fixed: #3534


content-blocker-lib-ios/src/TrackingPreventionConfig.swift
```diff
@@ -24,11 +24,22 @@ struct TrackingPreventionConfig {
         Defaults[.unblockedDomains].remove(domain)
     }
 
-    static func trackersAllowedFor(_ domain: String) -> Bool {
-        Defaults[.unblockedDomains].contains(domain)
+    static func trackersPreventedFor(_ domain: String) -> Bool {
```
chung-lai, line 27: I think we have this concept of whitelisting a domain and allow trackers for them. Is renaming this to `PreventedFor` make it easier to understand in the code? I found `trackersAllowedFor` easier to read somehow 😅

chung-lai (approved):
LGTM. Just left one feedback on renaming. Thanks for fixing this. Wonder if we are confident about this change or merge this after branch for some bake time of the change.

content-blocker-lib-ios/src/TrackingPreventionConfig.swift
evancrow: Makes sense, will change it back!

evancrow: Actually I'll keep both, make them two separate functions...

evancrow:
> Wonder if we are confident about this change or merge this after branch for some bake time of the change.

I would like to get this into this week's build, I think it works well and we will still have a few days to mess with it.

chung-lai (approved):
Latest change still look good!

evancrow:
Tests already passed, going to merge to make tonight's TestFlight

PR #3552, opened 2022-05-10
Update SpaceContainerView init logic

buracules:

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases

  and dark mode

## Issues addressed


Client/Frontend/Browser/Card/AddOrUpdateSpaceView.swift
```diff
@@ -240,19 +240,18 @@ struct AddOrUpdateSpaceView: View {
                     switch config {
                     case .addSpaceItem:
                         let data = SpaceEntityData(
-                            id: space.id.id,
+                            id: UUID().uuidString,
```
buracules, line 243: @yozuysal this is currently buggy if u try add more than one item at the same time it will assign same id to every each one of them and that will break the forEach.

buracules:
@jonathangarelick just a heads up i am changing initialization of the SpaceContinerView.

Client/Frontend/Browser/Card/AddOrUpdateSpaceView.swift
yozuysal: space entity itself should have an id , but I think this is the data that is temporarily constructed right? then it makes sense to just assign something random as you did to make the optics work, and then when we refresh the space (which should already happen now) it would use its actual id

Client/Frontend/Browser/Card/CardDetails.swift
```diff
@@ -489,37 +489,28 @@ class SpaceCardDetails: CardDetails, AccessingManagerProvider, ThumbnailModel {
...
-        spaceRef ?? manager.get(for: id)
-    }
-
     var title: String {
-        space?.displayTitle ?? ""
-    }
-
-    private var spaceRef: Space? = nil
-
-    init(space: Space, manager: SpaceStore) {
```
yozuysal, line 514: is there any way to keep using this on the side?

if you look below, we do a lot of manager.getAll().map {SpacevardDetail(space.....)} right now. that already is a loop over all spaces. If we pass id and also look for the space that has the id for each of these, then it is n^2 for each of these SpaceCardDetails generations....

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -509,7 +504,15 @@ class SpaceCardModel: CardModel {
     private var recommendationSubscription: AnyCancellable? = nil
     private var editingSubscription: AnyCancellable? = nil
     private var detailsSubscriptions: Set<AnyCancellable> = Set()
-    private var spaceNeedsRefresh: String? = nil
+    private var spaceNeedsRefresh: String? = nil {
+        didSet {
+            if let id = spaceNeedsRefresh {
+                manager.refreshSpace(spaceID: id)
```
yozuysal, line 510: hmm, why did we need this one? (I am guessing it is for getting pull to refresh working)

So basically, the reason we do this is, after any mutation (edit title, description, thumbnail, reorder) inside a single space, if we immediately refresh, then the user might do some other action immediately and we may want to refresh again and so on. So to avoid all these cascaded refreshes, we just record which space we need to refresh and only after detailedSpace is set to nil (meaning the detail is closed) we actually use spaceNeedsRefresh to call refresh.

Is there any way we can keep that logic and not incur a refresh every time it is set here?

yozuysal (changes requested):
two main questions. most of the changes look ok to me

Client/Frontend/Browser/Card/CardDetails.swift
buracules: This was for community spaces. I mean i can check isFollowing or not instead of checking allDetails and allProfiles first. But the other ones was already a computed property space was there only for id actually they were using manager.get(id).

Client/Frontend/Browser/Card/CardModels.swift
buracules: I am not sure why i put this here. It was probably something with the edit item bug. I can put back to didSet of detailedSpace and test it again.

Client/Frontend/Browser/Card/CardDetails.swift
buracules: When i assigned this space to item actually it broke the system for following spaces. It has to be accesed with manager right now maybe we have to fix that.

https://github.com/neevaco/neeva-ios/pull/3443/files

buracules: Instead of spaceRef i just assigned to the item and didnt call `updateSpace()` only called `updateDetails()` 
`updateSpace()` will be called only when its a community space (because we dont have details stored) and when refresh occurs.

I am still testing 😊

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -481,6 +481,12 @@ class SpaceCardModel: CardModel {
                         getLogCounterAttributesForSpaces(details: detailedSpace!))
                 return
             }
+
+            if let id = spaceNeedsRefresh {
+                manager.refreshSpace(spaceID: id)
+                spaceNeedsRefresh = nil
+
```
yozuysal, line 488: extra line

yozuysal (approved):

PR #3554, opened 2022-05-10
Cookie Cutter Logging

evancrow:
Adds logging for Cookie Cutter... just need to look into the spotty working of the engine and then it's release ready!

Issue: #3282


Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
```diff
@@ -57,6 +59,17 @@ class CookieCutterHelper: TabContentScript {
                 }
             case .increaseCounter:
                 cookieCutterModel.cookiesBlocked += 1
+            case .logProvideer:
```
chung-lai, line 62: nit: `logProvider` instead of `logProideer`

Client/Logging/LogConfig.swift
```diff
@@ -687,6 +691,9 @@ public struct LogConfig {
         case .PersonalSign: return .Web3
         case .TransactionSuccessful: return .Web3
         case .TransactionAttempted: return .Web3
+
+        // MARK: Cookie Cutter
+        case .CookieNoticeHandled: return .CookieCutter
```
chung-lai, line 696: couple things about this. We may want to enable this by default (add to `featureFlagEnabled` list); otherwise, we won't get data until the user is signed in and restart the app, hmm...we may want to revisit the list at some point as I felt we have been adding to it for most new categories.

Also, we should consider adding this to `shouldAddSessionID` in LogConfig as well so we can measure things like how many cookie popup we blocked. the cookie pop up does not seem sensitive data so we should be fine to add session id. @darinwf fyi

chung-lai (approved):
LGTM

PR #3566, opened 2022-05-11
Spaces preview mode; show login required sheet when clicking add comment

### LINKED ISSUE #3555: Spaces detail view "Add" comment button does nothing when user is signed out
We should either remove this button or turn it into an account creation promo. Ideally we'd do the latter once we have a design for that.

@jedireza @yozuysal FYI

jedireza:
Resolves #3555. We can leave as a WIP until we nail down a design and text copy.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

https://user-images.githubusercontent.com/979929/167743578-aa1f3b91-b51a-4504-ac8a-83589808702b.mp4

## Issues addressed
- https://github.com/neevaco/neeva-ios/issues/3555


Client/Frontend/Browser/Spaces/SpacesLoginRequiredView.swift
```diff
@@ -0,0 +1,40 @@
...
+
+struct SpacesLoginRequiredView: View {
+    @Environment(\.onSigninOrJoinNeeva) var onSigninOrJoinNeeva
+
+    var body: some View {
+        VStack {
+            Spacer()
+
+            Text("Please sign in")
+                .font(.title)
```
yozuysal, line 15: we have a handy `.withFont` modifier you can use for all text here. For the button below you don't need to set anything around font (weight, size etc) since the buttonStyle should take care of it.

yozuysal: A thought for later: It is good that we have this forked Spaces specific screen. We should probably think of an upsell for spaces here as a follow up.

Client/Frontend/Browser/Spaces/SpaceComments.swift
```diff
@@ -61,7 +61,11 @@ struct SpaceCommentsView: View {
                     .foregroundColor(.label)
                 Spacer()
                 Button(action: {
-                    model.addingComment = true
+                    if let _ = NeevaUserInfo.shared.id {
+                        model.addingComment = true
+                    } else {
+                        SceneDelegate.getBVC(for: nil).showSpacesLoginRequiredSheet()
```
yozuysal, line 67: FYI, this might break on iPads.. not sure if we have the BVC handy anywhere close unfortunately...

darinwf:
Thanks for jumping on this @jedireza !

darinwf (commented):
Did you consider using the same ErrorView that we have for Add-to-spaces as the placeholder?

jedireza:
> Did you consider using the same `ErrorView` (with the "this page is a little shy" UI) that we have for Add-to-spaces as the placeholder? We could then fix them all up in one pass by just updating the copy on that screen.

Yes. It’s coupled to having a graphql error. Then I thought we may want a different design/CTA. I’m  happy to adjust though.

darinwf:
> Yes. It’s coupled to having a graphql error. Then I thought we may want a different design/CTA. I’m happy to adjust though.

Ah, I hadn't considered the coupling to a GraphQL error. I thought it might be cheap to reuse the existing error view and that it would be a better starting point. Maybe still worth considering?

jedireza:
> I thought it might be cheap to reuse the existing error view and that it would be a better starting point. Maybe still worth considering?

Yep I had the same thought. Then when I saw the content of the message, I thought the copy "this page is a little shy" and "... view this page" didn't fit well for adding a comment.

image

Client/Frontend/Browser/Spaces/SpacesLoginRequiredView.swift
jedireza: Thanks for the note about `.withFont` and the `.neeva` button style.

Client/Frontend/Browser/Spaces/SpaceComments.swift
yozuysal: yep. thanks for checking!

darinwf:
> > I thought it might be cheap to reuse the existing error view and that it would be a better starting point. Maybe still worth considering?
> 
> Yep I had the same thought. Then when I saw the content of the message, I thought the copy "this page is a little shy" and "... view this page" didn't fit well for adding a comment.

Yeah, add-to-spaces sheet also has the same issue... it isn't a page either, but at the same time, that isn't something users will know about either UI surfaces. I don't think we are using "page" to mean a web page in other words.

It's okay to proceed with the current placeholder error page that you have. Just wanted to clarify. Think we need to invest in better error views for both of these.

darinwf (approved):

Client/Logging/LogConfig.swift
```diff
@@ -437,12 +438,17 @@ public struct LogConfig {
...
+            category == .FirstRun
             || category == .Stability
             || category == .PromoCard
             || category == .Web3
             || category == .Notification
             || category == .Cheatsheet
+
+        let validInteraction = path == .SpacesLoginRequired
+
+        return validCategory || validInteraction
```
jedireza, line 451: 📟  @chung-lai I think this is what we chatted about. Loog good?

chung-lai: LGTM! thanks!

hwiatneeva:
@jedireza are we looking for better error message design here?

jedireza:
@hwiatneeva we think it would help yes. With the right messaging it could also help promote the spaces feature. We merged since this as-is since it's better than broken.

hwiatneeva:
> @hwiatneeva we think it would help yes. With the right messaging it could also help promote the spaces feature. We merged since this as-is since it's better than broken.

TY for merging! That frees up dependency. //  I'll follow up with a discussion mock in the coming week.

PR #3573, opened 2022-05-11
Clean up - Black Friday Part

yasminyin96:
* Clean up -  Black Friday Part

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


yozuysal (approved):
thanks for the clean up!

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -172,33 +174,6 @@ class ZeroQueryModel: ObservableObject {
...
-                    )
-                },
-                onClose: {
-                    ClientLogger.shared.logCounter(
-                        .CloseBlackFridayPromo)
-                    Defaults[.seenBlackFridayFollowPromo] = true
-                    self.promoCard = nil
-                })
-        } else if promoCardTypeArm == .previewSignUp && shouldShowDefaultBrowserPromoCard() {
-            promoCard = defaultPromoCard()
```
chung-lai, line 201: We should not remove this `else if promoCardTypeArm == .previewSignUp && shouldShowDefaultBrowserPromoCard() {` part. We still show that in the app.

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -120,7 +120,9 @@ class ZeroQueryModel: ObservableObject {
         // TODO: refactor the promo card triggering logic to make it easier to customize
         let promoCardTypeArm = NeevaExperiment.arm(for: .promoCardTypeAfterFirstRun)
 
-        if (promoCardTypeArm == .control || promoCardTypeArm == nil)
+        if (promoCardTypeArm == .control
+            || promoCardTypeArm == nil
+            || promoCardTypeArm == .previewSignUp)
```
chung-lai, line 125: how does this extra condition `promoCardTypeArm == .previewSignUp` got added?

yasminyin96: The original is:
if (promoCardTypeArm == .control || promoCardTypeArm == nil) && shouldShowDefaultBrowserPromoCard()) {
    promoCard = defaultPromoCard()
} else if ...
...
else if (promoCardTypeArm == .previewSignUp && shouldShowDefaultBrowserPromoCard()) {
   promoCard = defaultPromoCard()
}

Therefore, I combine them into one if statement as
if (promoCardTypeArm == .control || promoCardTypeArm == nil || promoCardTypeArm == .previewSignUp) && shouldShowDefaultBrowserPromoCard()) {}

chung-lai: @EdwardJTL, yes, Yasmin is working on cleaning up the promo card logic by moving it into ViewModel as it's not very readable at the moment with a bunch of `if/else`

@yasminyin96, don't think we can simply combine it. The reason we had that in two different places between we want to maintain the ordering of the condition check :(

Can't wait for some clean up here!

yasminyin96: > @EdwardJTL, yes, Yasmin is working on cleaning up the promo card logic by moving it into ViewModel as it's not very readable at the moment with a bunch of `if/else`
> 
> @yasminyin96, don't think we can simply combine it. The reason we had that in two different places between we want to maintain the ordering of the condition check :(
> 
> Can't wait for some clean up here!

@chung-lai  Make sense, then could I just remove the black friday part first? And I'll look into if/else part later.

EdwardJTL: ah right, i didnt see the other checks in between

PR #3580, opened 2022-05-11
Handle first navigation for native navigation as well

jedireza:
Before we launch the new zero query updates we wanted to ensure that `Defaults[.didFirstNavigation]` was working for native spaces as well. Before this PR, we were not bumping `.didFirstNavigation` for native spaces navigation.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Tested in the simulator.

I can tell we were not bumping `.didFirstNavigation` because the bottom bar is hidden until a user has done a navigation.

| Before | After |
| --- | --- |
|  Kapture 2022-05-11 at 15 49 47 | Kapture 2022-05-11 at 15 51 48 |

## Issues addressed
N/A


Client/Application/NavigationRouter.swift
```diff
@@ -101,6 +101,8 @@ enum NavigationPath {
     }
 
     static func handle(nav: NavigationPath, with bvc: BrowserViewController) {
+        bvc.handleFirstNavigation()
```
jedireza, line 104: I'm not sure if we want this up here for any navigation, so I'm happy to move it into specific cases below if folks thing that's best.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1090,6 +1090,13 @@ class BrowserViewController: UIViewController, ModalPresenter {
         case finishedNavigation
     }
 
+    func handleFirstNavigation() {
```
chung-lai, line 1093: would be great if we can pass in a type attribute to tell us the first navigation is triggered by a space or link.

you can add a new attribute in `LogConfig` [here](https://github.com/neevaco/neeva-ios/blob/main/Client/Logging/LogConfig.swift#L703). And the value for the attribute can be `space` or `link` (we should be able to distinguish a link between direct url navigation vs preview suggestion from another event)

Client/Application/NavigationRouter.swift
chung-lai: May be we just add it to `space` case?

darinwf:
@jedireza and I discussed adding a new log event for just the spaces case since setting `didFirstNavigation` has other side-effects.

chung-lai:
> @jedireza and I discussed adding a new log event for just the spaces case since setting `didFirstNavigation` has other side-effects.

Ah, cool, just realized this PR was closed :)

PR #3586, opened 2022-05-12
Add log interaction for recommended space view

jedireza:
Adding a log interaction for viewing a recommended space. This is an alternative to bumping `Defaults[.didFirstNavigation]` (see https://github.com/neevaco/neeva-ios/pull/3580)

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

N/A

## Issues addressed
N/A


Client/Frontend/ZeroQuery/RecommendedSpacesView.swift
```diff
@@ -50,6 +50,8 @@ struct RecommendedSpacesView: View {
                                 for: url)
                             viewModel.bvc.hideZeroQuery()
                         }
+
+                        ClientLogger.shared.logCounter(.SpacesRecommendedDetailUIVisited)
```
jedireza, line 54: Should we only put this is the top branch above? IIUC the top branch will trigger a native app space and the bottom one will open the space in a browser tab.

darinwf: I wonder in what cases we would hit the "else" case.

darinwf (approved):

PR #3593, opened 2022-05-13
time based switcher history view

chassu:
This is still WIP and there are some issues I'm aware of, but I think some parts of it can be already reviewed to save some time.

Some issues that I'm aware of:

- ~~Card transition animation is missing the first time when a tab is brought back from archives. (select a tab in archives -> webview shows up -> click switcher button)~~
- ~~There should be a more granular way for the switcher to update tab groups when a tab is brought back from archives. (I temporarily change `updateRows` to `onDataUpdated` as a proof of concept). This is different from `buildRows` which is called when a tab is selected. By introducing archived tabs, there's a chance a tab group can be formed simply by selecting tabs from archives.~~
- ~~Clear All Archived Tabs doesn't do anything yet~~
- ~~ArchivedTabsPanelView refreshes on click (not ideal IMO)~~
- ~~Will add the UI to cover different sections based on `archivedTabsDuration`~~

An issue unrelated to this PR:

- I assume something about `FaviconView` is making the favicon changing values back and forth. Can be seen in the ArchivedTabsPanelView and it has existed in our tab switcher too, especially when a webview is loading.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/87332917/168216076-c3da13ab-ef2f-48d5-a39b-60722d9596dd.mp4


Client/Frontend/Browser/Card/ArchivedTabsPanelModel.swift
```diff
@@ -0,0 +1,50 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Shared
+
+enum ArchivedTabTimeSection: String, CaseIterable {
```
evancrow, line 7: Should be able to reuse the time enum from `HistoryModel` here instead of creating a new one.

Client/Frontend/Browser/Card/ArchivedTabsPanelModel.swift
```diff
@@ -0,0 +1,50 @@
...
+    var sites: [ArchivedTabTimeSection: [Tab]] = [:]
+
+    func itemsForSection(section: ArchivedTabTimeSection) -> [Tab] {
+        return sites[section] ?? []
+    }
+}
+
+class ArchivedTabsPanelModel: ObservableObject {
+    let tabManager: TabManager
+    var representativeTabsInAMonth: [Tab] = []
```
evancrow, line 22: Shouldn't need these if you have `groupedSites` right?

Client/Frontend/Browser/Card/ArchivedTabsPanelView.swift
```diff
@@ -0,0 +1,106 @@
...
+            VStack(spacing: 0) {
+                Button(action: {}) {
+                    HStack(spacing: 0) {
+                        Text("Archive tabs")
+                            .withFont(.bodyLarge)
+                            .foregroundColor(.label)
+                            .fixedSize(horizontal: false, vertical: true)
+                            .padding(.vertical, 10)
+                        Spacer()
+                        Group {
```
evancrow, line 38: What is the point of this empty `Group`? Is it still WIP?

Client/Frontend/Browser/Card/ArchivedTabsView.swift
```diff
@@ -23,14 +24,20 @@ struct ArchivedTabsView: View {
         }
     }
 
+    var bvc: BrowserViewController {
```
evancrow, line 27: I don't think you should plumb down `tabModel` to get the `BVC`. Would be better to create a callback in `ArchivedTabsView` that is set in a place that doesn't need the plumbing.

Client/Frontend/Browser/Card/ArchivedTabsPanelViewController.swift
```diff
@@ -0,0 +1,36 @@
...
+                \.onOpenURL,
+                {
+                    bvc.tabManager.createOrSwitchToTab(for: $0)
+                    bvc.browserModel.hideGridWithNoAnimation()
+                }
+            )
+            .environment(
+                \.selectionCompletion, { bvc.browserModel.hideGridWithNoAnimation() }
+            )
+            .environmentObject(bvc.browserModel)
```
evancrow, line 26: I don't think these need to be passed down. This was for the `OverlayView` in settings which isn't being used here.

Client/Frontend/Browser/Card/TabListView.swift
```diff
@@ -0,0 +1,106 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Defaults
+import Shared
+import SwiftUI
+
+struct ButtonView: View {
```
evancrow, line 9: Should give this a more descriptive name.

evancrow (commented):
Coming along well! Think a lot of next steps are just code cleanup and refactoring.

Client/Frontend/Browser/Card/ArchivedTabsPanelModel.swift
chassu: I changed it to reuse the enum from history, but some enum cases won't be used. Can you take a look and let me know if it looks fine? I originally separated them because I thought it would be awkward to have unused enum cases.

evancrow: Think you're right that it doesn't make sense to have an unused enum. Guess the way you had it before is fine.

Client/Frontend/Browser/Card/ArchivedTabsPanelModel.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    init(tabManager: TabManager) {
+        self.tabManager = tabManager
+
+        archivedTabsDurationSubscription = _archivedTabsDuration.publisher.sink {
+            [weak self] _ in
+            self?.loadData()
+            self?.objectWillChange.send()
+        }
+    }
```
chassu, line 52: I was hoping `ArchivedTabsPanelView` will pick up the change right when a user selects a new setting inside `ArchivedTabSettings.` But this doesn't seem to do anything. @evancrow  Thoughts?

The purpose of this is to ensure after the user selects a new settings, hit the back button and they will see an updated `ArchivedTabsPanelView` because the filtering logic has changed.

chassu: After an "optimization" I thought I made - avoid calling `onDataUpdated` when active tabs are selected, I realized that the tab switcher doesn't respond to `archivedTabsDuration` settings changes, and the user won't see the correct tabs unless they add/remove a tab. I have an idea for this now and am going to try it out. But I want to point out this issue.

chassu: Resolved by changing '.' to '_' for the user default key... 🙄

chassu:
won't make anymore functionality changes on this PR, will work on cleanups for now.

Client/Frontend/Browser/Card/ArchivedTabsListView.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    @Environment(\.selectionCompletion) private var selectionCompletion: () -> Void
+
+    var body: some View {
+        Button {
+            tabManager.selectTabFromArchive(tab)
+            selectionCompletion()
+        } label: {
+            HStack {
+                FaviconView(forSiteUrl: tab.url ?? "")
```
chassu, line 23: There could be a migration issue here, I just experienced a crash caused by `tab.url` used to not being stored, and the fallback "" will lead to crash when the `FaviconView` takes in an empty string as a parameter.

chassu, line 23: There could be a migration issue here, I just experienced a crash caused by `tab.url` used to not being stored, and the fallback "" will lead to crash when the `FaviconView` takes in an empty string as a parameter.

evancrow: Unwrap the `URL` before the `FaviconView` instead of passing an empty `String`?

Client/Frontend/Browser/Card/ArchivedTabsListView.swift
```diff
@@ -0,0 +1,106 @@
...
+            dict[groupName.key] = false
+        }
+    }
+
+    var body: some View {
+        LazyVStack(spacing: 8) {
+            var processed = getProcessedFromTabGroupDict()
+            ForEach(tabs, id: \.self) { tab in
+                if processed[tab.rootUUID] != nil {
+                    if processed[tab.rootUUID] == false {
```
evancrow, line 68: Can you try to reduce the number of nested `if`s here?

Client/Frontend/Browser/Card/ArchivedTabsPanelView.swift
```diff
@@ -0,0 +1,134 @@
...
+                    model.clearArchivedTabs()
+                }) {
+                    HStack(spacing: 0) {
+                        Text("Clear All Archived Tabs")
+                            .withFont(.bodyLarge)
+                            .foregroundColor(.red)
+                            .fixedSize(horizontal: false, vertical: true)
+                            .padding(.vertical, 10)
+                        Spacer()
+                        Group {
```
evancrow, line 75: Is this group supposed to be empty?

Client/Frontend/Browser/Card/ArchivedTabsPanelView.swift
```diff
@@ -0,0 +1,134 @@
...
+                } label: {
+                    EmptyView()
+                }
+
+                Color.secondarySystemFill
+                    .frame(height: 8)
+            }
+
+            // Archived tabs
+            LazyVStack(spacing: 0, pinnedViews: .sectionHeaders) {
```
evancrow, line 94: You shouldn't need to embed `LazyVStack`s inside a `LazyVStack`. I think you can keep the outermost one, and then make `ArchivedTabsListView` a `VStack`. Also maybe `ArchivedTabsListView` should be renamed to `ArchivedTabsListSectionView`?

chassu: Did the inner `LazyVStack` because I thought about an archived tab group that has 100 tabs and wonder if we want to do lazy loading for those.

chassu: ^changed to what you recommended

Client/Frontend/Browser/Card/ArchivedTabsListView.swift
evancrow: Yeah, `AnyView` isn't great. If we really wanted to refactor this, could make the code inside the `if` its own `View`, but I think this is fine for now.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -79,8 +89,11 @@ class TabCardModel: CardModel {
         self.objectWillChange.send()
     }
 
-    func updateRowsIfNeeded(force: Bool = false) {
-        if needsUpdateRows || force {
+    func updateIfNeeded(force: Bool = false) {
+        if needsUpdateGroup {
+            needsUpdateGroup = false
+            onDataUpdated()
+        } else if needsUpdateRows || force {
```
chassu, line 96: maybe change `force` to `forceUpdateRows` here?

Client/Frontend/Browser/Card/ArchivedTabsPanelViewController.swift
```diff
@@ -0,0 +1,31 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import SwiftUI
+
+class ArchivedTabsPanelViewController: UIHostingController<AnyView> {
+    init(bvc: BrowserViewController) {
```
darinwf, line 8: How about just passing `BrowserModel` here since it also has a reference to the `TabManager`? That way we avoid one more dependency on BVC.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -83,6 +84,44 @@ class TabManager: NSObject {
...
     }
 
+    var activeTabs: [Tab] {
+        return incognitoTabs
+            + normalTabs.filter {
+                return !$0.isArchived()
+            }
+    }
+
+    var activeTabGroups: [String: TabGroup] {
```
darinwf, line 94: I wonder if this will be an expensive to compute property. I wonder if we should compute and store the results to help speed things up.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -83,6 +84,44 @@ class TabManager: NSObject {
...
+        return normalTabs.filter {
+            return $0.isArchived()
+        }
+    }
+
+    var archivedTabGroups: [String: TabGroup] {
+        // In archivedTabsPanelView, there's special UI treatments for a child tab
+        // even if it's the only arcvhied tab in a group. Those tabs won't be filtered
+        // out below (see activeTabGroups for comparison).
+        return getAll()
```
darinwf, line 115: Ditto

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -243,6 +282,10 @@ class TabManager: NSObject {
         selectedTab.lastExecutedTime = Date.nowMilliseconds()
         selectedTab.applyTheme()
 
+        if updateGroup {
+            updateGroupPublisher.send()
```
darinwf, line 286: this should be only within the `notify` block below. we want to avoid all notifications if `notify` is false.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -502,6 +545,18 @@ class TabManager: NSObject {
         }
         tabGroupDict = temp
     }
+
+    func clearArchivedTabs() {
+        let tabsToKeep = self.tabs.filter { !archivedTabs.contains($0) }
```
darinwf, line 550: should this just be using `removeTabs` passing the set of archived tabs? Why re-invent another variant of the `removeTabs` function?

darinwf (commented):
Looking pretty good. Just some flags.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: This computed property is only called once during `onDataUpdated` and the information is technically stored inside allTabGroupDetails. I feel like this is okay..

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: I made `ArchivedTabsPanelModel` store `archivedTabGroups` during init, which is the only time this computed property ever get evaluated. Are we good with this?

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: `removeTabs` sends notifications such as `updateTabGroupsAndSendNotifications(notify: true)` and `sendSelectTabNotifications(previous: previous)` which I think we want to avoid when clearing archived tabs. I've switched to `removeTabs` but had to add extra logic to skip sending signals. Not sure if it's better.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: Well, if the user spams the ArchivedTabsPanelModel, we'll keep computing the property over and over again, which is probably not ideal, but one caveat: some tabs could get archived while the user spams it, so I think the question to ask here is "should we recalculate the archivedTabs/archivedTabsGroups every time a user launch the ArchivedTabsPanelView"? It might be good to get the most updated data, but I have a feeling that the chance of some tabs getting archived when the user is interacting with the ArchivedTabsPanelView is very very low.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
evancrow: Could make this a regular var and then update it in `updateTabGroups` or whenever tabs are archived/unarchived.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -83,6 +84,44 @@ class TabManager: NSObject {
...
+    var activeTabs: [Tab] {
+        return incognitoTabs
+            + normalTabs.filter {
+                return !$0.isArchived()
+            }
+    }
+
+    var activeTabGroups: [String: TabGroup] {
+        return getAll()
+            .reduce(into: [String: [Tab]]()) { dict, tab in
```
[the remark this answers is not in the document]
chassu: I'm not sure how to refactor this function (reduce -> filter -> reduce). Any tips?

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -438,12 +457,24 @@ class TabManager: NSObject {
...
             }
+
+        archivedTabGroups = getAll()
+            .reduce(into: [String: [Tab]]()) { dict, tab in
+                if tabGroupDict[tab.rootUUID] != nil && tab.isArchived() {
+                    dict[tab.rootUUID, default: []].append(tab)
+                }
+            }.reduce(into: [String: TabGroup]()) { dict, element in
+                dict[element.key] = TabGroup(children: element.value, id: element.key)
+            }
```
chassu, line 476: As requested, I changed `activeTabGroups` and `archivedTabGroups` into stored property. They are updated every time `updateTabGroupsAndSendNotifications` happens (could happen when 1. add tabs 2. close tabs 3. bring tabs back from archives). Compared with defining them as computed variables, this saves some computing from launching the `ArchivedTabsPanelView`, but adds compute to  `updateTabGroupsAndSendNotifications` since now it's computing activeTabGroups, archivedTabGroups, and `cleanUpTabGroupNames()` also becomes a bit more expensive to compute.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -486,6 +517,13 @@ class TabManager: NSObject {
 
     func cleanUpTabGroupNames() {
         // Write tab group name into dictionary
+        let tabGroups = getAll()
+            .reduce(into: [String: [Tab]]()) { dict, tab in
+                dict[tab.rootUUID, default: []].append(tab)
+            }.filter { $0.value.count > 1 }.reduce(into: [String: TabGroup]()) { dict, element in
+                dict[element.key] = TabGroup(children: element.value, id: element.key)
+            }
```
chassu, line 525: ArchivedTabGroups or ActiveTabGroups can't be used here, still need the old definition for all tab groups.

Client/Frontend/Browser/BrowserModel.swift
```diff
@@ -41,17 +41,14 @@ class BrowserModel: ObservableObject {
...
         if !gridModel.tabCardModel.getAllDetails(matchingIncognitoState: incognitoModel.isIncognito)
             .contains(where: \.isSelected)
         {
             showGridWithNoAnimation()
         } else {
-            if FeatureFlag[.enableTimeBasedSwitcher] {
-                gridModel.tabCardModel.updateRowsIfNeeded()
-            }
-
-            overlayManager.hideCurrentOverlay(ofPriority: .modal)
```
darinwf, line 54: This looks like a bad merge.

Client/Frontend/Browser/Card/ArchivedTabsListSectionView.swift
```diff
@@ -0,0 +1,118 @@
...
+            }
+            .padding(.trailing, -6)
+            .padding(.horizontal, GroupedCellUX.padding)
+            .padding(.vertical, 6)
+            .frame(minHeight: GroupedCellUX.minCellHeight)
+        }
+    }
+}
+
+struct ArchivedTabsListSectionVIew: View {
```
darinwf, line 48: nit: `VIew` -> `View`

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -486,19 +525,16 @@ class TabManager: NSObject {
...
-        tabGroups.forEach { group in
-            let id = group.key
-            if tabGroupDict[id] == nil {
-                tabGroupDict[id] = group.value.displayTitle
+        let tabGroups = getAll()
+            .reduce(into: [String: [Tab]]()) { dict, tab in
+                dict[tab.rootUUID, default: []].append(tab)
+            }.filter { $0.value.count > 1 }.reduce(into: [String: TabGroup]()) { dict, element in
+                dict[element.key] = TabGroup(children: element.value, id: element.key)
             }
```
chassu, line 533: @darinwf I know we talked about the change to run two loops with `activeTabGroups` and `archiveTabGroups`, but this leads to inconsistency in tab group titles as the group title is first child's title by default. If child tabs span across active and archived sections, it could show different titles even if the tabs belong to the same group.

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -676,28 +680,58 @@ class Tab: NSObject, ObservableObject {
...
         let lastExecutedTime = lastExecutedTime ?? Date.nowMilliseconds()
         let minusOneDayToCurrentDate =
-            FeatureFlag[.demoteAfter15secondsTimeBasedSwitcher]
+            FeatureFlag[.shortenTimeThresholdForArchivingTabs]
             ? Calendar.current.date(
                 byAdding: .second, value: -15, to: Date())
             : Calendar.current.date(
                 byAdding: .day, value: -1, to: Date())
-        guard let startOfOneDayAgo = minusOneDayToCurrentDate else {
+        let minusOneWeekToCurrentDate =
```
darinwf, line 688: Some of this code looks like it is duplicated into TabGroup. We should avoid duplicating this code.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -83,6 +86,13 @@ class TabManager: NSObject {
         return tabs.filter { $0.isIncognito }
     }
 
+    var activeTabs: [Tab] {
```
darinwf, line 89: Are you still planning on making this a stored property?

Client/Frontend/Browser/Tabs/TabGroup.swift
```diff
@@ -9,6 +9,15 @@ import Shared
...
     var children: [Tab]
     var id: String
+    var lastExecutedTime: Timestamp
+
+    init(children: [Tab], id: String) {
+        self.children = children
+        self.id = id
+        self.lastExecutedTime =
+            children.map { $0.lastExecutedTime ?? Date.nowMilliseconds() }.max()
+            ?? Date.nowMilliseconds()
```
chassu, line 19: https://github.com/neevaco/neeva-ios/issues/3739 is causing same tab group to show up in different sections, so I thought about creating lastExecutedTime for TabGroup...

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -486,19 +525,16 @@ class TabManager: NSObject {
 
     func cleanUpTabGroupNames() {
         // Write tab group name into dictionary
-        tabGroups.forEach { group in
-            let id = group.key
-            if tabGroupDict[id] == nil {
-                tabGroupDict[id] = group.value.displayTitle
+        let tabGroups = getAll()
```
darinwf, line 528: This looks expensive. Why didn't you want to just enumerate `activeTabGroups` and `archivedTabGroups` to get the set of all rootUUID -> title mappings?

Client/Frontend/Browser/Tabs/TabGroup.swift
```diff
@@ -9,6 +9,15 @@ import Shared
 struct TabGroup {
     var children: [Tab]
     var id: String
+    var lastExecutedTime: Timestamp
+
+    init(children: [Tab], id: String) {
+        self.children = children
+        self.id = id
+        self.lastExecutedTime =
```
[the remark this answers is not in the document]
chassu: I think you're right. Part of my change is to modify `TabGroup.lastExecutedTime` on `TabManager.selectTab`. I think turning it into a computed property makes sense.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
darinwf: I see. Hmm...

Is the goal of this function merely about garbage collecting elements of `tabGroupDict` that are no longer needed after a tab group is removed, or is its purpose more than that?

Perhaps an alternative would be to add to `tabGroupDict` upon forming a tab group for the first time.

Then this function would only need to be concerned with pruning tab group names that no longer exists.

OK for this to be a follow-up PR, I think.

The current code deserves a comment explaining why it is necessary to compute the merged set of tab groups.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: I was thinking about a case where child tabs of a tab group span across activeTabGroups and archivedTabGroups: the title for sides should be consistent. `displayTitle` simply gets the title of the first child, which means the second loop could overwrite the first title. Also, I noticed that the change we discussed will make custom tab group title no longer persistent because of `temp[group.key] = group.value.displayTitle`. I think we might actually need to keep the nil check (to not overwrite custom tab group names).

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
chassu: I think another goal of the function is to write newly created tab group's title into the dictionary. I'll undo the changes on this function and make some comments. We can figure out a follow-up PR to handle this.

Client/Frontend/Browser/Tabs/Tab.swift
chassu: I tried to make a public func in Shared/TimeConstants.swift, but got complaints because I"m using FeatureFlag in the function. I define the public func inside Tab.swift. What do you think?

Client/Frontend/Browser/Card/ArchivedTabsListSectionView.swift
```diff
@@ -0,0 +1,118 @@
...
+    @EnvironmentObject var model: ArchivedTabsPanelModel
+    @Default(.tabGroupNames) private var tabGroupDict: [String: String]
+
+    private let padding: CGFloat = 4
+    let tabManager: TabManager
+    let tabs: [Tab]
+    let section: ArchivedTabTimeSection
+
+    func getProcessedFromTabGroupDict() -> [String: Bool] {
+        return tabGroupDict.reduce(into: [String: Bool]()) { dict, groupName in
```
darinwf, line 60: This looks potentially expensive. Can some of the processing be moved to a ViewModel? We have learned that we should keep the `body` of our Views computationally as simple as possible.

Client/Frontend/Browser/Card/ArchivedTabsListSectionView.swift
```diff
@@ -0,0 +1,118 @@
...
+                            .children.filter { child in
+                                switch section {
+                                case .lastMonth:
+                                    return child.wasLastExecuted(.lastMonth)
+                                case .overAMonth:
+                                    return child.wasLastExecuted(.overAMonth)
+                                }
+                            }
+                        {
+                            LazyVStack(spacing: 0) {
```
darinwf, line 84: Nested LazyVStacks make me worry.

Client/Frontend/Browser/Card/ArchivedTabsListSectionView.swift
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+
+    var body: some View {
+        LazyVStack(spacing: 8) {
+            var processed = getProcessedFromTabGroupDict()
+            ForEach(tabs, id: \.self) { tab in
+                if processed[tab.rootUUID] != nil {
+                    if processed[tab.rootUUID] == false {
+                        // A tab group can have child tabs spread across different archived sections.
```
darinwf, line 71: This is begging for ViewModel treatment instead of doing this work in the `body`.

Client/Frontend/Browser/Card/ArchivedTabsPanelModel.swift
```diff
@@ -0,0 +1,63 @@
...
+
+    init(tabManager: TabManager) {
+        self.tabManager = tabManager
+        self.archivedTabs = tabManager.archivedTabs
+        self.archivedTabGroups = tabManager.archivedTabGroups
+
+        archivedTabsDurationSubscription = _archivedTabsDuration.publisher.dropFirst().sink {
+            [weak self] _ in
+            self?.tabManager.updateTabGroupsAndSendNotifications(notify: false)
```
darinwf, line 58: This feels backwards. Why is ArchivedTabsPanelModel responsible for calling `updateTabGroupsAndSendNotifications`? Shouldn't `TabManager` be the one to observe `archivedTabsDuration` itself and drive the updates?

The flow of updating should be more one directional. TabManager observes its inputs and generates notifications. Downstream from that the ArchivedTabsPanelModel would notice when TabManager changes and then update its data accordingly.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -46,6 +49,7 @@ class TabManager: NSObject {
     /// Publisher used to observe changes to the `selectedTab.webView`.
     /// Will also update if the `WebView` is set to nil.
     private(set) var selectedTabWebViewPublisher = CurrentValueSubject<WKWebView?, Never>(nil)
+    private(set) var updateGroupPublisher = PassthroughSubject<Void, Never>()
```
darinwf, line 52: Please add a comment explaining this new publisher.

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -77,8 +87,12 @@ class TabCardModel: CardModel {
         self.objectWillChange.send()
     }
 
-    func updateRowsIfNeeded(force: Bool = false) {
-        if needsUpdateRows || force {
+    func updateIfNeeded(forceUpdateRows: Bool = false) {
+        if needsUpdateGroup {
+            needsUpdateGroup = false
+            manager.updateTabGroupsAndSendNotifications(notify: false)
```
darinwf, line 93: I think we should avoid calling updateTabGroupsAndSendNotifications from outside of TabManager. Can it be the one that knows when it should do this work?

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -77,8 +87,12 @@ class TabCardModel: CardModel {
...
+            onDataUpdated()
+        } else if needsUpdateRows || forceUpdateRows {
```
darinwf, line 95: I'm confused by the `else` here. Why are these branches mutually exclusive? What if both `needsUpdateGroup` and `needsUpdateRows` are true?

darinwf: Oh, I get it... `onDataUpdated` also calls `updateRows`. This is not very clear. Should you also clear `needsUpdateRows` before calling `onDataUpdated`?

Client/Frontend/Browser/Card/BrowserPrimitive.swift
```diff
@@ -129,6 +129,10 @@ extension TabManager: ClosingManager, SelectingManager, AccessingManager {
         selectTab(tab, notify: true)
     }
 
+    func selectTabFromArchive(_ tab: Tab) {
```
darinwf, line 132: rather than introduce this special function that must be used to pull a tab out of the archive, why don't you just make `selectTab` work? i worry there could be other paths that might lead to selecting an archived tab. e.g., createOrSwitchToTab. it is going to get messy if we need to special case all calls to selectTab to check if the tab is archived. TabManager should just do this internally.

Can't selectTab just notice that it is selecting an archived tab and then trigger all of the tab group updating logic? Is the issue that you need to do that lazily from hideGridWithAnimation?

darinwf (commented):
This PR is complicated! I wish there were a way to break it down into simpler pieces.

darinwf (changes requested):
Requesting changes per above comments

Client/Frontend/Browser/Card/BrowserPrimitive.swift
chassu: Made the change to make selectTab work. Had to add an argument to other places too because of the SelectingManager protocol.

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -675,29 +679,19 @@ class Tab: NSObject, ObservableObject {
...
-                byAdding: .day, value: -7, to: Date())
-            guard let startOfLastWeek = minusOneWeekToCurrentDate else {
-                return true
-            }
-            return lastExecutedTime < Int64(startOfOneDayAgo.timeIntervalSince1970 * 1000)
-                && lastExecutedTime > Int64(startOfLastWeek.timeIntervalSince1970 * 1000)
+        return isLastExecutedTimeInTimeFilter(lastExecutedTime, byTime)
+    }
+
+    func isArchived() -> Bool {
```
darinwf, line 685: nit: Probably would be nicer as a computed property

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -244,6 +263,11 @@ class TabManager: NSObject {
         selectedTab.applyTheme()
 
         if notify {
+            if updateGroup {
```
darinwf, line 266: Rather than force the caller to know if this call to selectTab will alter tab groups, why not just have TabManager determine that? We can just check `tab.isArchived` right?

Client/Frontend/Browser/Card/BrowserPrimitive.swift
chassu: We need to update tab data when a tab is selected from archives because 1. they are not an active tab 2. they belong to a tab group. So the publisher is needed to make sure `CardModel.onDataUpdated` gets called so the CardGrid is up to date.

darinwf: Why not just have TabManager call `updateTabGroupsAndSendNotifications` passing `notify: true`? Is there some reason why calling `tabsUpdatedPublisher.send()` from within `selectTab` wouldn't work?

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -240,10 +256,15 @@ class TabManager: NSObject {
             updateWebViewForSelectedTab(notify: false)
         }
 
+        let isArchived = selectedTab.isArchived
         selectedTab.lastExecutedTime = Date.nowMilliseconds()
```
chassu, line 260: need to save `isArchived` ahead of time because updating `lastExecutedTime` could change the value we want to read.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -240,10 +256,15 @@ class TabManager: NSObject {
...
         selectedTab.applyTheme()
 
         if notify {
+            if isArchived {
+                // Tab data needs to be updated when a tab is brought back from archives.
+                updateTabGroupsAndSendNotifications(notify: true)
```
darinwf, line 266: Should we hoist this out of the `if notify` block so we run `updateTabGroupsAndSendNotifications` always and just vary the `notify` parameter. So like this:

```
if isArchived {
    updateTabGroupsAndSendNotifications(notify: notify)
}
if notify {
    ...
}
```

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -34,13 +34,15 @@ extension TabManager {
 
     func removeTabs(
         _ tabsToBeRemoved: [Tab], showToast: Bool = true,
-        updateSelectedTab: Bool = true
+        updateSelectedTab: Bool = true, fromArchive: Bool = false
```
darinwf, line 37: consider adding two parameters here instead that are more descriptive about the functions here:

`notify: Bool` and `dontAddToRecentlyClosed: Bool`

darinwf (approved):

PR #3597, opened 2022-05-13
Refactor PromoCard

yasminyin96:
- Refactor PromoCard

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -45,6 +45,14 @@ enum ZeroQueryTarget {
     static var defaultValue: ZeroQueryTarget = .existingOrNewTab
 }
 
+enum PromoCardDisplay: CaseIterable {
+    case shouldArmDefaultPromo
```
chung-lai, line 49: nit: either we prefix them with `shouldDisplay` or leave `should` out.

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -45,6 +45,14 @@ enum ZeroQueryTarget {
     static var defaultValue: ZeroQueryTarget = .existingOrNewTab
 }
 
+enum PromoCardDisplay: CaseIterable {
```
chung-lai, line 48: So we rely on the enum ordering to prioritize which card to show? Does allCases return the enum in order?

If we go down this path, we should put a comment and warning on top to make sure others know there is the ordering implication. Or we should have another array that store a list of enum which maintains the ordering that way which is more explicit.

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -97,110 +105,117 @@ class ZeroQueryModel: ObservableObject {
...
+            UserDefaults.standard.removeObject(forKey: "DidDismissDefaultBrowserCard")
+            Defaults[.didDismissDefaultBrowserCard] = true
+        }
+
+        // TODO: refactor the promo card triggering logic to make it easier to customize
+        for card in PromoCardDisplay.allCases {
+            if shouldDisplayPromoCard(promoCardDisplay: card) {
+                if card == .shouldSignUpOrWalletPromo {
+                    #if XYZ
+                        if Defaults[.cryptoPublicKey].isEmpty {
```
chung-lai, line 202: We cannot extract this logic into the shouldDisplay function?

chung-lai (commented):
Love the enum types introduction! it's much easier to maintain the ordering and display logic of each promo card. Thank you!

Left a few comments.

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
yasminyin96: I tried to put this into shouldDisplay with shouldSignUpPromo and shouldWalletPromo, but there are two errors:
1. #if cannot put into switch?
2. bvc.web3Model.showWalletPanel() not exist

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
yasminyin96: The current order is from shouldArmDefaultPromo -> shouldSignUpOrWalletPromo -> ... -> shouldDefaultPromo.

chung-lai: Right but i think it would be better if we maintain an array of enums and iterate through that array instead of doing `allCases`. That way, it's more explicit in the code where we define the ordering :)

yasminyin96: That makes sense. I'll add it.

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -97,85 +107,121 @@ class ZeroQueryModel: ObservableObject {
...
+        // allCases: Ordering shouldDisplayArmDefaultPromo -> shouldDisplayWalletPromo
+        // ->shouldDisplaySignUpPromo -> shouldDisplayReferralPromo
+        // -> shouldDisplaySignInPromo -> shouldDisplayDefaultPromo
+        for card in PromoCardDisplay.allCases {
+            if shouldDisplayPromoCard(promoCardDisplay: card) {
+                setDisplayPromoCard(promoCardDisplay: card)
+                break
+            } else {
+                promoCard = nil
+            }
```
EdwardJTL, line 224: a swifty way would be to write
```
if let card = PromoCardDisplay.allCases.first(where: {shouldDisplayPromoCard(promoCardDisplay:$0)}) {
setDisplayPromoCard(promoCardDisplay: card)
} else {
promoCard = nil
}
```
that also saves you from assigning the value to nil multiple times in the loop

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -97,85 +107,121 @@ class ZeroQueryModel: ObservableObject {
...
-            UserDefaults.standard.removeObject(forKey: "DidDismissDefaultBrowserCard")
-            Defaults[.didDismissDefaultBrowserCard] = true
+        if interaction == .OpenReferralPromo || interaction == .CloseReferralPromo {
+            attributes.append(ClientLogCounterAttribute(key: "source", value: "zero query"))
         }
+        ClientLogger.shared.logCounter(interaction, attributes: attributes)
+    }
 
-        // TODO: refactor the promo card triggering logic to make it easier to customize
+    func shouldDisplayPromoCard(promoCardDisplay: PromoCardDisplay) -> Bool {
```
EdwardJTL, line 118: these two function signatures could be more swifty. for example
`shouldDisplay(promoCard: PromoCardDisplay)` or
`shouldDisplayPromoCard(_ promoCard:  PromoCardDisplay)` 
to make the call sites look cleaner

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -45,6 +45,16 @@ enum ZeroQueryTarget {
     static var defaultValue: ZeroQueryTarget = .existingOrNewTab
 }
 
+// Priority Ordering from top to bottom
+enum PromoCardDisplay: CaseIterable {
```
EdwardJTL, line 49: can this enum be named `PromoCardType`, and each case just say`<Name>Promo`?

chung-lai: So this enum is meant for checking the condition whether to trigger the promo card. May be we can name it `PromoCardTrigger`?

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -45,6 +45,16 @@ enum ZeroQueryTarget {
     static var defaultValue: ZeroQueryTarget = .existingOrNewTab
 }
 
+// Priority Ordering from top to bottom
```
chung-lai, line 48: Thought we will be maintaining an enum array for making the ordering more explicit?

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -97,83 +115,113 @@ class ZeroQueryModel: ObservableObject {
...
-                        ClientLogger.shared.logCounter(
-                            .ClosePreviewSignUprPromo,
-                            attributes: EnvironmentHelper.shared.getAttributes()
-                        )
-                        self.promoCard = nil
-                        Defaults[.didDismissPreviewSignUpCard] = true
-                    }
-                } else {
-                    promoCard = nil
+            #endif
```
EdwardJTL, line 143: interesting that this compiles. if `XYZ` is false, the compiler should see no statement in this block, which would normally trigger an error. to be safe, i suggest we add
```
#else
break
#endif
```

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -97,83 +115,113 @@ class ZeroQueryModel: ObservableObject {
...
+            promoCard = .defaultBrowser {
+                self.handlePromoCard(interaction: .PromoDefaultBrowser)
+                self.bvc.presentDBOnboardingViewController(triggerFrom: .defaultBrowserPromoCard)
+            } onClose: {
+                self.handlePromoCard(interaction: .CloseDefaultBrowserPromo)
+                self.promoCard = nil
+                Defaults[.didDismissDefaultBrowserCard] = true
+            }
+        case .shouldDisplayWalletPromo:
+            #if XYZ
```
EdwardJTL, line 180: same here

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -97,83 +115,113 @@ class ZeroQueryModel: ObservableObject {
...
         let promoCardTypeArm = NeevaExperiment.arm(for: .promoCardTypeAfterFirstRun)
 
-        if (promoCardTypeArm == .control || promoCardTypeArm == nil)
-            && shouldShowDefaultBrowserPromoCard()
-        {
-            promoCard = defaultPromoCard()
-        } else if !Defaults[.signedInOnce] && !Defaults[.didDismissPreviewSignUpCard] {
+        switch promocard {
+        case .shouldDisplayArmDefaultPromo:
+            if (promoCardTypeArm == .control || promoCardTypeArm == nil)
```
EdwardJTL, line 131: can these directly return the bool value?

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -97,83 +115,113 @@ class ZeroQueryModel: ObservableObject {
...
+                && (Defaults[.signedInOnce] || !Defaults[.didDismissPreviewSignUpCard])
+            {
+                return true
+            }
+        case .shouldTriggerDefaultPromo:
+            if promoCardTypeArm == .previewSignUp && shouldShowDefaultBrowserPromoCard() {
+                return true
+            }
+        }
+        return false
```
chung-lai, line 165: after returning inside each `case` directly, we won't need this `return false` as it has exhausted all enum cases.

chung-lai (commented):
LGTM. Will approve after couple nits

chung-lai (approved):
LGTM! Thank you Yasmin. As discussed, let's merge this after branch.

PR #3599, opened 2022-05-13
Add Sign in with QR Code

yasminyin96:
* Add Sign in with QR Code Button
* Add invalid QR Code alert
* Add Feature Flag

### How was this tested?
- [X] I tested this manually

Direct to Neeva App & Invalid Alert
<video src="https://user-images.githubusercontent.com/104786010/169921670-e986d01d-d54f-44b7-b258-0b9916768830.MOV">

Feature Flag


Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -279,6 +281,23 @@ struct OtherOptionsPage: View {
...
+                    QRCodeScanButton(
+                        action: {
+                            showQRScanner = true
+                        },
+                        showQRScanner: $showQRScanner
+                    )
+                    .sheet(isPresented: $showQRScanner) {
+                        CodeScannerView(codeTypes: [.qr]) { response in
+                            if case let .success(result) = response {
+                                if let url = URL(string: result.string) {
```
EdwardJTL, line 294: can combine these

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -279,6 +281,23 @@ struct OtherOptionsPage: View {
...
+                                    UIApplication.shared.open(url)
```
EdwardJTL, line 295: does this still work correctly if the app is not the default browser?

Client/Frontend/Intro/AuthButton.swift
```diff
@@ -131,6 +131,33 @@ struct FirstRunCloseButton: View {
     }
 }
 
+public struct QRCodeScanButton: View {
+    var action: () -> Void
+    @Binding public var showQRScanner: Bool
```
EdwardJTL, line 136: this view doesn't seem to be using this value?

Client/Frontend/Intro/OtherOptionsPage.swift
EdwardJTL, line 294: we should also verify that this URL is our sign in url before opening it

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -279,6 +281,23 @@ struct OtherOptionsPage: View {
...
+                                }
+                                showQRScanner = false
```
EdwardJTL, line 297: do we want to present an error if this fails?

Client/Frontend/Intro/OtherOptionsPage.swift
yasminyin96: Good point.

Client/Frontend/Intro/OtherOptionsPage.swift
yasminyin96: It will redirect to Neeva account in Safari.

EdwardJTL: iirc we need to open this URL in our browser to get the cookie into our app @macy-n? maybe should use one of the browser view controller methods here

chung-lai:
@hwiatneeva, a prototype on QR code sign in. Would love to get some feedback on any direction we can go from here from the UX perspective.

hwiatneeva:
Thanks @yasminyin96  -- it's already working, hooray! I think it would be helpful to have it behind the flag so we can continue to test and improve. Having it testable informs the design.

Let me add this as my design TODO for next 1-2 weeks -- so you'll have to wait.

hwiatneeva (commented):
From design -- LGTM to push this **under a flag** so that can help inform a design revision. TY!

Same comment in https://github.com/neevaco/neeva-ios/pull/3599#issuecomment-1127687019

chung-lai:
@yasminyin96, looks like the unit test build fails.

yasminyin96:
> @yasminyin96, looks like the unit test build fails.
Screen Shot 2022-05-16 at 11 33 58 AM

It maybe do not have CodeScanner package. When I push the commit, I did not commit the change in 'Client.xcodeproj/project.pbxproj'.

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -279,6 +282,31 @@ struct OtherOptionsPage: View {
...
+                        QRCodeScanButton(action: {
+                            showQRScanner = true
+                        })
+                        .sheet(isPresented: $showQRScanner) {
+                            CodeScannerView(codeTypes: [.qr]) { result in
+                                switch result {
+                                case .success(let result):
+                                    guard let url = URL(string: result.string) else { return }
+                                    if !checkForSignInToken(in: url) {
+                                        DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
```
EdwardJTL, line 296: what does the .2 delay do here?

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -333,4 +361,19 @@ struct OtherOptionsPage: View {
                 attributes: EnvironmentHelper.shared.getFirstRunAttributes())
         }
     }
+
+    func checkForSignInToken(in url: URL) -> Bool {
```
EdwardJTL, line 365: this function does more than the signature suggests; it shouldn't open the url.

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -333,4 +361,19 @@ struct OtherOptionsPage: View {
...
+        // This is in case the App Clip sign in URL ends up opening the app
+        // Will occur if the app is already installed
+        if url.scheme == "https", NeevaConstants.isAppHost(url.host, allowM1: true),
```
EdwardJTL, line 368: one line per statement

EdwardJTL, line 368: this should be a guard let

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -333,4 +361,19 @@ struct OtherOptionsPage: View {
...
+            let components = URLComponents(url: url, resolvingAgainstBaseURL: true),
+            components.path == "/appclip/login",
+            let queryItems = components.queryItems,
+            let signInToken = queryItems.first(where: { $0.name == "token" })?.value
```
EdwardJTL, line 372: these two lines could be combined

EdwardJTL, line 372: this value is not used. maybe just test for nil?

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import CodeScanner
```
EdwardJTL, line 5: how is this package added? it doesnt seem to be checked into the project

Client/Frontend/Intro/OtherOptionsPage.swift
EdwardJTL: if you trace how the app scheme url is opened, it eventually just calls `bvc.switchToTabForURLOrOpen(newURL, isIncognito: isIncognito)` before going through a number of checks.
we should hook this into `FirstRunButtonActions`, so all these buttons follow the same call back logic. `BrowserViewController` provides the call back function for the action, where you can call  `switchToTabForURLOrOpen(url, isIncognito: isIncognito)` with the url in the qr code.

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -333,4 +361,19 @@ struct OtherOptionsPage: View {
...
+    func checkForSignInToken(in url: URL) -> Bool {
+        // This is in case the App Clip sign in URL ends up opening the app
+        // Will occur if the app is already installed
+        if url.scheme == "https", NeevaConstants.isAppHost(url.host, allowM1: true),
+            let components = URLComponents(url: url, resolvingAgainstBaseURL: true),
+            components.path == "/appclip/login",
+            let queryItems = components.queryItems,
+            let signInToken = queryItems.first(where: { $0.name == "token" })?.value
+        {
+            UIApplication.shared.open(URL(string: "neeva://open-url?url=\(url)")!)
```
EdwardJTL, line 374: see new comment on previous conversation

Client/Frontend/Intro/OtherOptionsPage.swift
EdwardJTL, line 368: this could be one regex check

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -205,10 +206,12 @@ public struct EmailForm: View {
 
 struct OtherOptionsPage: View {
     @EnvironmentObject var model: IntroViewModel
+    @EnvironmentObject var toastViewManager: ToastViewManager
 
     @State var email = ""
     @State var firstname = ""
     @State var password = ""
+    @State var showQRScanner = false
```
EdwardJTL, line 214: should this be in the view model?

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -279,6 +282,31 @@ struct OtherOptionsPage: View {
                         },
                         onSignInMode: $model.onSignInMode
                     ).padding(.top, 10)
+
+                    if FeatureFlag[.qrCodeSignIn] && model.onSignInMode {
+                        QRCodeScanButton(action: {
+                            showQRScanner = true
+                        })
+                        .sheet(isPresented: $showQRScanner) {
```
EdwardJTL, line 290: the view model handles presentation via overlay manager for the other sign in flows. Let's see if we can follow that here too

EdwardJTL, line 290: there is no way to dismiss this view

EdwardJTL, line 290: @chung-lai @hwiatneeva should we first present an interstitial with instructions for how to get this QR code?

Client/Frontend/Intro/OtherOptionsPage.swift
yasminyin96: After the scan page dismissed, then the alert will present.

Client/Frontend/Intro/OtherOptionsPage.swift
yasminyin96: Same comment in https://github.com/neevaco/neeva-ios/pull/3599#discussion_r876141526

yasminyin96: Same comment in https://github.com/neevaco/neeva-ios/pull/3599#discussion_r876141526

Client/Frontend/Intro/OtherOptionsPage.swift
yasminyin96: Same comment in https://github.com/neevaco/neeva-ios/pull/3599#discussion_r876141526

Client/Frontend/Intro/OtherOptionsPage.swift
yasminyin96: I just added this package locally not commit this

Client/Frontend/Intro/OtherOptionsPage.swift
chung-lai: Good question, we are trying to put this behind flag for now so design can experience it. Agree there has to be some kind of instructions on how to get QR code from a website.

Client/Frontend/Intro/OtherOptionsPage.swift
EdwardJTL: we try to avoid constants like this. can you call the present function after you clear the published variable?

Client/Frontend/Intro/OtherOptionsPage.swift
```diff
@@ -333,4 +361,26 @@ struct OtherOptionsPage: View {
                 attributes: EnvironmentHelper.shared.getFirstRunAttributes())
         }
     }
+
+    /// Checks for a sign in token in the URL and also handles the URL if true. (SceneDelegate)
+    func checkForSignInToken(in url: URL) -> Bool {
```
chung-lai, line 366: Instead of duplicating the functionality from `SceneDelegate`, we should add SceneDelegate into IntroViewModel and then call `checkFroSignInToken` directly. Or for now, we can start with calling this directly `SceneDelegate.getCurrentSceneDelegateOrNil().checkForSignInToken(in: xxx)`. But I would suggest we try to pass the SceneDelegate as `getCurrentSceneDelegateOrNil` is not recommended :)

Client/Frontend/Intro/AuthButton.swift
```diff
@@ -131,6 +131,32 @@ struct FirstRunCloseButton: View {
...
 
+public struct QRCodeScanButton: View {
+    var action: () -> Void
+
+    public var body: some View {
+        Button(action: action) {
+            let label = Text("Sign in with QR Code")
+            HStack {
+                Symbol(decorative: .qrcodeViewfinder, style: .labelMedium)
+                    .foregroundColor(.black)
```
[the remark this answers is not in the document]
yasminyin96: I switched to the dark mode, but it seems that sign-in page does not have dark mode.

https://user-images.githubusercontent.com/104786010/169599336-e70e5eff-839d-4df9-8f18-0c69a5f00da6.MP4

Client/Frontend/Intro/OtherOptionsPage.swift
chung-lai: agree. let's move this into `IntroViewModel`

Client/Frontend/Intro/OtherOptionsPage.swift
chung-lai: @yasminyin96, could we try to move the presentation as what Edward pointed out? Otherwise, I'm ok with merging this as is for now as this feature is behind flag. I'm sure we'll have many follow up changes needed for this :)

chung-lai (commented):
Change looks good, pending a couple comments.

Client/Frontend/Intro/OtherOptionsPage.swift
yasminyin96: Sure thing.

PR #3605, opened 2022-05-16
Close Incognito Tabs on Force Quit

evancrow:
Adds the code to detect when the app is force quit, as well as the ability to close the incognito tabs on force quit.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/168698410-6a51287c-5c07-4d69-9c54-fb53375a042d.mp4

### DIFF
--- a/Client/Application/SceneDelegate.swift
+++ b/Client/Application/SceneDelegate.swift
@@ -21,7 +21,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
     private static var activeSceneCount: Int = 0
     private var isInBackground = false
-    
+
     // MARK: - Scene state
     func scene(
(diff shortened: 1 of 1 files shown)


darinwf:
Interesting approach, but it could be a bit unreliable right? Could we use https://developer.apple.com/documentation/uikit/uiscenesession/3197959-staterestorationactivity as a way to know if the app was force closed in the background? E.g., if the user swiped away the app, then the scene would be closed. If you then start the app again, `stateRestorationActivity` would be nil, right?

evancrow:
I talked with @darinwf, and we have decided not to go forward with this PR. Because there is no reliable way to detect a force close of the app, this feature could accidentally backfire.

PR #3617, opened 2022-05-17
Add "open in incognito" option for the share to menu

jedireza:
Adds an "Open in Neeva Incognito" option to the share to menu.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Tested in the simulator.

image

https://user-images.githubusercontent.com/979929/168922479-f2a3e986-5acd-4e51-a129-e32c9432608a.mp4

## Issues addressed
N/A


Extensions/ShareTo/ShareToView.swift
```diff
@@ -49,6 +49,22 @@ struct ShareToView: View {
                             )
                         }
                         Divider()
+                        Button(action: {
+                            item.url.addingPercentEncoding(
```
darinwf, line 53: nit: would probably make sense to refactor this code into a shared helper function to share code between incognito and non-incognito cases.

darinwf (approved):
Nice! I wonder if there is a way to share the existing incognito.pdf asset instead of forking another copy of it.

Extensions/ShareTo/ShareToView.swift
jedireza: Refactored 👍

jedireza:
> Nice! I wonder if there is a way to share the existing incognito.pdf asset instead of forking another copy of it.

I was able to share the incognito asset but I backed that change out.
- Once I used the shared asset, I thought why not move the `open-in-neeva` asset into the main `Client/Assets/images.xcassets` file too?
- But, I actually like the explicit definition for which assets we use in this `ShareToView`.

Let me know if you have a strong opinion on this.

darinwf (approved):

PR #3619, opened 2022-05-18
Clean up unused default

jedireza:
In #3617 I noticed a [default being set](https://github.com/neevaco/neeva-ios/pull/3617/files#diff-3d164b0995c4f68820203a842cfc4fd0a7bb8a75947f751cbfd7a2d13665fb43R118) and curiosity lead me to look up what we use it for. It doesn't look like we use it.

FYI: This PR is based on my previous PR to keep the diff clean.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Tested in the simulator.

## Issues addressed
N/A


darinwf (approved):

darinwf:
Nice, thanks!

Shared/Prefs.swift
```diff
@@ -78,11 +78,6 @@ extension Defaults.Keys {
     public static let topSitesCacheSize = Defaults.Key<Int32?>("profile.topSitesCacheSize")
     public static let neevaUserInfo = Defaults.Key<[String: String]>("UserInfo", default: [:])
 
-    // MARK: - telemetry
-    public static let appExtensionTelemetryOpenUrl = Defaults.Key<Bool?>(
```
EdwardJTL, line 82: i'm a little wary about just removing the code. iirc this library is just syntactical sugar for UserDefaults. If we remove this now, and by change use the exact same default value in the future, then devices with an older user default list will have the value carried over, and our new "default" value won't apply.

maybe we can use this util https://github.com/neevaco/neeva-ios/pull/3059 to manually delete this key in this version?

jedireza: That's a good point. Thanks for sharing the `onAppUpdate` functionality.

jedireza:
@EdwardJTL take a look at the last commit. The `Defaults` library has a `reset` method. I didn't find a `remove` or `delete` method though. I also wonder the proper way to remove something from `Defaults` that isn't optional.

For example; if we defined a non-optional integer pref (say default to `0`), how would we successfully remove that? We'd need to remove the pref definition at some point.

EdwardJTL:
> @EdwardJTL take a look at the last commit. The `Defaults` library has a `reset` method. I didn't find a `remove` or `delete` method though. I also wonder the proper way to remove something from `Defaults` that isn't optional.
> 
> For example; if we defined a non-optional integer pref (say default to `0`), how would we successfully remove that? We'd need to remove the pref definition at some point.

looking at the documentation for the defaults library (assuming we didn't change it in our version), we should be able to access the underlying `UserDefaults` to make changes. https://github.com/sindresorhus/Defaults#its-just-userdefaults-with-sugar
With `UserDefaults`, we can call this function with the right key? https://developer.apple.com/documentation/foundation/userdefaults/1411182-removeobject
then we should be safe to remove the declaration in `Prefs.swift`

alternatively we can mark the property as deprecated using `@available(*, deprecated)`. if we decided to use the value again, we can manually reset the value then.

jedireza:
It looks like `Defaults.reset` calls `removeObject` under the covers, which should work; https://github.com/sindresorhus/Defaults/blob/d1e2109fc99a4471bc0193ae3f521fcf07695153/Sources/Defaults/Defaults.swift#L14

My goal for wrapping up this PR is to create a v0 convention for us to follow for deprecating prefs like this.

jedireza:
So in the case that we are deprecating a non-optional pref definition, we can;

- remove relevant code paths
- redefine the pref to be optional
- add to `onAppUpdate` calling `Defaults.reset(<key>)` with a date of deprecation comment next to it
- sometime in the future when we see most/all users are beyond a certain version, we can actually delete the pref definition and `Defaults.reset(...)` call in `onAppUpdate`

How does that sound?

PR #3620, opened 2022-05-18
Add ad block as a standalone option and additional rule list

chung-lai:
We want to make ad block as a standalone flag so user can turn it on/off.

* Make ad block as a standalone flag
* Add easylist as an additional rule file on top of easyprivacy

From reading https://easylist.to/, EasyPrivacy is an optional supplementary filter list for trackers so easylist doesn't seem like a superset of EasyPrivacy.

Screen Shot 2022-05-17 at 10 48 18 PM

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases

Turn on/off ad block toggle and see ad disappearing from www.yahoo.com


EdwardJTL (approved):
lgtm

Issue #3609, opened 2022-05-17
Modal overlay animation not smooth


jedireza:
Notice how you see a hard line and then fade when dismissing the popover.

## Screenshots or recordings

https://user-images.githubusercontent.com/979929/168871858-45fb17a0-5e6e-43e8-bdec-bb2fd6cb30ff.mp4

PR #3629, opened 2022-05-19
Improve OverlaySheet Animation

evancrow:
Moved the background from the `OverlaySheetView` (which was offset in its entirety by the `OverlayView`) to the actual `OverlayView` where it could be controlled separately.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/169198738-c9ae6a7d-ad8d-4f22-adc6-927ddc776c8c.mp4

## Issues addressed
Fixed: #3609


Client/Frontend/Overlay/Sheet/DismissBackgroundView.swift
```diff
@@ -35,30 +24,3 @@ struct DismissBackgroundView: View {
         .accessibilitySortPriority(-1)
     }
 }
-
-private struct DismissalObserverModifier: AnimatableModifier {
```
darinwf, line 39: Nice to not need this anymore.... but I thought this View was important for knowing when the sheet has been dismissed e.g. by dragging the overlay sheet down. Is that case still handled?

It looks like this may only be handling the case of when the background is tapped to dismiss the sheet.

Client/Frontend/Overlay/OverlayView.swift
```diff
@@ -99,25 +109,34 @@ struct OverlayView: View {
...
-            content
-                .offset(y: overlayManager.offset)
-                .opacity(overlayManager.opacity)
-                .onAnimationCompleted(for: overlayManager.displaying) {
-                    if let animationCompleted = overlayManager.animationCompleted {
-                        animationCompleted()
+            ZStack {
+                if addDismissableBackground {
+                    DismissBackgroundView(opacity: overlayManager.opacity / 5) {
+                        overlayManager.hideCurrentOverlay()
```
darinwf, line 115: any concerns about ordering problems here? what if this overlay is no longer current when this line of code is reached? is there anyway to specifically reference this sheet when asking to hide it?

darinwf (commented):
The screen capture of this change looks great. Some questions though...

Client/Frontend/Overlay/Sheet/DismissBackgroundView.swift
evancrow: Yeah, that's all it did, it's no longer necessary.

Client/Frontend/Overlay/OverlayView.swift
evancrow: The `OverlayView` only has access to the current `overlayView` in `OverlayManager`. I can make it more specific with `hideCurrentOverlay(ofPriority: .modal)` though.

EdwardJTL: `addDismissableBackground` only includes `sheet` and `popover`, but priority `modal` includes more. it would be nice if these conditions can be matched automatically so that the background will always dismiss the views it's presented along and only those ones... not sure how we can do that tho :/

evancrow: We could make a separate priority that is just `sheet` and `popover`?

EdwardJTL (approved):

PR #3635, opened 2022-05-19
Add neeva community to spaces grid for logged out users

buracules:
- Added community spaces to spaces grid for not logged in users

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/15333612/169357280-8e707572-bc1d-43b6-aeee-684c5e68e748.mp4

## Issues addressed


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -534,13 +534,10 @@ class SpaceCardModel: CardModel {
         manager.spotlightEventDelegate = SpotlightLogger.shared
 
         NeevaUserInfo.shared.$isUserLoggedIn.sink { isLoggedIn in
+            self.manager = isLoggedIn ? .shared : .suggested
```
yozuysal, line 537: hmm, I think update the default assigned manager in line 531 as well. This one will only update when the logged in state changes. (I think it does get an update at every start as a side effect, but making sure that the proper default is set right also is important imo

buracules: Thats why i removed initializer changes it does get an update. Now i removed parameter from initializer. And init function will set based on user logged in or not.

Issue #3634, opened 2022-05-19
Cookie Cutter first instance dialog reports different number of trackers than shield icon


darinwf:
## Expected behavior

What's reported in the bottom sheet should match what is reported in the shield.

## Actual behavior

The numbers are different.

## Screenshots or recordings

<img src="https://user-images.githubusercontent.com/79867437/169218204-35a626d3-adb6-4874-b324-518e786e4fdb.jpg" width="300">

## Hardware and software

- Your device: iPhone 12 
- OS version: 15.4.1
- App version: Neeva Browser 1.41.x

FYI @chung-lai @evancrow

PR #3639, opened 2022-05-19
Fix Tracker Stats

evancrow:
Change how `numTrackers` is set, much more reliable now. Also, will update even if the `trackingMenuView` isn't open yet.

### How was this tested?

## Issues addressed
Fixed: #3634


darinwf (approved):
LGTM, thanks!

PR #3640, opened 2022-05-20
Zero query; first run footer branding

jedireza:
This PR adds some branding to the zero query UI before someone does their first web navigation.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

| Light | Dark |
| --- | --- |
| image | image |

## Issues addressed
N/A


Client/Frontend/ZeroQuery/ZeroQueryView.swift
```diff
@@ -160,9 +161,25 @@ struct ZeroQueryView: View {
...
+            firstRunBranding
         #endif
     }
 
+    @ViewBuilder private var firstRunBranding: some View {
+        if !Defaults[.didFirstNavigation] {
+            Spacer()
+            Spacer()
+            VStack(spacing: 8) {
+                Image("neeva-logo-secondary")
```
darinwf, line 173: Can we utilize the existing neeva logo ("splash") and wordmark ("neeva-letter-only") graphics that we already have in the project? We can use "neeva-letter-only" as a template image to give it a different color, I suspect.

jedireza: Hwi and I glanced at the images assets yesterday and noticed that the splash image's leaf was tri-color and not solid. Thanks for mentioning this, I just learned about image rendering modes.

I removed the new files and used the existing `splash` and `neeva-logo-secondary` assets instead. The visual difference is so minor, it's probably best to move forward w/o adding new files. (main screenshots updated)

hwiatneeva (commented):
Adde one comment about reducing the logo N size a bit. 
maxHeight: 22  for visual balance
The rest is LGTM.

Thanks @jedireza!

PR #3643, opened 2022-05-20
Add SpaceIntroView to SpaceGrid

buracules:
- Add SpaceIntroView inside Space Grid
- Also changed UI a little 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

|Spaces Grid for logged out user|
|---|
|Simulator Screen Shot - iPhone 13 Pro - 2022-05-20 at 19 42 04|


Shared/API/SpaceStore.swift
```diff
@@ -715,3 +715,10 @@ public class SpaceStore: ObservableObject {
         }
     }
 }
+
+extension SpaceStore: Equatable {
+    public static func == (lhs: SpaceStore, rhs: SpaceStore) -> Bool {
```
yozuysal, line 720: hehe. This makes sense for now, but I worry it might cause issues down the line. Is there a way to do the default assignment differently to not need the equivalence check?

yozuysal (approved):
This lgtm, but I think it would be good to avoid either avoid the equatable extension, or do it in a way that is more future proof. Suggested space ID is a very niche use of SpaceStore. For example, we might add a MockSpaceStore in the future with nil suggestedSpaceID and then it will be equivalent to SpaceStore.shared etc...

Issue #3631, opened 2022-05-19
Long press menu on back / forward buttons does not show up in Incognito mode


darinwf:
## Expected behavior

Long press on back / forward button in an incognito tab and expect to see a menu of all of the navigations within that tab.

## Actual behavior

This menu does not appear.

## Steps to reproduce the problem

1. Open incognito tab
2. Navigate to some page
3. Click a link
4. Long press on back button
5. No menu is shown

## Hardware and software

- Your device: iPhone 12
- OS version: 15.4.1
- App version: Neeva Browser 1.41.4

evancrow:
It appears that the list isn't showing in normal mode either...

evancrow:
Appears to be a problem with SQL inside `BackForwardListModel`, which hasn't been touched. As well as the fact that we were passing an empty array of URLs to the SQL since we never set `listItems` 🤦‍♂️.

PR #3647, opened 2022-05-21
Fix BackForwardListView

evancrow:
For whatever reason, the SQL that has never been changed stopped working. I updated it to use SQL similar to the history code and now it works fine. The issue with incognito mode has not been fixed though. Because we don't cache sites in incognito mode, we can't load them to use in the `BackForwardListView`. We could set up code to fallback use the `backForwardListItem` but that would be a lot of plumbing to something that likely isn't used often at all.

### How was this tested?
- [x] I added automated tests

## Issues addressed
Fixed: #3631


darinwf:
Why do we need to make SQL queries to populate the back / forward list at all? Why wouldn't we just always base this on `WKWebView.backForwardList`?

Firefox supports showing the back / forward list when you long press on the back button in an incognito tab, so we must have regressed this from Firefox somehow.

I think it is important for this feature to work in incognito mode. While this is a rarely needed feature, when you get stuck on a site that keeps navigating you forward when you try to go back, this can be the only way to get back to where you started (e.g., to the search results page).

evancrow:
I was using `Site` to get the page title for each URL, just now realized that is also included in a `WKBackForwardListItem`, so I'm just gonna switch over to that. Don't know what happened with this PR, been using half a brain cell to fix this...

Storage/SQL/SQLiteHistory.swift
```diff
@@ -140,14 +140,20 @@ open class SQLiteHistory {
     public func getSites(forURLs urls: [String]) -> Deferred<Maybe<Cursor<Site?>>> {
         let inExpression = urls.joined(separator: "\",\"")
         let sql = """
-            SELECT history.id AS historyID, history.url AS url, title, guid, iconID, iconURL, iconDate, iconType, iconWidth
-            FROM view_favicons_widest, history
-            WHERE history.id = siteID AND history.url IN (\"\(inExpression)\")
+            \(Self.selectSiteSharedSQL())
```
darinwf, line 143: Are these SQL related changes still needed?

evancrow: Well without these changes it's technically broken so probably good to leave them in.

evancrow: That is correct. 

The only other place this code is used is the `presentActivityViewController` for `PinToTopSites`. What I found was without this code change, no sites would load, but I haven't yet verified if the pinning is broken.

darinwf: I see. Can you fork this fix into a separate PR? I know it is a bit of extra work to do that, but it helps us have more focused PRs, etc.

darinwf (approved):

PR #3659, opened 2022-05-23
fix select tab logic after removing a tab for time based switcher + index fix

chassu:
The existing logic to select a right or left tab could potentially select a card in a different section. If the flag is enabled, we should change the definition of viableTabs.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Fixed the repro path that darin found out (select a tab not in a group -> go to card grid -> close tab)

Before:

https://user-images.githubusercontent.com/87332917/169774435-ceafe8de-21dd-42df-b286-5941a4f22d41.mp4

After:

https://user-images.githubusercontent.com/87332917/169774451-c9249a4e-d609-46df-9585-5108dddb828c.mp4


Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -95,7 +95,15 @@ extension TabManager {
...
-                if let rightOrLeftTab = viableTabs[safe: deletedIndex]
+                if FeatureFlag[.enableTimeBasedSwitcher] {
+                    // select the next recent tab in today section
+                    let viableTabsToday = viableTabs.filter {
+                        $0.wasLastExecuted(.today)
+                    }
+                    selectTab(
+                        mostRecentTab(inTabs: viableTabsToday), previous: selectedTab,
+                        notify: notify)
+                } else if let rightOrLeftTab = viableTabs[safe: deletedIndex]
```
darinwf, line 106: hmm, why do you want to lose the logic that selects an adjacent tab? that seems nice so the card grid doesn't have to scroll far to show the last used tab.

darinwf:
Seems like we should restrict viable tabs to the today section first and then run the existing leftOrRight algorithm. Only have special handling if there are no other tabs in the today section.

chassu:
Yes. I think leftOrRight within the viableTabsToday makes sense. If there are no other tabs in the today section, I think the behavior is to not select any new tabs and then disable the done button?

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -9,7 +9,11 @@ private let log = Logger.browser
 
 extension TabManager {
     func removeTab(_ tab: Tab, showToast: Bool = false, updateSelectedTab: Bool = true) {
-        guard let index = tabs.firstIndex(where: { $0 === tab }) else { return }
+        guard
+            let index = FeatureFlag[.enableTimeBasedSwitcher]
+                ? normalTabs.firstIndex(where: { $0 == tab })
+                : tabs.firstIndex(where: { $0 === tab })
+        else { return }
```
chassu, line 16: I don't know whether it makes sense that we used to get the index by doing `tabs.firstIndex(where: { $0 === tab })` because the deletedIndex is later used to retrieve elements from `incognitoTabs` or `normalTabs`. I haven't found a way to break things with this logic though. Am I missing something here?

chassu: confirmed with Evan and it seems to be right to use normalTabs/incognitoTabs instead. Made a change.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -9,7 +9,15 @@ private let log = Logger.browser
 
 extension TabManager {
     func removeTab(_ tab: Tab, showToast: Bool = false, updateSelectedTab: Bool = true) {
-        guard let index = tabs.firstIndex(where: { $0 === tab }) else { return }
+        guard
```
evancrow, line 12: Would be good to have a comment explaining this.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -82,7 +90,13 @@ extension TabManager {
     private func updateSelectedTabAfterRemovalOf(_ tab: Tab, deletedIndex: Int, notify: Bool) {
         let closedLastNormalTab = !tab.isIncognito && normalTabs.isEmpty
         let closedLastIncognitoTab = tab.isIncognito && incognitoTabs.isEmpty
-        let viableTabs: [Tab] = tab.isIncognito ? incognitoTabs : normalTabs
+        let viableTabs: [Tab] =
```
evancrow, line 93: Ditto

PR #3660, opened 2022-05-23
Add thumbnail for newly created spaces

### LINKED ISSUE #3654: Better defaults for space item thumbnails
Currently, when we add an item to a space, two things can happen

- Users doesn't click Edit Item and on the server side, we check if the page has og:image and assign that if there is one. We also look for other images, but this part is pretty error prone since it is not running a headless browser but just making a request. (Anything loaded by JS doesn't get picked up)
- User clicks Edit Item and we show a list of thumbnails they can choose from. If they do, we set the thumbnail.

We actually scrape all the thumbnail URLs earlier on and use them only when the user chooses to Edit Item. This is kept as a URL: [URL] map inside SpaceCardModel and is thrown away.

Instead of this, we can always pick the first image URL (they come sorted in size) and assign that to thumbnail proactively. This way, especially for cases where there isn't a og:image tag, we will have more reliable defaults.

buracules:
Downloaded first `jpg`, `jpeg` or `png` image from thumbnailUrl list and assigned it to thumbnail.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

## Issues addressed
* Fixes #3654


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1898,12 +1898,17 @@ extension BrowserViewController {
...
-
-            self.showAddToSpacesSheet(
-                url: url, title: updater?.title ?? title,
-                description: description ?? updater?.description ?? output?.first?.first,
-                thumbnail: thumbnail,
-                importData: importData, updater: updater)
+            let thumbnailUrl = thumbnailUrls.first(where: {
+                $0.absoluteString.hasSuffix("jpeg") || $0.absoluteString.hasSuffix("jpg")
+                    || $0.absoluteString.hasSuffix("png")
+            })
```
yozuysal, line 1904: we should probably null check this before calling downloadImage?

buracules: Actually i tried this one and it was working while it was nil as well. SDWebImage handling that. I can test it again

yozuysal (approved):

PR #3663, opened 2022-05-23
getSites forURLs SQL Fix

evancrow:
Cherry-picked from #3647


darinwf:
Looks like this PR is now in a bad state missing the meaty bits of the intended change....

evancrow:
Think I'm just going to close this actually, not really worth investigation

PR #3667, opened 2022-05-23
Restore App State on Close/Open

evancrow:
Restores the app UI to either the tab/card grid/spaces/detail view after the app is closed and then reopened. There is still a minor issue where sometimes the UserDefault state stores the wrong value. Not sure why it started happening in this case, will investigate some more.

### How was this tested?
- [x] I tested this manually


darinwf:
Test driving, and one issue I notice is that the thumbnail for the currently selected tab seems to get lost.

Steps to repro:

1. Start with an empty tab switcher
2. Open a new tab (e.g., to hacker news)
3. Click on a link
4. Open a new tab also to hacker news
5. Now you have two tabs in a tab group
6. Open the tab switcher
7. Force close the browser
8. Open the browser again 
 
Notice that you are now looking at the tab switcher but the first tab is missing a thumbnail.

darinwf:
Test driving, and one issue I notice is that the thumbnail for the currently selected tab seems to get lost.

Steps to repro:

1. Start with an empty tab switcher
2. Open a new tab (e.g., to hacker news)
3. Click on a link
4. Open a new tab also to hacker news
5. Now you have two tabs in a tab group
6. Open the tab switcher
7. Force close the browser
8. Open the browser again 
 
Notice that you are now looking at the tab switcher but the first tab is missing a thumbnail.

Client/Application/SceneDelegate.swift
```diff
@@ -66,6 +68,26 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
         window!.rootViewController = bvc
         bvc.tabManager.selectedTab?.reload()
```
darinwf, line 70: maybe we should not be calling reload on the selectedTab when opening back to the card grid? that's probably why the selected tab has a blank screenshot.

Client/Application/SceneDelegate.swift
```diff
@@ -134,6 +156,12 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         bvc.downloadQueue.pauseAll()
     }
 
+    func setSceneUIState(to state: SceneUIState) {
+        DispatchQueue.main.async { [self] in
```
darinwf, line 160: why does this need to happen asynchronously?

darinwf, line 160: why does this need to happen asynchronously?

Client/Frontend/Browser/BrowserModel.swift
```diff
@@ -36,9 +36,8 @@ class BrowserModel: ObservableObject {
     var notificationViewManager: NotificationViewManager
 
     func showGridWithAnimation() {
-        if gridModel.switcherState != .tabs {
-            gridModel.switcherState = .tabs
-        }
+        gridModel.setSwitcherState(to: .tabs)
```
darinwf, line 39: This was an optimization to avoid unnecessary updates. You found this change necessary?

Client/Application/SceneDelegate.swift
evancrow: Didn't have to do with reloading the tab, but I removed it anyway. The issue was with the updated `screenshotUUID` not being set in `DiskImageStore`.

Client/Application/SceneDelegate.swift
evancrow: So a weird thing with `UserDefaults`, sometimes if you don't set it from `DispatchQueue` it fails to save which is the issue I was dealing with earlier.

Discussed in [this SO thread](https://stackoverflow.com/questions/41302015/userdefaults-is-not-saved-with-swift)

darinwf: Another thing mentioned in that thread is that it might be valuable to call `UserDefaults.standard.synchronize()` after setting the value.

My worry is that using `DispatchQueue.main.async` could increase the chances of UserDefaults not being persisted during shutdown.

Ideally we'd need to wrap this in a beginBackgroundTask if it is going to happen asynchronously.

darinwf: I think calling `UserDefaults.standard.synchronize()` would be a bad idea though because `setSceneUIState` is called on the main thread and we don't want to block the main thread on I/O.

Client/Application/SceneDelegate.swift
darinwf: ok

Client/Application/SceneDelegate.swift
darinwf: @evancrow - Want to make sure you see my suggestion about using `beginBackgroundTask` to wrap the async update to UserDefaults.

evancrow: Yup! Will take a look.

Client/Application/SceneDelegate.swift
```diff
@@ -156,8 +157,10 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     }
 
     func setSceneUIState(to state: SceneUIState) {
-        DispatchQueue.main.async { [self] in
-            scenePreviousUIState = state.rawValue
+        backgroundProcessor.performTask {
+            DispatchQueue.main.async { [self] in
```
darinwf, line 161: This isn't quite the right way to use `BackgroundTaskProcessor`. In order for it to do its job, you need the task to run to completion and do all of the work it needs to do before the closure passed to `performTask` returns. In this case, adding the `DispatchQueue.main.async` call breaks that. You could change it to a `DispatchQueue.main.sync` if you need the task to run on the main thread. Alternatively, if it is safe to write to `scenePreviousUIState` from a background thread, then you could just drop the `DispatchQueue.main.async` here.

Is the Defaults library thread-safe?

My recommendation is to not use the Defaults library for this one and to just use UserDefaults directly. That way you can be confident about writing to it from a background thread without worry about strange side-effects. Since you don't need any of the SwiftUI benefits of Defaults, this is probably low cost to change.

darinwf (approved):

PR #3668, opened 2022-05-23
[DON'T MERGE] Remove Safari Extension

evancrow:
Due to issues with Apple App Store review, we have decided to remove the Safari Extension, Extension in favor of submitting it as its own product. This PR removes all code from the Neeva iOS App.


jedireza:
IIUC the ideal approach is to first create the independent app and then try our best to migrate current extension users to it before removing it from our core app.

darinwf:
Is it possible to hold off on this PR until we know for sure that we need to do this? Can we build the extension as a standalone app in addition to having it bundled?

evancrow:
Yes agreed, have this set up as needed. I just moved the separate app into the iOS App directory.

evancrow:
@darinwf Is this no longer needed?

evancrow:
Closing the PR, will keep the branch around though... just in case

PR #3670, opened 2022-05-23
Remove tracking protection toggle

chung-lai:
We decided to merge cookie cutter and tracking protection toggle into one. We will use the new `cookie_cutter_enabled` flag and migrate our existing `content_blocking_enabled` flag to `cookie_cutter_enabled`

Screen Shot 2022-05-23 at 4 53 26 PM

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


evancrow:
If we're removing the tracking protection toggle and keeping the Cookie Cutter toggle, does it make more sense to keep the `CookieCutterEnabled` flag?

Client/Frontend/Browser/CookieCutter/CookieCutterModel.swift
```diff
@@ -14,8 +14,8 @@ enum CookieNotices: CaseIterable, Encodable, Decodable {
 extension Defaults.Keys {
     static let cookieCutterOnboardingShowed = Defaults.Key<Bool>(
         "profile.prefkey.cookieCutter.onboardingShowed", default: false)
-    static let cookieCutterEnabled = Defaults.Key<Bool>(
-        "profile.prefkey.cookieCutter.isEnabled", default: true)
+    static let contentBlockingEnabled = Defaults.Key<Bool>(
+        "profile.prefkey.trackingprotection.normalbrowsing", default: true)
```
darinwf, line 18: nit: I suggest when creating a new `Defaults` key to avoid `.` characters as they get in the way of using `Defaults` in SwiftUI. I recommend using `_` characters instead.

darinwf (approved):

chung-lai:
> If we're removing the tracking protection toggle and keeping the Cookie Cutter toggle, does it make more sense to keep the `CookieCutterEnabled` flag?

Yes, we can have a follow up PR to rename that. Didn't want to rename it in this release as we will need to modify the version upgrade to move the existing users' setting to the new name :D

Client/Frontend/Browser/CookieCutter/CookieCutterModel.swift
chung-lai: yes, totally. This was for an existing setting, will address that when we move to a new name :)

chung-lai:
> > If we're removing the tracking protection toggle and keeping the Cookie Cutter toggle, does it make more sense to keep the `CookieCutterEnabled` flag?
> 
> Yes, we can have a follow up PR to rename that. Didn't want to rename it in this release as we will need to modify the version upgrade to move the existing users' setting to the new name :D

Actually, thinking about it, may be easier to take care of upgrade flow so we will save another code change in the future :P

chung-lai:
@darinwf @evancrow renamed the default variables as users haven't had those in their app yet so it's good time to do it now. Also, migrating `content_blocking_enabled` to `cookie_cutter_enabled` so everything is named after cookie cutter. PTAL

Client/Frontend/Browser/CookieCutter/CookieCutterModel.swift
chung-lai: Let's change them all :)

evancrow (approved):
LGTM

Issue #3678, opened 2022-05-24
Update bootstrap.sh to use `yarn` instead of `npm`


jedireza:
We should also remove the `package-lock.json` since that's what `npm` is using and `yarn` ignores.

PR #3680, opened 2022-05-24
bootstrap; replace `npm` calls with `yarn` + syntax consistency

jedireza:
Updates the bootstrap script to use `yarn` instead of `npm`. We shouldn't be mixing the two. I also fixed some inconsistent syntax as well as updated a doc that was referencing `npm`.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested the bootstrap script locally on the terminal.

N/A

## Issues addressed
Resolves https://github.com/neevaco/neeva-ios/issues/3678


darinwf (approved):
👍

PR #3517, opened 2022-05-05
Add Find Tab in CardGrid OverflowMenu

evancrow:
- Creates a generic `FindView`
- Adds option to open the Find Tab menu from `CardGrid` 
- Search for tab code

### How was this tested?
- [x] I added automated tests

### Screenshots and screen recordings

https://user-images.githubusercontent.com/37356984/167012874-de25877d-a2a9-4a4f-84d7-ac30323a314f.mp4

## Issues addressed
Fixed: #3504

PR #3681, opened 2022-05-24
make scroll view scroll after bottom tab is closed

chassu:
this change was merged before, but accidentally got removed in https://github.com/neevaco/neeva-ios/pull/3517/files


evancrow (approved):
LGTM - also, let's mark this as cherry-pick-needed

Issue #3655, opened 2022-05-22
Add space to a space


yozuysal:
Currently it is fairly hard to add a space to a space even though we have access to everything we need.

As a starting point, we should add a menu item to Space Detail called "Add to a Space". On click, we can open the addToSpace sheet (similar to how addToAnotherSpace works in SpaceDetailList.swift line 125) with the necessary info (Space url, Space name, Space description, Space thumbnail will be all available).

PR #3683, opened 2022-05-24
Burak/add to space

buracules:
- Add space to space capability
- Fixed navigation for space inside space

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/15333612/170106944-074a62c1-5dee-4192-94f1-e28dc7010db5.mp4

## Issues addressed
* #3655


Client/Frontend/Browser/Spaces/SpaceTopView.swift
```diff
@@ -259,6 +265,7 @@ struct SpaceTopView: View {
             content: {
                 if let space = space, !space.isDefaultSpace {
                     deleteButton
+                    addToAnotherSpaceButton
```
yozuysal, line 268: for this one, we should also check whether the user can edit this space or not. (Delete button turns into unfollow button if the user can't edit, but this button shouldn't show at all

yozuysal (approved):

Client/Frontend/Browser/Spaces/SpaceTopView.swift
buracules: Do we not want adding another people spaces to ours?

yozuysal: ugh.. yeah you are right. we do...

PR #3687, opened 2022-05-24
zero query; only set min height for first run + not incognito

jedireza:
This PR fixes a regression, primarily for the incognito zero query screen. Essentially we should only be setting the min-height when we're in zero query first run, and not incognito.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

| before | after |
| --- | --- |
| image | image (1) |

## Issues addressed
N/A


Client/Frontend/ZeroQuery/ZeroQueryView.swift
```diff
@@ -136,7 +136,10 @@ struct ZeroQueryView: View {
                     }
                     Spacer()
                 }
-                .frame(minHeight: geom.size.height)
+                // only set for zero query first run
+                .if(!viewModel.isIncognito && !Defaults[.didFirstNavigation]) { view in
```
chung-lai, line 140: user should only be able to change to incognito mode after doing first navigation. `!Defaults[.didFirstNavigation]` should be sufficient and we don't need the `isIncognito` check?

chung-lai (commented):
Good catch! We may want to cherry pick this one.

Client/Frontend/ZeroQuery/ZeroQueryView.swift
jedireza: Yes that's fair. FIxed.

chung-lai (approved):

PR #3689, opened 2022-05-24
Overlay Refactoring

evancrow:
- Removes the `WindowManager` which is no longer used
- Increases the use of `hideOverlay` instead of `hideCurrentOverlay` which is more reliable
- Fixes a bug where full-screen views were assigned as `modal` instead of `fullScreen`
- Moves some other code around for easier reading


jedireza:
@evancrow if you merge `main` into your branch the failed test should get cleared up.

EdwardJTL:
Simulator Screen Shot - iPad Air (5th generation) - 2022-05-25 at 14 10 13
fyi first run shows up as a "modal" looking view on ipad. maybe that's reason we had it assigned as `modal` instead of `fullscreen`? just for consideration. the close button still works!

evancrow:
That's the iPad Fullscreen system element actually (so it should look like that). I was the one that accidentally labeled `FullScreen` as `Modal`.

Issue #3686, opened 2022-05-24
Undismissable sign in promo card show up after default browser interstitial


chung-lai:
## Expected behavior

Preview sign up promo card should show up

## Actual behavior

Undismissable sign in promo card show up

## Steps to reproduce the problem

1. Install app
1. Open Zero query

## Screenshots or recordings

<img width="628" alt="Screen Shot 2022-05-24 at 1 42 36 PM" src="https://user-images.githubusercontent.com/87332578/170136252-344e0138-4438-4869-be60-bcffc4eaddad.png">

@jedireza noticed the undismissable sign in promo card show up after installing app.

jedireza:
IIUC this should _not_ show up before a first navigation has took place.

PR #3691, opened 2022-05-24
Fix undismissable sign in promo card show up after default browser interstitial

yasminyin96:
* Fix undismissable sign in promo card show up after default browser interstitial

### How was this tested?
- [X] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/104786010/170144333-3d6f6d8f-6e11-419b-aa00-ed3f1eedc3b2.mov

## Issues addressed
* https://github.com/neevaco/neeva-ios/issues/3686


Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -139,7 +139,6 @@ class ZeroQueryModel: ObservableObject {
             #endif
         case .shouldTriggerSignUpPromo:
             return !Defaults[.signedInOnce] && !Defaults[.didDismissPreviewSignUpCard]
-                && Defaults[.didFirstNavigation] && NeevaConstants.currentTarget != .xyz
```
chung-lai, line 142: We should do the same thing for Wallet promo. https://github.com/neevaco/neeva-ios/blob/05628a8b5f073bfd60a2e5c49fb58d8098e63289/Client/Frontend/ZeroQuery/ZeroQueryModel.swift#L127-L135

yasminyin96: Could we combine them together into shouldTriggerWalletOrSignup since they are under same condition: `!Defaults[.signedInOnce] && !Defaults[.didDismissPreviewSignUpCard]`. Or leave them separately.

yasminyin96:
```
case .shouldTriggerWalletOrSignupPromo:
  return !Defaults[.signedInOnce] && !Defaults[.didDismissPreviewSignUpCard]
```


Then in setPromo function:
```
case .shouldTriggerWalletOrSignupPromo:
  #if XYZ
         if Defaults[.cryptoPublicKey].isEmpty {
              promoCard = .walletPromo {
                   self.bvc.web3Model.showWalletPanel()
              }
          } else {
              promoCard = nil
          }
   #else
          if Defaults[.didFirstNavigation] && NeevaConstants.currentTarget != .xyz {
                    promoCard = .previewModeSignUp {
                        self.handlePromoCard(interaction: .PreviewModePromoSignup)
                        self.signIn()
                    } onClose: {
                      self.handlePromoCard(interaction: .ClosePreviewSignUpPromo)
                      self.promoCard = nil
                      Defaults[.didDismissPreviewSignUpCard] = true
                    }
           } else {
                promoCard = nil
           }
       #endif
```

chung-lai: i think it makes sense to have them separately and we can iterate on it in the future.

chung-lai: i think we may need to combine them otherwise the sign up card will always get triggered before the wallet one unless we add `#if` check.

PR #3695, opened 2022-05-25
suppress promo before first navigation

jedireza:
This PR suppresses any promos on zero query before any navigations have happened.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

N/A

## Issues addressed
N/A

### DIFF
--- a/Client/Frontend/ZeroQuery/ZeroQueryModel.swift
+++ b/Client/Frontend/ZeroQuery/ZeroQueryModel.swift
@@ -252,17 +252,24 @@ class ZeroQueryModel: ObservableObject {
     }
 
     func shouldShowDefaultBrowserPromoCard() -> Bool {
+        let currentTargetIsClient = NeevaConstants.currentTarget == .client
+        let didNotDismissDefaultBrowserCard = !Defaults[.didDismissDefaultBrowserCard]
+        let didNotSetDefaultBrowser = !Defaults[.didSetDefaultBrowser]
+        let didFirstNavigation = Defaults[.didFirstNavigation]
+
         let notSeenInterstitial =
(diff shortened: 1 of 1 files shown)


Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -248,17 +252,24 @@ class ZeroQueryModel: ObservableObject {
...
         let notSeenInterstitial =
             !Defaults[.didShowDefaultBrowserInterstitial]
             && !Defaults[.didShowDefaultBrowserInterstitialFromSkipToBrowser]
-
-        return NeevaConstants.currentTarget == .client && !Defaults[.didDismissDefaultBrowserCard]
-            && !Defaults[.didSetDefaultBrowser]
-            && Defaults[.didFirstNavigation]
+        let satisfiesFreqRule = satisfyDefaultBrowserPromoFreqRule()
+        //let lastDefaultBrowserInterstitialSkipped = DefaultBrowserInterstitialChoice(
+        //    rawValue: Defaults[.lastDefaultBrowserInterstitialChoice]) == .skipForNow
```
jedireza, line 265: @chung-lai I was able to narrow down why the test case fails to this. When we close the default browser interstitial, this ends up being `true`.

chung-lai: we do want this behavior to show only default browser promo card to user who skipped the default browser interstitial

jedireza: Ok I'll close. It was mainly an exercise in debugging where this was happening.

PR #3697, opened 2022-05-25
Preparing for build 227

jedireza:


chung-lai (approved):
We should still merge this otherwise the version script may act funky when it sees the branch already exists. It's not a big deal, we just simply skip one build version for `main`

Issue #3696, opened 2022-05-25
testTriggerSignInModalAndClose failure


chung-lai:
Test `testTriggerSignInModalAndClose` seem to fail. There are few things going on:

* For our experiment control arm, we show "Set as Default Browser" promo card. We may want to set that to be the experiment arm or disable the default browser promo by setting `Defaults`'s `didDismissDefaultBrowserCard`
* We may consider moving the test into a new suite and set the `Defaults` properly so we can make sure we are testing the promo card logic properly

fyi @jonathangarelick

jedireza:
Moved to ClickUp 🔒 https://app.clickup.com/t/3w359mz

PR #3698, opened 2022-05-25
Disable testTriggerSignInModalAndClose

chung-lai:
#3696 as it's failing in many branches


yozuysal (approved):

chung-lai:
Will skip this test for now and we will have a follow up PR for fixing it by setting the right default values to make it a first run experience.

PR #3700, opened 2022-05-25
Add automated test to cancel button in signin/signup page

yasminyin96:
* Add automated test to cancel button in signin/signup page

### How was this tested?
- [X] I added automated tests

## Issues addressed
* https://github.com/neevaco/neeva-ios/issues/3699


XCUITests/NeevaSignInTests.swift
```diff
@@ -153,5 +160,14 @@ class NeevaSignInTests: BaseTestCase {
         waitForExistence(app.buttons["Sign in with Apple"])
         waitForExistence(app.buttons["Sign in with Google"])
         waitForExistence(app.buttons["Sign in with Microsoft"])
+
+        testCancelButton()
+    }
+
+    func testCancelButton() {
```
chung-lai, line 167: Did we mean to test this functionality on its own? Not entirely sure how automation picks up the test. Does it look at any function prefix with `test` for testing? If so, may be we should rename this to `closeSignInModal`? 

@jonathangarelick, wdyt?

yasminyin96: Not test its own. Because there are 3 cancel buttons in signin and signup page, I planed to call this function instead of duplicated code.

chung-lai: got it, in that case, may be it's less confusing to call this `tapCloseButton` so we are sure it's not an individual test :)

chung-lai (approved):
LGTM! thank you. pending test passing before we merge :)

yasminyin96:
> LGTM! thank you. pending test passing before we merge :)

I saw that build for iPad failed, so I push the new commit which also works for iPad.

PR #3701, opened 2022-05-25
Add ad block toggle

chung-lai:
* Add ad blocking toggle
* Only enable ad blocking toggle in strict mode
* Add ad blocking filter list when toggle is ON

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

| Strict mode w/o ad block | Strict mode with ad block | Standard mode |
| --- | --- | ---
| Screen Shot 2022-05-25 at 12 45 41 PM | Screen Shot 2022-05-25 at 12 45 42 PM | Screen Shot 2022-05-25 at 2 51 32 PM |


chung-lai:
@hwiatneeva, would love a review on the design change. Thanks!

chung-lai:
> Suggestion on ui states here, but please chime in with holes and disagreement:
> 
> 1. Standard: Always turn the ad block switch **OFF & DEACTIVATED** (regardless of the previous status)
> 2. Since the Standard always forces the ad block switch to OFF & DEACTIVATED per 1, the Strict starts with the same state, i.e. OFF & ACTIVATED
> 3. Strict: **Start with** the ad block switch OFF & **ACTIVATED**, and the user can manually turn ON and OFF the ad block switch
> 4. When goes from Strict --> Standard: per 1, the ad block switch turns back to the initial state OFF & DEACTIVATED.

That makes sense. User should turn on ad blocker intentionally so we don't have to persist the toggle for them. And setting it off between switching standard vs strict, it will be less confusing having that state persist. Thanks!

evancrow (approved):
LGTM one nit

PR #3703, opened 2022-05-26
Cookie Cutter Menu Fixes

evancrow:
Hide site-specific settings toggle when the global setting is disabled.

### How was this tested?
- [x] I tested this manually

Disabled global setting UI:


chung-lai (approved):
LGTM!

Issue #3705, opened 2022-05-26
Neeva Login Sheet Shows Multiple Times


evancrow:
Steps to repro:
- Log out
- Perform search
- Wait for the sign-in sheet and close it
- Sheet pops up again (should not)

https://user-images.githubusercontent.com/37356984/170543470-4ed75e99-e636-47f6-91bf-786ae34e778d.mp4

FYI @chung-lai

chung-lai:
This one is interesting. Not able to reproduce on the testflight build. And also the sign in screen I see is different from the modal view one. We should be seeing a preview user sign in screen hm..

jedireza:
I reproduced this 1:1 on the latest test flight build. My steps:
- Be logged in
- Log out
- Run search (i used `lemons` too 😉)
- See login sheet, dismiss it
- See login sheet again

https://user-images.githubusercontent.com/979929/170832566-8ab0c9da-6752-4471-8423-880e001601b7.mov

PR #3707, opened 2022-05-26
Prevent IntroView Present Loop

evancrow:
Because of the weird delegate layout in `TabManager` which needs to be refactored, there were multiple calls made to `presentIntroViewController` from `decidePolicyFor navigationAction`. Now before presenting the `IntroView`, check to make sure a `FullScreenModal` isn't already presented.

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #3705


Client/Frontend/Browser/BrowserViewController/BrowserViewController+IntroView.swift
```diff
@@ -17,6 +17,11 @@ extension BrowserViewController {
         completion: (() -> Void)? = nil,
         onDismiss: (() -> Void)? = nil
     ) {
+        // Confirm a FullScreenView isn't already visible.
+        if case .fullScreenModal = overlayManager.currentOverlay {
```
chung-lai, line 21: Pretty sure this cover most cases. Wonder whether we can also check the current full screen modal is indeed a login modal?

chung-lai: Sure, i think that's sufficient too.

chung-lai (approved):
LGTM. One thing worth confirming if `dismiss` is being called all the time even on iPad when user taps on the background to dismiss the screen.

chung-lai:
May worth cherry picking this.

evancrow:
> LGTM. One thing worth confirming if `dismiss` is being called all the time even on iPad when user taps on the background to dismiss the screen.

Good catch! Moved the code so it will be called for all interaction paths.

PR #3708, opened 2022-05-26
Custom icons; fix ITMS-90892 warning

jedireza:
This PR should fix the ITMS-90892 warning/issue we saw from App Store Connect.

SO Ref: https://stackoverflow.com/a/70186649

### How was this tested?
N/A -- I wonder if we could verify this locally before a test flight release 🤔

### Manual test cases
N/A

N/A

## Issues addressed
N/A


chung-lai (approved):
LGTM. I know there is a binary validation we can try but I had no luck with that. I would say we should just upload another Testflight build.

chung-lai:
Project change looks good! Btw, is the feature available for Wallet app? If so, make sure we set the target for those images to be available on Wallet target too.

jedireza:
No, it's not. We ensure it's not xyz in settings. That would likely need different branding considerations.

PR #3711, opened 2022-05-27
Fix Open Page Settings

evancrow:
Fixes a bug where opening settings to a specific page wouldn't work if the section wasn't yet rendered (because `List` is lazy). The fix was just to scroll to the section, bringing it into view, rendering the `NavigationView`.

https://user-images.githubusercontent.com/37356984/170608738-9cd85f10-d54e-4904-b8f7-6f5947dd05f7.mp4


chassu (approved):
LGTM

Issue #3714, opened 2022-05-27
Support submission reporting incorrect source


jedireza:
On the ticket I'm seeing the source as `IOSWeb3App`.

🔒 https://github.com/neevaco/neeva/issues/69297
🔒 https://github.com/neevaco/neeva/issues/69365

PR #3716, opened 2022-05-27
send feedback mutation; fix source parameter

jedireza:
This PR fixes the source parameter of the feedback form submission. I noticed it was always being set as `IOSWeb3App`.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested this in the simulator. Here are the last two test submission records: 
- 🔒 https://github.com/neevaco/neeva/issues/69371
- 🔒 https://github.com/neevaco/neeva/issues/69372

N/A

## Issues addressed
Resolves https://github.com/neevaco/neeva-ios/issues/3714


Shared/UI/Feedback/SendFeedbackView.swift
```diff
@@ -333,7 +333,8 @@ public struct SendFeedbackView: View {
                         shareResults: shareResults,
                         requestId: (requestId?.isEmpty ?? true) ? nil : requestId,
                         geoLocationStatus: geoLocationStatus,
-                        source: .iosWeb3App,
+                        source: NeevaConstants.appGroup.starts(with: "group.xyz")
```
jedireza, line 336: I spent a while trying to figure out why `NeevaConstants.currentTarget` isn't available here. The extension to `NeevaConstants` for this only linked to `Client` and `XYZClient`. I tried to link it to shared and while it was available, it didn't work as expected and was always `.client`.

I looked into passing the `FeedbackSource` down through the views, but that started to feel way too complicated.

Let me know if I missed something there.

yozuysal (approved):
thanks for catching this. I think as long as it works for client target we should be fine.

PR #3722, opened 2022-05-28
Revert constraint changes for image context menu

### LINKED ISSUE #3721: New image context menu long press delay + visuals
I noticed on the latest test flight build the long press feels very long, I almost gave up on it thinking it wasn't going to work. Also the visuals I experienced were different than the screenshots. The URL row is expanded for me and the menu looks very small and hard to tap.

https://user-images.githubusercontent.com/979929/170832225-926a3f20-a4a7-4d92-a885-3522315f3bc6.MP4

Related to: https://github.com/neevaco/neeva-ios/pull/3709

buracules:
- Revert latest constraint changes for image context menu
### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

|Before|After|
|---|---|
|Simulator Screen Shot - iPhone 13 Pro - 2022-05-28 at 18 49 27|Simulator Screen Shot - iPhone 13 Pro - 2022-05-28 at 18 46 17|

## Issues addressed
* Fixes #3721


jedireza:
The menu looks better now. The font size of the menu items seems small still.

context-menu-sizes

But regarding the long press; if we compare it to the tab menu, it takes a lot longer to trigger. I did notice that if I long press then release, it comes up quicker, but that didn't feel like the standard/traditional long press behavior across iOS.

https://user-images.githubusercontent.com/979929/170833414-6d05ede9-0901-466e-868f-6e4361537ab1.mp4

jedireza:
If the text "Add to Space with Image" is too long and why we have a smaller font size, I think "Add to Space" is reasonable since in this context you're working with a particular image.

evancrow:
Do we know why the UIViewExtensions caused the previous issue? I don't think we should go back to setting each view side's constraint, makes it harder to read and difficult to maintain in the future.

buracules:
@evancrow i will change it back i dont have much time today

buracules:
Fixed the problems with extension i missed a greaterThanOrEqual. Also updated UI.

Simulator Screen Shot - iPhone 13 Pro - 2022-05-29 at 06 02 33

jedireza (approved):
Thanks for the quick turn around! I test drove and things are much better. I think we may want to revisit the long press timing/release mechanics but we don't need to block on it.

PR #3727, opened 2022-05-30
Fix BackForwardList Full Screen Cutoff

evancrow:
### How was this tested?
- [x] I tested this manually

Before:

https://user-images.githubusercontent.com/37356984/171034560-427bd517-7fa5-40bc-8134-2776904edf98.mp4

After:

https://user-images.githubusercontent.com/37356984/171034567-76865a6b-6c6c-4e48-b731-766e966345ea.mp4


Client/Frontend/Browser/BackForwardListView.swift
```diff
@@ -50,92 +50,103 @@ struct BackForwardListView: View {
...
     }
 
     var body: some View {
-        VStack(spacing: 0) {
-            DismissBackgroundView(opacity: 0.2) {
-                overlayManager.hideCurrentOverlay()
-            }.animation(nil).transition(.fade)
+        GeometryReader { geom in
+            let maxHeight: CGFloat = {
+                geom.size.height - 60
```
darinwf, line 125: How is this magic constant of 60 determined? Can it be calculated from other constants?

evancrow: 60 is the minimum amount of padding between the top of the view and the top of the BackForwardList so that way the user can tap to close, totally made up. Will add a comment.

darinwf: OK, yeah, please add a comment about it. Thanks!

darinwf (approved):

Issue #3724, opened 2022-05-29
Cannot Open Duplicated Tab


evancrow:
Steps to repro:
- Open a tab
- Go to CardGrid
- Long press on tab and duplicate it
- Tap duplicated tab but it never opens (even though it is selected)

FYI @chassu

PR #3729, opened 2022-05-30
Fix Duplicate Tab

evancrow:
Refactored the logic down a bit, and made a reusable function. Not exactly sure what the problem was, but my refactoring seemed to fix the bug. I know the tab wasn't opening because the completion handler on `scrollToSelectedTab` in `hideGridWithAnimation` wasn't being called with the old code.

### How was this tested?
- [x] I added automated tests

## Issues addressed
Fixed: #3724


Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -176,6 +176,13 @@ extension TabManager {
         }
     }
 
+    func duplicateTab(_ tab: Tab, incognito: Bool) {
+        guard let url = tab.url else { return }
+        let newTab = addTab(
+            URLRequest(url: url), afterTab: tab, isIncognito: incognito)
+        selectTab(newTab, notify: true)
```
darinwf, line 183: There seems to be two changes:
1. The old code was passing `previous: tab` as a parameter to `selectTab`. The new code no longer does that. 
2. The old code was setting `newTab.rootUUID`. The new code no longer does that.

Those seem to be the only real changes here. Why would those changes fix the bug? Does it break anything else by not passing `previous: tab` or not settings `rootUUID`?

darinwf: I think this change would have the effect of causing the duplicated tab to be part of the same tab group as the parent tab. That doesn't seem right. We want the duplicated tab to be a new top-level tab.

evancrow: The previous code also added the duplicate tab to the existing tab group so I didn't want to change that but sounds like that may be a bug? I think passing `previousTab` added it to the tab group, which is what `afterTab` is doing here.

darinwf: I see... I think I remembered incorrectly the behavior of the app. Despite what I wrote before, I'm okay keeping the behavior of the duplicated tab being in the same tab group.

darinwf (approved):
LGTM

Issue #3730, opened 2022-05-30
Performance; slowness with 300+ tabs


jedireza:
User report via Slack:

> I've ended up having a very very very large number of tabs. Now it takes 6 seconds to even close a tab.
> Finally gave up and closed all tabs. 301!
> 301 tabs

## Hardware and software

- Your device: unknown
- OS version: 15.5
- App version: Neeva Browser  1.43.1 (227)

darinwf:
I'm also seeing this. I think it means we need to do some combination of 1) optimize buildRows, 2) minimize calls to buildRows, 3) explore a way to support removing a tab without calling buildRows, 4) enable tab archiving and take advantage of it to greatly speed up buildRows.

@chassu FYI

darinwf:
I used the option to add 500 tabs, and then removed a single tab. What I observe:

1- buildRows is called two times
2- under the simulator, seeing it cost about 12-17 milliseconds each call
3- on an iphone 12, seeing it cost about 26-33 milliseconds each call
4- there seems to be a significant delay before buildRows gets called

Point 4 makes me think there is more going on here than just buildRows being slow. And 26-33 msec x 2 does not explain the sluggishness we are seeing.

darinwf:
Yeah... looking at TabManager.removeTabs, with 500 tabs on an iPhone 12:

I see that removeTabs is taking about 500 msec. buildRows is just a fraction of that...

evancrow:
The whole app is slow, not just when closing tabs. So the issue probably isn't with any of the close functions but something that affects everything.

darinwf:
> The whole app is slow, not just when closing tabs. So the issue probably isn't with any of the close functions but something that affects everything.

I agree there may be multiple issues, but closing a single tab when you have 100s of tabs is one issue that really stands out and should be easy to debug.

I'm seeing that TabCardModel.onDataUpdated is what's slow, and like I said, it isn't all explained by buildRows. That's interesting since the part that isn't buildRows is simpler.

evancrow:
I ran the profiler and it looks like a SwiftUI updater that calls `FaviconView` is using a lot of CPU (and almost a second of run time).

evancrow:
Think that might be a good place to start. I will try removing `FaviconView` to see if that fixes the issue to confirm if that is the problem.

:)

[...]

PR #3735, opened 2022-05-31
Stop computing allDetailsWithExclusionList and representativeTabs

darinwf:
This seems to greatly speed up onDataUpdated.

Issue #3730

### How was this tested?
- [x] I tested this manually


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -33,7 +33,7 @@ class TabCardModel: CardModel {
     private(set) var incognitoRows: [Row] = []
     private(set) var timeBasedIncognitoRows: [TimeFilter: [Row]] = [:]
     private(set) var allDetails: [TabCardDetails] = []
-    private(set) var allDetailsWithExclusionList: [TabCardDetails] = []
+    private(set) var allDetailsWithExclusionList: [TabCardDetails] = []  // Unused
```
evancrow, line 36: Why keep this around if it's not used?

evancrow (approved):

chassu (approved):

Client/Frontend/Browser/Card/CardModels.swift
darinwf: It is to make the compiler happy. `CardModel` requires that this property exist. I think we should follow-up with a PR to remove this field from `CardModel`.

PR #3738, opened 2022-06-01
Fix Close Tab Keyboard Shortcut

evancrow:
Fixes a bug where the `WebView` wouldn't update when calling `removeTabs`. Also changes this specific shortcut to use `removeTab` since that better fits the use case.

### How was this tested?
- [x] I added automated tests


evancrow:
Don't know if a change was made to KIF, but now `UITests` will fail if you try and find an object via its accessibilityIdentifier. Only `accessibilityLabel` works.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -416,6 +416,7 @@ class TabManager: NSObject {
 
     func sendSelectTabNotifications(previous: Tab? = nil) {
         selectedTabPublisher.send(selectedTab)
+        updateWebViewForSelectedTab(notify: true)
```
chung-lai, line 419: Will this cause any side effect? Wonder why this is needed only for keyboard case.

chung-lai (commented):
One comment and looks like there is test failure.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
evancrow: I actually added this here since it isn't just needed for the keyboard case. Bug was caused by `WebView` not being updated when the `selectedTab` changed, this makes sure that the notification is sent.

chung-lai (approved):
LGTM after addressing the test failure.

PR #3743, opened 2022-06-01
Add follow support for spaces also small fix for SpaceTopView

buracules:
- Fixed the issue SpaceTopView buttons not showing when you open a space from Profile UI
- Added follow space support

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/15333612/171464335-5ff67d41-c72d-4f0c-9f0d-bcb48b6db294.mp4


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -533,7 +533,7 @@ class SpaceCardModel: CardModel {
 
     private func listenManagerState() {
         self.anyCancellable = manager.$state.sink { [weak self] state in
-            guard let self = self, self.detailedSpace == nil, case .ready = state,
```
buracules, line 536: @yozuysal Do you think this is gonna break something? I wanted to update spaces in the background.

Shared/API/SpaceStore.swift
```diff
@@ -376,12 +376,20 @@ public class SpaceStore: ObservableObject {
...
             switch result {
-            case .success:
+            case .success(let data):
                 Logger.browser.info("Space followed")
-                completion()
+                if let model = data.first {
+                    let space = Space(from: model)
+                    space.contentData = model.entities
+                    self.allSpaces.append(space)
+                    completion()
```
buracules, line 389: Actually i dont need completion handler here i am gonna delete this one. And the reason i did it like this is because we are already trying to remove refresh for mutations. I am just using the data i get from this request

buracules:
@jonathangarelick FYI

Client/Frontend/Browser/Card/CardModels.swift
buracules: I didn't see anything weird yet but maybe i didn't test it enough

Shared/API/SpaceStore.swift
buracules: Actually i will take this change back and just call `refresh()` for this PR and i can handle this on the other PR i am working on (Its about removing unnecessary refreshes).

yozuysal (approved):

Issue #3651, opened 2022-05-21
Sometimes search suggestions on zero-query shows the initial new user set of suggestions instead of the personalized suggestions


darinwf:
## Expected behavior

I expect to see search query suggestions on the zero-query based on my local search history.

## Actual behavior

Sometimes, not sure how to repro this, I instead see the default set of search queries. See screenshot below.

## Screenshots or recordings

<img src="https://user-images.githubusercontent.com/79867437/169672049-77093b96-7739-4ae5-a9a7-1b276c1cf750.jpg" width="300">

## Hardware and software

- Your device: iPhone 12
- OS version: 15.4.1
- App version: Neeva Browser 1.42.0 (223)

FYI @jedireza @yozuysal

jedireza:
I have a suspicion that there is a race condition of some sort happening. Seeing all 3 of the example queries implies the `suggestedQueries` is empty.

https://github.com/neevaco/neeva-ios/blob/4b91333e61e0b69a952203c535345199f7fe1e4a/Client/Frontend/ZeroQuery/ZeroQuerySuggestionModels.swift#L52-L61

darinwf:
Just FYI, that I continue to see this with the latest TestFlight. Not sure how to repro :-/

darinwf:
Perhaps it would be helpful to add some logging using `Logger.browser.info`, then I can look at those log files on my device when I observe this problem.

darinwf:
I still see this bug. What I also just noticed is that when this bug repros, I also see that the History UI fails to load. The History UI will just be blank and indicate that there is no history. So both of these issues are probably explained by some bug causing the history db to not load properly. Restarting the app solves the problem in these cases.

darinwf:
I continue to see this bug. I think the underlying database connection is getting into a bad state. Not sure why that is happening yet.

I see this in the logs:

```
2022-08-22 16:45:59.543 [Debug] [SwiftData.swift:552] init(filename:flags:schema:files:) > Database successfully created or updated.
2022-08-22 16:45:59.546 [Debug] [SwiftData.swift:630] interrupt() > Interrupt
...
2022-08-22 16:46:20.297 [Debug] [Profile.swift:178] _shutdown() > Shutting down profile.
2022-08-22 16:46:20.303 [Debug] [SwiftData.swift:621] deinit > deinit: closing connection on thread <_NSMainThread: 0x28091c940>{number = 1, name = main}.
[...]

PR #3746, opened 2022-06-01
Simplify SuggestedSearchesModel

darinwf:
Adds reload logging.

Issue #3651

### How was this tested?
- [x] I tested this manually


darinwf:
Previous logic would call reload too many times. This simplifies matters by having the `SuggestedSearchesModel` be a `StateObject` for the `SuggestedSearchesView`, and then that view is responsible for calling reload from its `onAppear`.

Also moves the addition of the example searches into `reload`. That way the view can just directly consume the `@Published` array of `SuggestedSearch` objects.

Finally added some logging to help catch missing reloads in the field.

I suspect that making this code more normal should resolve whatever issue we were having with SwiftUI not noticing when `suggestions()` would return a new value or whatever other issue may have been going on to cause the bug.

darinwf:
Yeah that first entry is the top "frecency" result. Maybe it does not decay easily. In my testing, it was another search query that occupies the top-slot, so I think it is probably working as it did before.

Issue #3734, opened 2022-05-31
Weird jump after tab animation 


yasminyin96:
## Expected behavior

No weird jump.

## Actual behavior

Weird jump after tab animation 

## Steps to reproduce the problem

1. Open tab
2. Click Show Tabs button

## Screenshots or recordings
Incognito Mode

https://user-images.githubusercontent.com/104786010/171262594-784e4285-fb85-4a77-b84b-52991d04b7ff.mov

Normal Tabs

https://user-images.githubusercontent.com/37356984/171258746-fafe2b27-fc1e-4919-935e-6279122285f1.mp4

https://user-images.githubusercontent.com/104786010/171262661-790c11e6-52d8-4298-8499-e8b59dc40e30.mov

darinwf:
Looks like this bug can be closed now?

Thanks for fixing this issue! I saw it too and was about to file a bug for it :-)

PR #3747, opened 2022-06-01
Fix weird jump after tab animation

yasminyin96:
* Fix weird jump after tab animation

### How was this tested?
- [X] I tested this manually

https://user-images.githubusercontent.com/104786010/171488931-ef358244-b7f1-4e3f-a42d-a4d67426023b.mov

## Issues addressed
* https://github.com/neevaco/neeva-ios/issues/3734


Client/Frontend/Browser/Card/Card.swift
```diff
@@ -231,6 +231,7 @@ struct Card<Details>: View where Details: CardDetails {
                                 .padding(1.5)
                                 .padding(tabDetails.isSelected ? 0 : -1.5)
                                 .contextMenu(menuItems: tabDetails.contextMenu)
+                                .padding(tabDetails.isSelected ? -1.5 : 0)
```
evancrow, line 234: Isn't this the exact opposite of the padding above? Does this add the extra padding back to the `ContextMenu`.

yasminyin96: I think this padding is added to the button.
`
    button
         .if(!animate) { view in
                view
                    .padding
                      ...
`

yasminyin96: Line 231, it adds padding 1.5 to all tab grids.
Line 232, it changes the padding of unselected tab grid to the original padding in order to remove the unnecessary edge when open the contextMenu.
Line 234, it changes the padding of selected tab to the original padding.

evancrow: Makes sense, please add a comment though for future reference :)

PR #3750, opened 2022-06-01
Minor code simplification

darinwf:
Just check tabManager.selectedTab instead of searching for the selected tab.

Issue #3745

### How was this tested?
I'm actually not quite sure how to test this scenario. I'm not sure when the grid would be hidden and the incognito state would not be that of the selected tab. I need to do some archaeology maybe to understand why this check was originally added.


jedireza (approved):
Code changes look reasonable and I didn't notice any animation issues.

Issue #3756, opened 2022-06-01
Webview keeps reloading on a website


chassu:
On main, I saw the webview keeps reloading on this site: https://www.cnbc.com/2022/05/23/ebay-enters-nft-business-with-assist-from-hockey-icon-wayne-gretzky-.html and the app gets slower as it keeps reloading.

I don't think this is archived tabs related.

https://user-images.githubusercontent.com/87332917/171519733-bd8679f2-63ef-43f3-bee0-6f547c5e1c7b.mp4

chung-lai:
This could be cookie cutter related. It happens when I enable the feature in `1.41.4`. cc @evancrow . It stops reloading when I turn off cookie cutter.

evancrow:
@jlynch630 Any thoughts?

evancrow:
Think I remember something about `flagSite`?

PR #3758, opened 2022-06-02
Flag sites

evancrow:
Sets up support for `flagSite` from the Cookie Cutter Engine, which prevents a reload loop. Because `flagSite` is often not called, which results in a reload loop, a domain will also be flagged after a cookie was handled. All flags are reset when user preferences change.

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #3756


jlynch630 (approved):

chung-lai (approved):
LGTM. I tried out the branch and it does seem to reload the page once (may be it adds to the flagged site only after the page is loaded) Assume this doesn't happen often, we should be okay to ship this :)

Thanks!

Issue #3488, opened 2022-05-01
ArchivedTabsView Placement Needs Improvement


evancrow:
It does not stick to the bottom when there aren't enough tabs to push it down:
<img src=https://user-images.githubusercontent.com/37356984/166159185-93593a96-cd97-485d-8539-45cdde6a778a.png width = 300>

Should reposition the `EmptyCardGridView` to sit in between the `ArchivedTabsView` and the `TopBar` since it looks odd right now:
<img src=https://user-images.githubusercontent.com/37356984/166159205-0f067444-f979-49c2-aeab-2addc9ea147a.png width = 300>

Maybe when there are no tabs, it should be in a `VStack` with the `EmptyCardGridView`?

cc @chassu

chung-lai:
@yasminyin96, may be this is an interesting one to look at

PR #3764, opened 2022-06-02
Stick ArchivedTabsView to bottom

yasminyin96:
* Stick ArchivedTabsView to bottom

### How was this tested?
- [X] I tested this manually

https://user-images.githubusercontent.com/104786010/171741154-e0f705b9-bde5-48f1-8b2a-0e020017885b.mov

iPad
Screen Shot 2022-06-02 at 2 28 43 PM

Disabled ArchivedTabsView
Screen Shot 2022-06-02 at 2 27 24 PM

## Issues addressed
* https://github.com/neevaco/neeva-ios/issues/3488


darinwf:
Not quite sure I understand how this change solves the issue. One thing I noticed is that things don't look right in landscape mode:

Simulator Screen Shot - iPhone 13 - 2022-06-02 at 14 56 04

yasminyin96:
> Not quite sure I understand how this change solves the issue. One thing I noticed is that things don't look right in landscape mode:
> 
> Simulator Screen Shot - iPhone 13 - 2022-06-02 at 14 56 04


I thought that if tab row is less than 2, the archivedview will stick to the bottom. But it seems not work for landscape. I will take a look.
I found that this issue also exists on main with no tab.
IMG_3920

chung-lai:
> > Not quite sure I understand how this change solves the issue. One thing I noticed is that things don't look right in landscape mode:
> > Simulator Screen Shot - iPhone 13 - 2022-06-02 at 14 56 04
> 
> I thought that if tab row is less than 2, the archivedview will stick to the bottom. But it seems not work for landscape. I will take a look. I found that this issue also exists on main with no tab. IMG_3920

fyi @chassu about the issue on main with no tab in archived tabs

yasminyin96:
> Test drove; makes sense. We can solve the landscape issue in a separate PR since it isn't introduced by this PR.

Got it. I'll create an issue for the landscape issue. Thank you.

PR #3768, opened 2022-06-03
Unblock neevascope preview mode and organize files

EdwardJTL:
- Add new type to represent data from both APIs
- Show error view only when sheet is empty
- Organize and clean up views
- moved data filtering and checking from menu view to view model

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested on:
- recipe (both allowed and not allowed)
- product page
- map entity page
- knowledge card page
- UGC backlink page

Verified that data shows when not signed in (recipe, query terms).


macy-n (approved):

macy-n:
LGTM!

PR #3774, opened 2022-06-03
Add Performance Tests

evancrow:
Added tests:
- Test performance of opening and closing CardGrid
- Test performance of deleting all tabs

The tests work just have to figure out how to use the results and pass/fail...


evancrow:
Closing this because there is no way to actually use the performance measurement in a CI test. Values are just printed to console and can't be used in a CI pass/fail environment.

PR #3776, opened 2022-06-04
Postpone NeevaScope Promos until page finishes loading

EdwardJTL:
- Add cached state management struct for UGC promo
  - mapping between raw url and cached UGC promo state
  - each state has two `Bool`s, one for promo indicator and one for bubble

CheatsheetPromoModel holds on to a storage value and two `@Published` display states
- Two (and a half) combine pipelines to listen to tab changes
  - One pipeline listens to all URLs and instructs the state storage to check bloom filter [synced on background]
    - only update state if url is not already in cache
    - only place where `CanonicalURL` routine is called
    - if bloom filter is ready, store state based on hit result
  - Another pipeline listens to selected tab [on main]
    - only respond when tab changes
    - resets display states
    - attach a new pipeline to each tab [synced on background]
      - this pipeline only responds when tab `isLoading` is false
      - it resets display states if url is bad
      - if should show try cheatsheet promo, set state
      - else, reads state from storage
- A set of methods to respond to user input 
- A timer method to auto clear the UGC promo
- Also removed the dimming background for UGC indicator promo

This also affects when the default promo is shown. 

~The problem with popover appearing too early still exists. even though the popover is defined on the Neeva button, it can appear before the button itself is on screen, and then it triggers on disappear immediately and clears the state.~
^ Current workaround is to present promo with a delay

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
1. Tested that old onboarding and recipe promo still show.
2. Tested by going to a url that is known to redirect and cause flashes
3. Tested switching b/w tabs with/without indicators, and opening and closing tab switch without switching tab, and state being remembered.

*Note* in video 1 there is a bubble on neeva icon on SRP. this is a bug and its been resolved after the recording.
*Note* the animation in the popover is added by iOS 16. looks pretty tho

| 1 | 2 | 3 |
|---|---|---|
| <video src="https://user-images.githubusercontent.com/44557675/172561501-3ee915c2-6216-4f32-a220-926c8286b6f7.MP4"> | <video src="https://user-images.githubusercontent.com/44557675/172561492-40f07d23-b72f-4db6-9874-cf4e3871979b.MP4"> | <video src="https://user-images.githubusercontent.com/44557675/172561212-aba291e8-ff43-4cf7-81b5-4a1f75691522.MP4"> |


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -133,6 +100,59 @@ public class CheatsheetPromoModel: ObservableObject {
     }
 
     // MARK: - Private Functions
+    private func subscribe(to tab: Tab) {
+        self.tabEventSubcription = Publishers.CombineLatest3(
+            Defaults.publisher(.showTryCheatsheetPopover),
+            tab.$url,
```
darinwf, line 106: I'm a little nervous about accessing properties of `Tab` on a background thread. I think this function should only be called on the main thread.

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -133,6 +100,59 @@ public class CheatsheetPromoModel: ObservableObject {
...
+            if !Defaults[.seenTryCheatsheetPopoverOnRecipe],
+                DomainAllowList.isRecipeAllowed(url: url)
+            {
+                return .tryCheatsheet
+            }
+
+            // else show popover if seen SRP intro screen
+            if showPopover.newValue,
+                // cheatsheet is not used on NeevaDomain
+                !NeevaConstants.isInNeevaDomain(url),
```
darinwf, line 127: Not likely to happen, but what if `NeevaConstants.appHost` is changed while this code is running on a background thread?

darinwf (commented):
I have some concerns about the multi-threading here....

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
EdwardJTL: I don't know if it is unsafe to get the projected value on a background thread. Although i suppose if making the combine chain is not too expensive, this can be on the main thread

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -133,6 +100,59 @@ public class CheatsheetPromoModel: ObservableObject {
...
+        .receive(on: Self.queue)
+        .filter { _, _, isLoading in
+            return !isLoading
+        }
+        .map { showPopover, url, _ -> PromoType? in
+            guard let url = url else {
+                return nil
+            }
+
+            if !Defaults[.seenTryCheatsheetPopoverOnRecipe],
```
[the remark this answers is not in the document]
EdwardJTL: I believe so. `Defaults` is a wrapper around `UserDefaults`, which is thread safe

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
EdwardJTL: I confirmed that subscribing to the projected value is not strictly thread safe and is best done on the main actor

Client/Frontend/Overlay/Content/Cheatsheet/Promo/CheatsheetPromoModel.swift
```diff
@@ -28,74 +90,227 @@ public class CheatsheetPromoModel: ObservableObject {
...
+    /// Subscribe to tabManager's selectedTabPublisher
+    func subscribe(to tabManager: TabManager) {
+        if Defaults[.useCheatsheetBloomFilters] {
+            selectedURLSubscription = tabManager.selectedTabURLPublisher
+                .receive(on: Self.queue)
+                .removeDuplicates()
+                .compactMap { url -> URL? in
+                    guard let url = url,
+                        !(url.scheme == InternalURL.scheme),
+                        !NeevaConstants.isInNeevaDomain(url)
```
darinwf, line 119: note: `isInNeevaDomain` checks the mutable `NeevaConstants.appHost` ... not that it is likely for that value to be modified on the main thread while this code is running, and it would only be modified by us during development, but it is a potential data race. if there is an easy way to avoid it, we should.

An easy fix would be to just move the `.receive(on: Self.queue)` down to just before calling `stateStorage.provision(url:)`.

darinwf (approved):

PR #3782, opened 2022-06-06
Move Where isDisplaying is Set [IntroView]

evancrow:
This was originally set up in #3707, but `onAppear` would sometimes run too late so we would still end up with the `IntroView` loop. Moving it to before we call the actual view to display is "100%" reliable.


chung-lai (approved):
🚀  ship it!

Issue #3766, opened 2022-06-03
Native app link hand-off can leave an empty tab around


jedireza:
## Screenshots or recordings

https://user-images.githubusercontent.com/979929/171761459-7332171d-9da9-4390-9ef5-803645dee76c.MP4

evancrow:
Looked a bit into this in the past, will try and finish that

PR #3793, opened 2022-06-06
Close Blank Tabs on App Foreground

evancrow:
Whenever the app is backgrounded, the app will now quietly close tabs whose URL is `about:blank`.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/172252355-d792b762-d35e-4744-bdef-ce5751b3ff95.MP4

## Issues addressed
Fixed: #3766


Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -190,4 +190,10 @@ extension TabManager {
             }
         }
     }
+
+    // MARK: - Blank Tabs
+    /// Removes and closes blank tabs opened from `target=_blank` or `about:blank`.
```
jedireza, line 195:
```suggestion
    /// Removes any tabs with the location `about:blank`. Seen when clicking web links that open native apps.
```

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -811,13 +811,6 @@ class BrowserViewController: UIViewController, ModalPresenter {
         )
     }
 
-    func openBlankNewTab(isIncognito: Bool = false) {
```
jedireza, line 814: Found some dead code I presume.

jedireza (approved):
Nice to see this approach worked out. Code looks good. I left a suggestion for the comment line; your call. 🚀

Client/Frontend/Browser/BrowserViewController.swift
evancrow: That’s right!

EdwardJTL (commented):
as a thought: how do we feel about doing the clean up on app enter foreground? or before making the system call to open an external app?
bringing this up because I was under the impression that we have limited time/resources to save tabs when app is going into background

evancrow:
Interesting thought. I'm not sure how much of an impact this will have, but I think that making the move to willEnterForeground could be better. Will try it out and report back!

evancrow:
@EdwardJTL Moved it to `willEnterForeground`. Tried it with 600+ tabs and performance is good!

Issue #3785, opened 2022-06-06
Drag-n-drop does not work when time based tab switcher is enabled / archived tabs view is enabled


darinwf:
TestFlight 1.44.1 w/ "Enable time based tab switcher" and "Enable archivedTabsView in switcher" options enabled.

Unable to drag-n-drop reorder tabs:
https://user-images.githubusercontent.com/79867437/172213464-1762d88f-d473-4639-968f-2dcd88e1bd24.MP4

FYI @chassu

evancrow:
@chassu Can this be closed?

PR #3796, opened 2022-06-06
fix time based switcher drag and drop + refactoring

chassu:
Since `activeTabs` is now a stored property, `CardModel.OnDataUpdated()` will still hold the old tabs array, which makes drag and drop look like it wasn't working. The operation actually still took effect, but would have to wait until an add/remove operation to be visible.

The proposed solution here is to update activeTabs if tabs are rearranged.

### How was this tested?
- [x] I tested this manually

before: drag and drop wasn't working

after: 

https://user-images.githubusercontent.com/87332917/172266416-3a94b450-6f54-46b1-8f44-bd68c7467e1e.mp4

## Issues addressed
#3785


evancrow:
Can you drag between time sections?

chassu:
I don't think we allow dragging between time sections by design. That said, with how `buildRows` filters the tabs, it's not possible to drag a tab across sections because even if the index w.r.s. to the tabs array changes, its `lastExecutedTime` remains the same.

evancrow (approved):
Makes sense. This LGTM, just one thing to address.

PR #3797, opened 2022-06-07
Remove experiment arm for default browser interstitial

chung-lai:
Decided not to launch the experiment as it does not provide any positive outcome (and may regress some metrics). We will revisit this later

@hwiatneeva @darinwf @bikramneeva fyi

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


jedireza (approved):
Didn't test drive. Code looks good.

PR #3798, opened 2022-06-07
Set Cookie Popup to 1 if Flagged

evancrow:
If Cookie Popup is already flagged, then we should just set the notices to 1 instead of letting it stay at 0. As discussed in Magpie Weekly.


chung-lai (approved):
LGTM

PR #3451, opened 2022-04-26
Add remind me later in default browser interstitial

chung-lai:
* Change "Skip for now" to "Remind me Later" in default browser interstitial
* Show system notification permission dialog when user taps "Skip for now"
* Add this feature as experiment
* Add default browser onboarding notification **3** days after user accepts notification dialog
* Disable notification permission prompt after user finish tour flow
* Add open default browser settings deep link
* Not to register notification token if user is not signed in

@hwiatneeva, would love copy on the notification, here is what I have right now:

**Title:** Make Neeva your Default Browser
**Message:** Block invasive trackers across the Web. Open links safely with blazing fast browsing.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests
- [ ] Existing automated tests are enough to cover my changes

### Screenshots and screen recordings

Remind me later button flow:

https://user-images.githubusercontent.com/87332578/165361815-9531ea0b-4804-426e-9deb-db5a5842414a.mov

Default Browser Notification:

https://user-images.githubusercontent.com/87332578/165361850-8455ecd0-3546-441c-8bb4-741a7d838949.mov

Keep interstitial after open settings:

https://user-images.githubusercontent.com/87332578/165361929-4e4ac38e-5e13-4af4-ba55-7cb5201e348b.mov

Dismiss:

https://user-images.githubusercontent.com/87332578/165361945-f4be21ef-82da-442e-9290-8ac6011110f5.mov

## Issues addressed
* Link(s) to Github issue(s) (if applicable)

PR #3800, opened 2022-06-07
fix webui message helper regression

jedireza:
Fixes a regression where we don't process any non-tour `webui` message.

In https://github.com/neevaco/neeva-ios/pull/3451, we added back the tour step checking. One thing I didn't dive deeper into was if the `TourStep` enum didn't match, the `guard` short-circuits the remainder of the handler.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

N/A

## Issues addressed
N/A


chung-lai (approved):
Good find! LGTM.

Wonder what regression does it cause? What are the non-tour messages we currently process?

jedireza:
Sign up: https://github.com/neevaco/neeva-ios/pull/3418

PR #3802, opened 2022-06-07
Burak/thumbnail url for spaces

buracules:
- Add URL support for spaces thumbnail

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Shared/NeevaFeatureFlags.swift
```diff
@@ -100,6 +101,7 @@ public class NeevaFeatureFlags {
             case .enableSpaceDigestDeeplink: return "ios.enable_space_digest_deep_link"
             case .enableSpaceDigestCard: return "ios.enable_space_digest_card"
             case .enableBacklink: return "backend.enable_backlinks_cheatsheet"
+            case .enableCloudinaryThumbnails: return "space.enable_cloudinary_thumbnails"
```
yozuysal, line 104: I dont think we will need this?

yozuysal (approved):
interesting that WebImage always had support for base64 strings and we never utilized this. lgtm!

Shared/NeevaFeatureFlags.swift
buracules: Right we wont use this after m1 changes i will remove this now

PR #3810, opened 2022-06-08
Launch control arm of the default browser experiment

chung-lai:
Completely go back to the control arm of the experiment so we don't restore the interstitial screen. Though we can argue restoring interstitial makes sense, we should either completely go to control or variant and we have decided to go back to control at this moment. cc @bikramneeva 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -545,11 +545,6 @@ class BrowserViewController: UIViewController, ModalPresenter {
 
                 _ = NeevaExperiment.startExperiment(for: .promoCardTypeAfterFirstRun)
                 NeevaExperiment.logStartExperiment(for: .promoCardTypeAfterFirstRun)
-            } else if let didDismiss = Defaults[.didDismissDefaultBrowserInterstitial],
-                !didDismiss
-                    && !Defaults[.didFirstNavigation]
-            {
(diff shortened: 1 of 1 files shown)


chung-lai:
> What is the reason for making this change? Are we seeing a dip in people making the first navigation?

Just checked, the first navigation is pretty similar between control and variant. However, this restoring logic was part of the last experiment, there is a slight drop (probably not stats sig) in default browser conversion and drop on notification prompt accept rate.

This change may probably unrelate to the result that we saw but felt like we should either go to control or variant. we should not mix the experiment arms which may introduce some unknown factors to the metrics.

chung-lai:
> I see... but isn't this code part of enabling the user to see the default browser educational UI again if they need to come back to the app having gone to OS settings? The app might get killed in the background, so this is about restoring the state to what it was so the user can still see the default browser educational UI. Right?

That's right. Unfortunately the interstitial restoring from background kill was part of the experiment variant arm so it's hard to rule out if that change has anything to do with the result. When I was testing the app this morning, I wonder if user kills the app on purpose to dismiss the screen and continue their flow. However, first navigation rate seem to suggest that's not the case.

May be I just overthink about this, we can treat this more like a bug fix and leave this restore logic in.

bikramneeva:
I would agree. We should just keep the restore logic in as bug fix

Issue #3805, opened 2022-06-07
Regression: Space filter requires double tap


jonathangarelick:
A bit hard to repro, but I experienced this on both `main` and the app store release.

https://user-images.githubusercontent.com/41495718/172495438-78689cd1-cbd9-4478-aac2-b3291b9dc763.mp4

cc @evancrow

evancrow:
What happened to my fix with the `DispatchQueue`? Or are you trying to replace the `GroupedButton`s with an actual `Button`?

PR #3811, opened 2022-06-08
Fix SpaceFilter

evancrow:
Fixed this by wrapping both calls in a `DispatchQueue`. I'm not exactly sure why the `GroupedRowButtonView` operates differently from a regular `Button`,  but when passing UI changes through the action completion handler you have to ensure it's run on the main thread. I tried replacing the `GroupedRowButtonView` with a SwiftUI generic `Button` and the `DispatchQueue` wasn't necessary (even though both are initialized the same way).

cc @darinwf for any thoughts.

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #3805


chassu (approved):
LGTM

Issue #3759, opened 2022-06-02
Tab Group Overlap iPad CardGrid


evancrow:
When two Tab Groups are side by side, they overlap on iPad:

![image](https://user-images.githubusercontent.com/37356984/171550518-45d440f6-fb31-462b-8eec-a94aa1d3924c.png)

FYI @chassu

PR #3814, opened 2022-06-08
TabGroup UI + ScrollTo Improvements

evancrow:
- Fixed excess padding on sides of tab groups
- Fixed overlap of tab groups
- `ScrollTo` in `CardGrid` will now center the `selectedTab` (also prevents the top being cutoff)
- Creates some new helper functions for navigating arrays
- Some refactoring

### How was this tested?
- [x] I tested this manually

Before:
 

After:
 

https://user-images.githubusercontent.com/37356984/172725029-77cfd528-d638-4adb-a428-1d6435bd1f43.mp4 

https://user-images.githubusercontent.com/37356984/172725040-13fa011b-32a5-42fe-ace7-161f53d2181a.mp4

## Issues addressed
Fixed: #3759


chassu:
Are we leaving gaps between groups on the same row? I remember before the regression happened, a small space was left out in between two groups.

chassu:
Thanks for the `scrollTo` fix!

evancrow:
> Are we leaving gaps between groups on the same row? I remember before the regression happened, a small space was left out between two groups.

Oh sure, will adjust the padding

evancrow:
If I try adding extra padding in, it looks kinda odd:
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-06-08 at 16 26 01

Think it's best to not have the extra padding in between two tab groups because that way the cards are still in line with the others.

chassu (approved):
LGTM

PR #3815, opened 2022-06-08
force tab data update in addTabForURLs

chassu:
Due to how time-based switcher works, we need to ensure tab data are updated when `addTabsForURLs` are called. This resolves an issue when archived tab is enabled, creating 100 tabs won't immediately be visible.


Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -29,10 +29,10 @@ extension TabManager {
             for tab in newTabs {
                 tab.rootUUID = rootUUID
             }
-
-            self.updateAllTabDataAndSendNotifications(notify: false)
         }
 
+        self.updateAllTabDataAndSendNotifications(notify: false)
```
evancrow, line 34: Is `tabsUpdatedPublisher.send()` on line 42 redundant now?

chassu: No, I don't think so. It's still needed to invoke `cardModel.onDataUpdated` so it grabs the updated tab data.

evancrow (approved):

Issue #3813, opened 2022-06-08
card transition animation glitches on long taps


chassu:
https://user-images.githubusercontent.com/87332917/172719421-c8159703-3a69-4de9-ac04-115c9df73c8d.mp4

I was trying to fix this yesterday. My theory is that there's a small window when the gesture is considered as a tap but also long enough to trigger the context menu. If the user releases the finger before the context menu is fully displayed, then there will be an overlay card on top of the transition card.

I realized it's actually pretty easy to repro.

jedireza:
Moved to ClickUp 🔒 https://app.clickup.com/t/3w35bd7

PR #3819, opened 2022-06-09
add animation to scaleEffect on tab cards

chassu:
### How was this tested?
- [x] I tested this manually

before:

https://user-images.githubusercontent.com/87332917/172738867-683909ef-60dd-4b8c-8dad-efca6b04b243.mp4

after:

https://user-images.githubusercontent.com/87332917/172952429-3e7c830d-d8d0-47bd-b2f2-a3f505e374de.mp4

@darinwf do you want to try it out and see how it feels? I wonder if this makes https://github.com/neevaco/neeva-ios/issues/3813 more noticeable.


chassu:
@evancrow lmk if you like this change

evancrow:
Updated animation, with grid toolbar padding fixed:

https://user-images.githubusercontent.com/37356984/190305091-6fbb08dc-77e7-4938-8570-e7d99ab6fdd9.mp4

Client/Frontend/Browser/BrowserTopBarView.swift
```diff
@@ -21,8 +21,10 @@ struct BrowserTopBarView: View {
     @ViewBuilder var switcherTopBar: some View {
         if chromeModel.inlineToolbar {
             SwitcherToolbarView(top: true)
+                .padding(.bottom, 7.5)
         } else {
             GridPicker()
+                .padding(.bottom, 4.5)
```
darinwf, line 27: nit: How about moving these magic constants to `BrowserTopBarViewUX`. Bonus points if they can be computed relative to other constants, but OK to just hard code them in this file.

darinwf (commented):
@chassu's original fix was to add an animation to the `scaleEffect` here:

Screen Shot 2022-09-14 at 9 48 30 PM

I don't see that in the updates you posted @evancrow, and when I try out the branch, it looks like the card still snaps to the 95% size when pressing it.

Or maybe I'm missing something.. I'm not sure I understand the conditional animation you have on the CardTransitionModifier now.

darinwf:
@evancrow - It's a good fix for the padding. Maybe take that as a separate PR?

evancrow:
Sure, it's a small change. I included it because it helps with the animation.

evancrow:
> @chassu's original fix was to add an animation to the `scaleEffect` here:
> 
> 
> 
> Screen Shot 2022-09-14 at 9 48 30 PM
> 
> 
> 
> I don't see that in the updates you posted @evancrow, and when I try out the branch, it looks like the card still snaps to the 95% size when pressing it.
> 
> 
> 
> Or maybe I'm missing something.. I'm not sure I understand the conditional animation you have on the CardTransitionModifier now.

Ah okay, I couldn't really tell what the fix or the problem was :). That code must have changed in the merge so I'll bring it back. 

My fix just made the overall animation smoother and the condition I added prevented a flickering of the card when opening the card grid (caused by the animation, not an existing but).

chassu:
@evancrow looks like you're trying to fix the top padding when animating the tab card to full screen. This original PR is trying to animate the card when it's pressed down so it doesn't snap into 95% size. 

You can bring the change back, but I just want to point out that the animation fix in the original PR could cause funky interactions with the long press menu.

evancrow:
Ohhh I see now. I discussed with Darin earlier, I thought the regular transition was snapping to much. Okay now I have better idea of how to continue the fix, thanks.

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
```diff
@@ -52,7 +52,6 @@ struct CardTransitionModifier<Details: CardDetails>: ViewModifier {
         content
             .zIndex(details.isSelected ? 1 : 0)
             .opacity(details.isSelected && cardTransitionModel.state != .hidden ? 0 : 1)
-            .animation(nil)
```
darinwf, line 55: Recommend adding this back. Think we need it.

darinwf (approved):
LGTM with one fix. I think we can revisit some of the other aspects of this PR that Charles was working on separately.

PR #3827, opened 2022-06-10
Fix Delay Create WebView

evancrow:
Because `updateWebViewForSelectedTab` was also being called from `sendSelectTabNotifications` with no respect to `shouldCreateWebViewUponSelect`, the code to delay creating the `WebView` until after the select tab animation was useless basically. This moves the call to `updateWebViewForSelectedTab` to one place.

### How was this tested?
- [x] I tested this manually


chassu (approved):
LGTM

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -256,10 +256,6 @@ class TabManager: NSObject {
             return
         }
 
-        if selectedTab.shouldCreateWebViewUponSelect {
-            updateWebViewForSelectedTab(notify: false)
-        }
-
```
darinwf, line 262: If `notify` is false, then will we fail to create the `WKWebView`? Should there be special handling for when `notify` is false?

darinwf: `notify` false just means we should not send notifications. it doesn't mean we should skip doing other important work. probably important to make sure the state of the `Tab` remains consistent, so if a tab is selected, it should probably have a `WebView`, right? Anyways, `notify` params are intended to help suppress duplicate, spammy notifications. We don't have to worry about spamming creation of the `WebView` since it is cached and will only be created the first time.

darinwf (approved):
LGTM thanks!

Issue #3823, opened 2022-06-10
Add "Yesterday" section to time-line tab switcher


darinwf:
Add "Yesterday" section to time-line tab switcher

This will be helpful so that the tabs you had open Yesterday do not disappear into the "Past 7 days" section when you wake up the next morning.

See https://app.clickup.com/t/2w32u5r for reference

@chassu FYI

PR #3835, opened 2022-06-13
add yesterday section to time based switcher

chassu:
### How was this tested?
- [x] I tested this manually

## Issues addressed
#3823


Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -869,10 +873,16 @@ public func isLastExecutedTimeInTimeFilter(_ lastExecutedTime: Timestamp, _ byTi
             byAdding: .second, value: -15, to: Date())
         : Calendar.current.date(
             byAdding: .day, value: -1, to: Date())
-    let minusOneWeekToCurrentDate =
+    let minusTueDaysToCurrentDate =
```
evancrow, line 876: nit: Two (instead of Tue)?

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -869,10 +873,16 @@ public func isLastExecutedTimeInTimeFilter(_ lastExecutedTime: Timestamp, _ byTi
...
         FeatureFlag[.shortenTimeThresholdForArchivingTabs]
         ? Calendar.current.date(
             byAdding: .second, value: -30, to: Date())
+        : Calendar.current.date(
```
evancrow, line 880: In PR #3824, which is still under review, I created a new helper function that you can use to simplify this create date code. Once that merges you will be able to refactor a lot of this down to `Date.getDate(dayOffset: -(days))`.

No action item here, just something for future cleanup

evancrow (approved):
LGTM, one nit

Client/Frontend/Browser/Tabs/Tab.swift
chassu: can't believe I did that 😂

PR #3836, opened 2022-06-13
Fix archived tab button placement

yasminyin96:
* Fix archived tab button placement
(set TabCardContainer `minHeight` to `height of TabCardContainer - height of archivedtab button`
instead of based on rows of tabs)

### How was this tested?
- [X] I tested this manually

### Screenshot
Screen Shot 2022-06-13 at 11 42 28 AM

Screen Shot 2022-06-13 at 11 42 56 AM

Screen Shot 2022-06-13 at 11 37 26 AM

https://user-images.githubusercontent.com/104786010/173423307-5e56cc1e-efe6-4912-a5f4-ff2d9e2a7c33.mov

Screen Shot 2022-06-13 at 11 40 34 AM

https://user-images.githubusercontent.com/104786010/173423439-4578421a-3851-4218-a7f3-11b7942555f1.mov


Client/Frontend/Browser/Card/Grid/CardsContainer.swift
```diff
@@ -123,6 +93,7 @@ struct TabGridContainer: View {
                 }
             )
         }
+        .frame(minHeight: geom.size.height - 120, maxHeight: .infinity, alignment: .top)
```
chassu, line 96: Can you define a constant for this? Maybe add it to `CardUX` and you can do something like `geom.size.height - CardUX.ArchivedTabsViewHeight`

chassu (approved):
LGTM!

Client/Frontend/Browser/Card/Grid/CardsContainer.swift
yasminyin96: Will do

Client/Frontend/Browser/Card/Card.swift
```diff
@@ -19,6 +19,7 @@ enum CardUX {
     static let CloseButtonSize: CGFloat = 24
     static let HeaderSize: CGFloat = ButtonSize + 1
     static let CardHeight: CGFloat = 174
+    static let ArchivedTabsViewHeight: CGFloat = 80
```
chassu, line 22: Actually, I think it might be better to just declare this inside the CardsContainer class. The height of the archivedTabsView is not really related to CardUX imo.

yasminyin96: Make sense.

PR #3842, opened 2022-06-13
Open to ZeroQuery After Being In-Active 

evancrow:
If the user opens the app after they haven't opened it within the hour, we will show the ZeroQuery UI (unless the app was opened from a link). Behind a flag.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/173435219-0512a6b3-dc35-4c4b-9c8d-539ce0db4ed7.MP4


Client/Application/SceneDelegate.swift
```diff
@@ -158,6 +132,53 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+                break
+            }
+        }
+
+        DispatchQueue.main.async { [self] in
+            // Show the ZeroQuery UI if the user hasn't opened the app within the hour.
+            if FeatureFlag[.openZeroQueryAfterLongDuration],
+                sceneLastOpenedTime.hoursBetweenDate(toDate: Date()) > 1
+            {
+                bvc.chromeModel.setEditingLocation(to: true)
```
darinwf, line 175: Does this work if the browser is currently showing the tab switcher or a spaces detail view?

Client/Application/SceneDelegate.swift
```diff
@@ -20,6 +20,9 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     var bvc: BrowserViewController!
     private var geigerCounter: KMCGeigerCounter?
 
+    @Default(.sceneLastOpenedTime) private var sceneLastOpenedTime
+    private var URLHandledOnLaunch = false
```
darinwf, line 24: style-nit: start variable name with lower case characters... `urlHandledOnLaunch` instead

Client/Application/SceneDelegate.swift
evancrow: Works from both but the `SpaceDetailView` is kinda weird. Thinking that we should only open `ZeroQueryView` if the user is in the Tab or Tab CardGrid.

darinwf: What is weird about the `SpaceDetailView` case? Can you attach a screen capture of that case?

evancrow: https://user-images.githubusercontent.com/37356984/175105418-82b17ad3-e7b0-4cd4-bc32-2012467ddc0f.mp4

It's actually not that weird, it just erases the `SpaceDetailView` state

darinwf: OK, I see. It would be nice if we didn't have that transition animation away from the `SpaceDetailView`. Maybe if we determine at startup that we want to just show the zero-query, we should not bother restoring to the `SpaceDetailView`?

evancrow: Made that change, also did the same for when restoring to the Spaces CardGrid

Client/Application/SceneDelegate.swift
darinwf: looks like you might have missed this

Client/Application/SceneDelegate.swift
```diff
@@ -158,6 +132,59 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+        // Restoring SceneUIState.
+        if FeatureFlag[.restoreAppUI] {
+            let sceneUIState = SceneUIState(rawValue: scenePreviousUIState)
+
+            switch sceneUIState {
+            case .cardGrid(let switcherState, let isIncognito):
+                switch switcherState {
+                case .tabs:
+                    bvc.browserModel.showGridWithNoAnimation()
+                    bvc.browserModel.gridModel.switchToTabs(incognito: isIncognito)
```
darinwf, line 162: why do we want to call this if shouldSetEditingLocationToTrue is set? is it about preserving incognito mode?

darinwf: Please add a comment about this since it is not super obvious. Thanks!

PR #3844, opened 2022-06-13
add confirmation prompt for clearing archived tabs

chassu:
Add a confirmation prompt when the user clears archived tabs.

https://user-images.githubusercontent.com/87332917/173468593-4606d377-19b4-46f0-a999-b742b909304f.mp4

## Issues addressed
#3838


Client/Frontend/Browser/Card/ArchivedTabsPanelView.swift
```diff
@@ -38,6 +39,27 @@ struct ArchivedTabsPanelView: View {
...
+                .fixedSize(horizontal: false, vertical: true)
+                .padding(.vertical, 10)
+            Spacer()
+        }
+        .padding(.horizontal, 16)
+        .frame(height: 52)
+    }
+
+    var clearAllArchiveButtonTitle: String {
+        return "Are you sure you want to close all archived tabs?"
```
evancrow, line 56: This does not need to be a computed value, should be a `private let`

evancrow (approved):
LGTM one comment

PR #3869, opened 2022-06-15
fix open setting

chung-lai:
There is a timing issue between when the deep link is handled in notification and the app state. This should be a safe fix but we should look for a better solution to make sure the app state is set up properly before handling the notification deeplink.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/87332578/173922454-72fbdf19-2bd7-4c17-8c53-396461f5eb22.mov


Client/Application/NavigationRouter.swift
```diff
@@ -274,7 +274,9 @@ enum NavigationPath {
         guard let page = page else {
             return
         }
-        bvc.openSettings(openPage: page)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) {
```
jedireza, line 277: Have we tried shorter than 1 second?

chung-lai: i tried 500ms and it seem to work fine. I'll shorten it

Client/Application/NavigationRouter.swift
```diff
@@ -274,7 +274,9 @@ enum NavigationPath {
         guard let page = page else {
             return
         }
-        bvc.openSettings(openPage: page)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
jedireza, line 277:
```suggestion
        // TODO: look for a better solution to make sure the app state is set up properly before handling the notification deep link
        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```

jedireza (approved):
Cool. Thanks for adding the video. I suggest adding a TODO comment and filing a followup issue to track.

PR #3870, opened 2022-06-15
make pinned tabs stay in today section

chassu:
confirmed with @Ria-yu in design open hour that child tabs could expire when the tab group is pinned. (Technically, we don't directly allow pinning of a tab group. If a child tab is pinned, we promote the group to the top.)

### How was this tested?
- [x] I tested this manually

The screen recording is just trying to show that pinning a child tab won't bring other child tabs from archive. It takes a deliberate action to select from ArchivedTabsView to bring tabs back from archives. 

https://user-images.githubusercontent.com/87332917/173904937-ccac2d2d-e513-4c37-893b-8333384305da.mp4


Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -704,7 +704,8 @@ class Tab: NSObject, ObservableObject {
         // The fallback value won't be used. tab.lastExecutedTime is
         // guaranteed to be non-nil in configureTab()
         let lastExecutedTime = lastExecutedTime ?? Date.nowMilliseconds()
-        return isLastExecutedTimeInTimeFilter(lastExecutedTime, byTime)
+        return isPinned
+            ? byTime == .today : isLastExecutedTimeInTimeFilter(lastExecutedTime, byTime)
```
darinwf, line 708: nit: adding the `isPinned` exception inside `wasLastExecuted` means that the name "`wasLastExecuted`" is no longer quite right. Is there a better name for this function that would be inclusive of pinning?

darinwf (approved):

Client/Frontend/Browser/Tabs/Tab.swift
chassu: I can't think of a good name. So I added some comments instead. Is that okay?

darinwf: Well, the comment is not the best because it is just repeats what the code says.

See https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/ for some good advice on writing comments. See rule 1 in particular :)

I see a couple options: 1) rename the function to be more correct or 2) move the pinning check to the caller of this function.

Maybe 2 doesn't work because you have the TabGroup function of the same name and function that calls this function. So renaming is probably the answer.

An option for 1: consider names that express membership or belonging to a time section.

darinwf (approved):
Thanks, that name works!

PR #3884, opened 2022-06-17
Remove refresh calls for Spaces

buracules:
- Remove refresh calls for delete, unfollow, follow and pin space.
- Remove `delete space` button for default spaces
- Some cleanup in SpaceFilterView for preparation to sort feature

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


buracules:
cc @jonathangarelick

Client/Frontend/Browser/Card/Card.swift
```diff
@@ -379,7 +377,9 @@ struct Card<Details>: View where Details: CardDetails {
                     view
                         .padding(1.5)
                         .contextMenu {
-                            removeButton
+                            if !(details.item?.isDefaultSpace ?? false) {
```
buracules, line 380: I can remove this one from this PR if you want @yozuysal

Shared/API/SpaceStore.swift
```diff
@@ -738,8 +759,53 @@ public class SpaceStore: ObservableObject {
...
+    }
+
+    public func pinSpace(with id: String) {
+        if let item = allSpaces.first(where: { $0.id.id == id }) {
+            item.isPinned.toggle()
+            sort(updatedSpaces: [item])
+        }
+    }
+
+    public func sort(by type: SpaceSortState? = nil, updatedSpaces: [Space]? = nil) {
```
buracules, line 784: This is gonna only use dateFormatter for now but i wanna add descended and ascended support as well so there will be another parameter when i add the sort feature @yozuysal

Shared/API/SpaceStore.swift
```diff
@@ -738,8 +759,53 @@ public class SpaceStore: ObservableObject {
...
         return SpaceServiceProvider.shared.pinSpace(spaceId: spaceId, isPinned: !isPinned)
     }
 }
+
+extension SpaceStore {
+    public func deleteSpace(with id: String) {
+        if let index = allSpaces.firstIndex(where: { $0.id.id == id }) {
+            allSpaces.remove(at: index)
+            updatedSpacesFromLastRefresh = allSpaces
+            state = .ready
```
yozuysal, line 773: I think this is the right thing to do but more consistently: The state of the store might already be .ready here so even though you get the ping you need, it is not very consistent.

I think we should add a new state ( we already have refreshing, cancelled, ready). If we were to add a state = mutating_with_local_only_update before we create the request  and then set back to ready when we were all done (both with the mutation request and also the local update) I think it would be more consistent and easier to understand.

So when we query for data we call refresh and it gets set to refreshing and then ready. If we are mutating and updating locally, it gets set to mutating_with_local_only_update and then ready. This is true for all various updates here. This way any View that should be affected by these updates can just declare SpaceStore as an ObservedObject and update accordingly.

yozuysal (commented):
A few comments about structure (all additive stuff).

FYI, another thing to look into is to not need refreshSpace. That can also save us some network calls. We can do that in a follow up PR after this, once we settle down on a setup.

Shared/API/SpaceStore.swift
buracules: I think i missed that one. But i have to store sorted array somewhere i dont wanna sort it everytime i access it. Maybe a  sorted copy for list?

buracules: We can also save users last sort type (if we want the sort feature) and use that type for all until it changes but its discussion for spaces weekly i guess. Also we are gonna need createdDate.

Shared/API/SpaceStore.swift
buracules: > A few comments about structure (all additive stuff).
> 
> FYI, another thing to look into is to not need refreshSpace. That can also save us some network calls. We can do that in a follow up PR after this, once we settle down on a setup.

Are you talking about updateSpace() when updatedSpacesFromLastRefresh count is 1?

yozuysal: look for calls to refreshSpace() , yeah I think it falls under the case where updatedSpacesFromLastRefresh == 1

Client/Frontend/Browser/Card/CardListViews.swift
```diff
@@ -2,16 +2,92 @@
...
 import Shared
 import SwiftUI
 
+class SpaceCardsViewModel: ObservableObject {
+    @ObservedObject var spacesModel: SpaceCardModel
+    @Published var dataSource: [SpaceCardDetails] = []
+    private var subscriber: AnyCancellable?
+
+    init(model: SpaceCardModel) {
+        self.spacesModel = model
```
yozuysal, line 15: do we need this? we just need to setup the subscription afaict and things should just work without needing to keep a circular reference.

Also note, @ObservedObject is only useful in the context of Views. It basically makes sure that whatever object you are observing triggers a View update when a objectWillChange signal is received. To connect models with signals from other models, you need the publisher->subscription setup you are doing below.

Client/Frontend/Browser/Card/CardListViews.swift
```diff
@@ -2,16 +2,92 @@
...
+                    }
+                self.dataSource = spaces.sortSpaces(by: sort).filterSpaces(by: filter)
+            }
+    }
+
+}
+
 struct SpaceCardsView: View {
-    @EnvironmentObject var spacesModel: SpaceCardModel
+    var spacesModel: SpaceCardModel
```
yozuysal, line 32: 🔥. But I think we can go even a step further and not keep the reference at all. It is in the environment so children can reach it OK, we only need it for construction

Client/Frontend/Browser/Card/CardListViews.swift
```diff
@@ -2,16 +2,92 @@
...
-        ForEach(spacesModel.detailsMatchingFilter, id: \.id) { details in
+        ForEach(viewModel.dataSource, id: \.id) { details in
             FittedCard(details: details)
                 .id(details.id)
         }.accessibilityHidden(spacesModel.detailedSpace != nil)
     }
 }
+
+extension MutableCollection where Self == [SpaceCardDetails] {
+    //todo: I will re-write this function with sort feature /Burak
```
yozuysal, line 49: // TODO: (Burak) Rewrite with sort feature

Client/Frontend/Browser/Card/CardListViews.swift
```diff
@@ -2,16 +2,92 @@
...
+        subscriber = spacesModel.$allDetails
+            .combineLatest(spacesModel.$filterState, spacesModel.$sortType)
+            .sink { (arr, filter, sort) in
+                var spaces =
+                    arr
+                    .filter {
+                        NeevaFeatureFlags[.enableSpaceDigestCard]
+                            || $0.id != SpaceStore.dailyDigestID
+                    }
+                self.dataSource = spaces.sortSpaces(by: sort).filterSpaces(by: filter)
```
yozuysal, line 25: filter first, sort later so that you sort less?

buracules: That makes sense 😀

Client/Frontend/Browser/Card/CardListViews.swift
```diff
@@ -2,16 +2,91 @@
...
+        self.spacesModel = spacesModel
+        self.viewModel = SpaceCardsViewModel(spacesModel: spacesModel)
+    }
 
     var body: some View {
-        ForEach(spacesModel.detailsMatchingFilter, id: \.id) { details in
+        ForEach(viewModel.dataSource, id: \.id) { details in
             FittedCard(details: details)
                 .id(details.id)
         }.accessibilityHidden(spacesModel.detailedSpace != nil)
```
yozuysal, line 42: actually, can you check whether we still need this. Go to a detail view with the accessibility setting turned on and try to have it read out loud the toolbar buttons (back button etc). Now that we have moved to a proper NavigationView, this may not be needed anymore. It used to be that without this here, while in the detail view you would still see the space grid cards in the accessibility tree (so you were hearing Space titles being read out loud)

yozuysal (approved):
lgtm % one last thing to check

Client/Frontend/Browser/Card/CardListViews.swift
buracules: For this case it looks ok but i think there is another problem with voice over here. It focuses archived tabs etc. before spaces
https://user-images.githubusercontent.com/15333612/175720551-59675ea8-bb70-466f-ae25-af2eee225784.MP4

yozuysal: hmm, okay. I think this is a separate issue though. Can you file a github issue on this (we need to set tabs as accesibiity hidden while spaces is shown) and then in this PR remove the accessibility hidden and the dependency on spaceCardModel? Then we are good to land

buracules: Filed here #3932

Issue #3882, opened 2022-06-17
Crash trying to dismiss Neeva Scope promo


darinwf:
Happening on 1.45.4:

```
#0	(null) in Swift runtime failure: Unexpectedly found nil while implicitly unwrapping an Optional value ()
#1	0x0000000100674290 in CheatsheetPromoModel.dismissPromo(on:) at /Users/dfisher/src/neeva-ios/Client/Frontend/Overlay/Content/Cheatsheet/Promo/CheatsheetPromoModel.swift:267
#2	0x0000000100674080 in CheatsheetPromoModel.dismissPromo(on:) at /Users/dfisher/src/neeva-ios/Client/Frontend/Overlay/Content/Cheatsheet/Promo/CheatsheetPromoModel.swift:267
#3	0x000000010095aae0 in closure #1 in BrowserBottomBarView.toolbar.getter at /Users/dfisher/src/neeva-ios/Client/Frontend/Browser/BrowserBottomBarView.swift:27
#4	0x000000010087e268 in partial apply for closure #1 in TabToolbarView.neevaButton.getter at /Users/dfisher/src/neeva-ios/Client/Frontend/TabChrome/Toolbar/TabToolbarView.swift:89
#5	(null) in 0x100eec000 ()
#6	(null) in partial apply for implicit closure #2 in implicit closure #1 in WrappedButtonStyle.makeBody(configuration:) ()
#7	(null) in closure #1 in PressableGestureCallbacks.dispatch(phase:state:) ()
#8	(null) in thunk for @escaping @callee_guaranteed () -> () ()
#9	(null) in thunk for @escaping @callee_guaranteed () -> (@out ()) ()
#10	(null) in thunk for @escaping @callee_guaranteed () -> () ()
#11	(null) in static Update.end() ()
#12	(null) in EventBindingManager.send(_:) ()
#13	(null) in specialized EventBindingBridge.send(_:source:) ()
#14	(null) in UIKitGestureRecognizer.send(touches:event:phase:) ()
#15	(null) in @objc UIKitGestureRecognizer.touchesBegan(_:with:) ()
#16	(null) in -[UIGestureRecognizer _componentsEnded:withEvent:] ()
#17	(null) in -[UITouchesEvent _sendEventToGestureRecognizer:] ()
#18	(null) in __47-[UIGestureEnvironment _updateForEvent:window:]_block_invoke ()
#19	(null) in -[UIGestureEnvironment _updateForEvent:window:] ()
[...]

PR #3885, opened 2022-06-17
fix neeva scope crash

macy-n:
fixes https://github.com/neevaco/neeva-ios/issues/3882


EdwardJTL (approved):

darinwf (commented):
Ah, because `stateStorage` is defined as type `PromoStateStorage!`. Perhaps it would be better to just define it as `PromoStateStorage?` so that it will be more clear that it is necessary to treat it as an optional at spots like this. That could be a good way to avoid future bugs like this.

EdwardJTL:
> Ah, because `stateStorage` is defined as type `PromoStateStorage!`. Perhaps it would be better to just define it as `PromoStateStorage?` so that it will be more clear that it is necessary to treat it as an optional at spots like this. That could be a good way to avoid future bugs like this.

I used implicitly unwrapped optional on purpose; Darin we had a conversation before about how we use those to catch errors in the code during alpha. In this case, I wanted to make sure that all accesses to the storage is guarded by the flag. 

I also now realize that even if accesses are guarded by the flag, the flag could still be flipped to `true` after the class has been initialized. I agree that we can change this to a non optional or an `?` type, depending on how we want to roll out this feature

Issue #3645, opened 2022-05-20
Design a more reasonable error for when `suggestd` isn't feeling well


jedireza:
Currently if `suggestd` is borked, we take over the screen. We could probably come up with a more graceful experience. Even just showing one suggest line as an error might be better than taking over the whole screen.

## Screenshots or recordings

<img width="350" src="https://user-images.githubusercontent.com/979929/169617313-3e1badb8-82d1-4e77-bfc4-4c7351d2810a.gif" />

<img width="350" src="https://user-images.githubusercontent.com/979929/169617118-d2bc6d50-068f-45c4-a433-1d9d5c503ae1.png" />

EdwardJTL:
might still be good to send a log when that happens. Right now this goes to user-feedback which is very visible

chung-lai:
We should have some server-side monitoring for this so we can get alert quicker instead of waiting for user feebback.

fyi, @kelvinso, @yifanj0

yozuysal:
when suggestd fails, we should just show history suggestions and silently log? That way the user may still keep going with their task?

jedireza:
Moved to ClickUp 🔒 https://app.clickup.com/t/3w359g4

PR #3888, opened 2022-06-17
Hide Neeva Search when suggestion fails

yasminyin96:
* Hide Neeva Search when suggestion fails

### How was this tested?
- [X] I tested this manually

## Issues addressed
* https://github.com/neevaco/neeva-ios/issues/3645

### DIFF
--- a/Client/Frontend/Suggestions/SuggestionModel.swift
+++ b/Client/Frontend/Suggestions/SuggestionModel.swift
@@ -205,12 +205,11 @@ class SuggestionModel: ObservableObject {
             }
             switch result {
             case .failure(let error):
-                ClientLogger.shared.logCounter(
-                    .SuggestionErrorLoginViewImpression,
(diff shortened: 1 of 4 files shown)


yasminyin96:
Do we need to show offline error view if the suggestion fails due to offline?


With other errors, do we need show the error under the Neeva Search row, or just hide it?

chung-lai:
> Should we show offline error view if the suggestion fails due to offline? 
> 

For the offline case, I think having an offline screen makes sense since user can't do much in the browser if they are offline :)

> With other errors, do we need show the error under the Neeva Search row, or just hide it? 

Having some type of error like "Neeva Search is not available" or other graceful message will help user on questioning why there is no Neeva suggestion. Any thought, @hwiatneeva @yozuysal ?

chung-lai (commented):
Code LGTM, let's see what others think about the open question

chung-lai:
There seems to be more work need to be there for this. Let's close for now and revisit later.

Issue #3854, opened 2022-06-14
Old tabs before new tab switcher is enabled will be archived without URL and favicon


chassu:
Due to the fact that we didn't properly store and read `tab.url` and `tab.title` until #3593. For tabs that the user hasn't interacted with since before the PR is merged, those tabs will be archived without URL and favicon. I feel like there will be plenty of tabs like this, until all the tabs in the tab switcher are created after that PR. Should we fix the UI for those tabs? and what would be a good way?

Without interacting with the tabs:

![Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-13 at 15 44 00](https://user-images.githubusercontent.com/87332917/173671595-6e7ab319-138b-47b9-970c-7d49b593f787.png)

After bringing (some of) them back from archives and getting archived again:

![Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-13 at 15 45 47](https://user-images.githubusercontent.com/87332917/173671613-28a5b44e-c9aa-4079-b516-f7591c6ba892.png)

chassu:
@darinwf curious what you think about this

darinwf:
It would be nice to fix this, but it sounds from your description that there is no way without loading the tabs again right? I don't think we want to load them again.

chassu:
yeah, I believe those tabs have to be reloaded to persist the favicon and url :/ and it's hard to come up with a fallback option if it DNE

PR #3895, opened 2022-06-22
fix missing url and favicon in archived tabs

chassu:
Use values that have been stored before the change in #3593 

### How was this tested?
- [x] I tested this manually

before:

Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-13 at 15 45 47

after:

Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-22 at 11 03 15

## Issues addressed
#3854


Client/Frontend/Browser/Card/ArchivedTabsListSectionView.swift
```diff
@@ -20,17 +20,18 @@ struct ArchivedTabsRowView: View {
             selectionCompletion()
         } label: {
             HStack {
-                if let url = tab.url {
+                if let url = tab.sessionData?.currentUrl {
```
darinwf, line 23: Should we instead use `sessionData?.currentUrl` as fallback for when `tab.url` is nil?

darinwf: Can't `sessionData` be nil if the tab is not a zombie? I know it will be rare for an archived tab to not be a zombie but it could happen right?

darinwf: Also do we need to do any special handling for internal URLs? Maybe we need to unwrap them to show the intended URL to the user?

chassu: My understanding is that sessionData will be kept forever so we might not need `tab.url` here.

chassu: did some experiments and I wasn't right about this.

chassu: For special handling for internal URLs: say for example of the archived tabs is some invalid site like "foo.bar", we would like to show "foot.bar" instead of the internal URL we use for the error page, right?

darinwf: Right

Issue #3904, opened 2022-06-23
Zombifying tabs from debug setting doesn't initialize sessionData


chassu:
This could cause losing some data (e.x. backForwardList) of tabs and other potential bugs when testing.

PR #3921, opened 2022-06-24
Fix ZombieTab Losing SessionData 

evancrow:
This bug wasn't about the `SessionData` not being restored, it was never set in the first place. To fix that, when tabs were being saved and `SessionData` was created for the `SavedTab`, attach that to the regular `Tab` as well.

### How was this tested?
- [x] I tested this manually

### Manual test cases
- Navigate to site and tap some links to create a history
- Switch to another tab
- Make all tabs zombies from debug settings
- Go back to the original tab and confirm BackForwardList is restored

## Issues addressed
Fixed: #3904


Client/Frontend/Browser/Tabs/SavedTab+ConfigureExtension.swift
```diff
@@ -12,35 +12,32 @@ extension SavedTab {
...
+        let currentItem: WKBackForwardListItem! = tab.webView?.backForwardList.currentItem
 
-                sessionData = SessionData(
-                    currentPage: currentPage, urls: urls,
-                    queries: queries.map { $0?.typed },
-                    suggestedQueries: queries.map { $0?.suggested },
-                    queryLocations: queries.map { $0?.location },
-                    lastUsedTime: tab.lastExecutedTime ?? Date.nowMilliseconds()
-                )
+        // Freshly created web views won't have any history entries at all.
```
yozuysal, line 17: hmm, where in this call do we guarantee that this tab has a freshly created webview?

I guess I am following the part we are now assigning sessionData to tab. sessionData, but not following why we are skipping the sessiondata == nil check

evancrow: Well, I think before we were setting the `SessionData` from somewhere else, which must have gotten removed by mistake. So the check for `SessionData` being `nil` will always return `true`, so it just makes sense to create it once and set it here. Should update the comment.

Client/Frontend/Browser/Tabs/SavedTab+ConfigureExtension.swift
```diff
@@ -12,35 +12,33 @@ extension SavedTab {
...
+        if currentItem != nil {
+            // Here we create the SessionData for the tab and pass that to the SavedTab.
+            let navigationList = tab.webView?.backForwardList.all ?? []
+            let urls = navigationList.compactMap { $0.url }
+            let currentPage = -(tab.webView?.backForwardList.forwardList ?? []).count
+            let queries = navigationList.map {
+                tab.queryForNavigation.findQueryFor(navigation: $0)
             }
+
+            tab.sessionData = SessionData(
```
darinwf, line 28: I'm a little nervous about setting `sessionData` on `Tab` from the `SavedTab` constructor. One concern is that `Tab.sessionData` could become out of sync with `Tab.webView`'s state. It makes sense to me that `Tab.sessionData` should be nil when `Tab.webView` is non-nil. That way there is no confusion about which is the source of truth. It also makes sense to me that when we are about to set `Tab.webView` to nil that we would want to take a snapshot of the back/forward list and generate `Tab.sessionData`.

I think you are essentially achieving the above with this code, but it feels a bit fragile to be doing it as a side-effect of creating a `SavedTab`. I think it would be better to be more explicit in the code. We should have a method to generate a `sessionData` given a `WKWebView` instance, and then explicitly assign that to `Tab` at the right times closer to the point where we are dropping `Tab.webView`.

Makes sense?

evancrow: Yes. I created a dedicated method and was able to clean some of the code. I still believe that `SavedTab.init`, is the most reasonable place to call this.` SavedTab.init` is called for all save paths, and when adding a tab to recently closed, which records the `SessionData` before the `WebView` is removed.

Client/Frontend/Browser/Tabs/SavedTab+ConfigureExtension.swift
```diff
@@ -12,35 +12,13 @@ extension SavedTab {
...
-                sessionData = SessionData(
-                    currentPage: currentPage, urls: urls,
-                    queries: queries.map { $0?.typed },
-                    suggestedQueries: queries.map { $0?.suggested },
-                    queryLocations: queries.map { $0?.location },
-                    lastUsedTime: tab.lastExecutedTime ?? Date.nowMilliseconds()
-                )
-            }
-        }
+        tab.createSessionData()
```
darinwf, line 15: I agree it is a convenient pinch point to make this call, but like I said, it is fragile to do this here disconnected from closing `Tab.webView`. It is also strange that initializing a `SavedTab` would mutate a parameter.

How about just having a method on `Tab` that puts the `Tab` into a zombie state and returns a `SavedTab`?

evancrow: Modified `Tab.close()` (which is our ZombieTab function) to do this

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -697,6 +696,43 @@ class Tab: NSObject, ObservableObject {
...
+                currentPage: currentPage, urls: urls,
+                queries: queries.map { $0?.typed },
+                suggestedQueries: queries.map { $0?.suggested },
+                queryLocations: queries.map { $0?.location },
+                lastUsedTime: lastExecutedTime ?? Date.nowMilliseconds()
+            )
+        }
+    }
+
+    func saveSessionDataAndCreateSavedTab(isSelected: Bool, tabIndex: Int?) -> SavedTab {
```
darinwf, line 724: nit: make these `private` methods

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -697,6 +696,43 @@ class Tab: NSObject, ObservableObject {
         let lastExecutedTime = lastExecutedTime ?? Date.nowMilliseconds()
         return isLastExecutedTimeInTimeFilter(lastExecutedTime, byTime)
     }
+
+    func createSessionData() {
```
darinwf, line 700: nit: call this `saveSessionData` to be consistent with the other method naming... plus it doesn't return a value so "create" is better for those kinds of methods

darinwf (approved):

PR #3924, opened 2022-06-24
Add call back queue param through QueryController

EdwardJTL:
Apollo accepts queue argument on which completion handler is called, which defaults to the main queue.
this change adds the queue param to the `QueryController` class.

Also updating the two controllers used for neeva scope to directly execute on their background queues instead of hopping back onto the shared background queue from the main queue.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
neevascope still works


yozuysal (approved):
Just to double check, these are both only used by Neevascope, right?

EdwardJTL:
> Just to double check, these are both only used by Neevascope, right? 

Yup :)

PR #3927, opened 2022-06-24
Remove lazyHStack in reddit comments

EdwardJTL:
with the lazy stack, the scrollview only expands its height when the first item in the lazy hstack's height increased. When the other items' height changes, the scrollview does not resize itself.

We just have to remove the lazy h stack now.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


yozuysal (commented):
two quick questions:

- How many comments do we generally have in these Views? <10 or more?
- Why cherrypick? Is this addressing a blocker?

EdwardJTL:
> two quick questions:
> 
> * How many comments do we generally have in these Views? <10 or more?
> * Why cherrypick? Is this addressing a blocker?

right now the server can give a LOT of them back, depending on the page. we have a plan to revisit the API in the future.
it is a release blocker since the comments won't show correctly.

yozuysal:
> > two quick questions:
> > 
> > * How many comments do we generally have in these Views? <10 or more?
> > * Why cherrypick? Is this addressing a blocker?
> 
> right now the server can give a LOT of them back, depending on the page. we have a plan to revisit the API in the future. it is a release blocker since the comments won't show correctly.

Can we attach the issue this is blocking for future reference?

If we drop the lazy now, wouldnt it create a performance issue for cases where we have a lot of them?

EdwardJTL:
> > > two quick questions:
> > > 
> > > * How many comments do we generally have in these Views? <10 or more?
> > > * Why cherrypick? Is this addressing a blocker?
> > 
> > 
> > right now the server can give a LOT of them back, depending on the page. we have a plan to revisit the API in the future. it is a release blocker since the comments won't show correctly.
> 
> Can we attach the issue this is blocking for future reference?
> 
> If we drop the lazy now, wouldnt it create a performance issue for cases where we have a lot of them?

as discussed, limit to the first 10 comments. it runs smoothly on simulator.

yozuysal (approved):

PR #3929, opened 2022-06-25
Bring Back the CardStrip!

evancrow:
3rd times the charm :)

Done:
- Removed the bottom option
- Cleaned up the code
- Added drag and drop
- Only tabs from the today section are used
- Tabs in a TabGroup are shown in a compacted UI until one of them is selected

To do:
- [x] TabGroup support
- [x] Handle case with a lot of tabs
- [x] Incognito support
- Tests (follow up when flag is flipped)

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/175751762-b02ce4e2-65d7-4a46-bdd0-5c06f4d266f4.mp4


evancrow:
Tab groups + Lots of tabs:

https://user-images.githubusercontent.com/37356984/175792470-31227c73-1538-43e0-9778-0ec7f2ce2408.mp4

evancrow:
Regular:
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-25 at 17 44 00

Context menu:
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-25 at 17 44 09

Pinned tab:
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-25 at 17 44 13

TabGroup (not selected):
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-25 at 17 54 03

TabGroup (selected w/ pinned tab):
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-25 at 17 44 43

darinwf:
This is a lot of code change! Not everything is isolated behind the flag, and a lot of the code changes are to fiddly / sensitive code. Want to review those changes carefully. I'd ideally like to see this kind of large PR broken down into smaller PRs separating out the refactoring bits from the tab strip implementation.

But I think there's also a large question about how we want to represent tabs on a large screen. The traditional tab browser presumes that tabs have a certain permanence to them. They live at a particular spot, etc.

Our approach has thus far been to treat tabs differently. We probably want to do some design work before committing to a path. I realize this of course is just changing behavior behind a flag, so it isn't necessary to have it all figured out before we land such a PR. Just saying that I have a lot of questions about the design and worry that a traditional approach will box us in.

evancrow:
Synced offline, will work to separate the refactoring

evancrow:
Should be ready for review now!

Client/Frontend/Browser/BrowserView.swift
```diff
@@ -12,6 +12,7 @@ struct BrowserContentView: View {
     let cardGrid: CardGrid
     var topBarHeight: CGFloat
 
+    @EnvironmentObject private var browserModel: BrowserModel
```
darinwf, line 15: Looks like this is not needed and best to avoid extra environment objects that could trigger a `body` call.

Client/Frontend/Browser/CardGrid/Card/CardDetails.swift
```diff
@@ -768,6 +767,11 @@ class TabGroupCardDetails: CardDropDelegate, ObservableObject {
                     manager: manager,
                     isChild: true)
             })
+
+        setIsSelected(selectedTabRootId: tabManager.selectedTab?.rootUUID)
+        selectedTabListener = tabManager.selectedTabPublisher.sink { selectedTab in
```
darinwf, line 772: nit: make this a `[weak self]` to avoid the reference cycle

Client/Frontend/Browser/CardGrid/Card/CardDetails.swift
```diff
@@ -780,6 +784,10 @@ class TabGroupCardDetails: CardDropDelegate, ObservableObject {
         }
     }
 
+    func setIsSelected(selectedTabRootId: String?) {
```
darinwf, line 787: nit: probably better to make this function take a `Tab` and move the `.rootUUID` bit here. That way the callsites get simplified a bit.

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -43,29 +43,17 @@ struct TabGridContainer: View {
...
-            if let row = tabModel.getRows(incognito: false).first { row in
-                row.cells.contains(where: \.isSelected)
-            } {
-                if row.index == 2 {
-                    // scroll to today header
-                    return [TabCardModel.todayRowHeaderID]
-                } else {
-                    return row.id
-                }
+        if let row = tabModel.getRows(incognito: isIncognito).first(where: { row in
```
darinwf, line 46: nice cleanup

Client/Frontend/Browser/CardGrid/Strip/CardStripModel.swift
```diff
@@ -0,0 +1,39 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Combine
+import Shared
+import SwiftUI
+
+class CardStripModel: ObservableObject {
+    @ObservedObject var incognitoModel: IncognitoModel
```
darinwf, line 10: hmm, I thought `@ObservedObject` only made sense on SwiftUI `View` implementations.

Client/Frontend/Browser/CardGrid/Strip/CardStripModel.swift
```diff
@@ -0,0 +1,39 @@
...
+    @ObservedObject var tabCardModel: TabCardModel
+    @ObservedObject var tabChromeModel: TabChromeModel
```
darinwf, line 12: can these just use `let` ?

Client/Frontend/Browser/CardGrid/Strip/CardStripView.swift
```diff
@@ -0,0 +1,80 @@
...
+            return row.id
+        }
+
+        return nil
+    }
+
+    @ViewBuilder
+    var content: some View {
+        HStack(spacing: 0) {
+            ForEach(model.rows) { row in
```
darinwf, line 43: note: I think using `model.rows` works here because `CardStripView` observes `IncognitoModel` and `IncognitoModel`. It is a bit subtle though as just reading the code it would appear at first glance that a dependency on `IncognitoModel` would not be necessary, but it clearly is. Maybe there is a better approach.

evancrow: `IncongitoModel` is used outside of `model.rows`. `selectedRowId` uses `incognitoModel.isIncognito`.

darinwf (approved):

darinwf:
As mentioned before, I'm assuming this is feature flag gated and we'll have future discussions about whether or not to enable this and/or other design changes we might want to consider at that time. Thanks for giving it this nice pass!

PR #3931, opened 2022-06-26
Only Select Tabs in Today on Launch

evancrow:
When the app is launched, only select a tab that if it is in the today section. Before this PR, a tab in an older section or even an archived tab would be selected. Now, if there is no tab last used today, just open to the `CardGrid`.

### How was this tested?
- [x] I tested this manually


chassu (approved):
LGTM

PR #3941, opened 2022-06-27
Don't show cheatsheet promo when neeva button is disabled

EdwardJTL:
We had previously disabled neeva button (and spaces button) on non-webpages and error pages.
But we could still be showing promo on those pages.

Changes:
- Make checks used for these buttons available as class functions
- add these checks in Cheatsheet Promo Model

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Tested manually on a dead page that would've triggered ugc promo


yozuysal (approved):
would these not be useful utilities in general? Should they be inside URLExtensions instead of class functions here? We can name isWebPage isDisplayURLWebPage or something along those lines.

PR #3947, opened 2022-06-28
Add parent tab to `createOrSwitchToTab`

EdwardJTL:
We want to connect back button to cheatsheet, which opens url with `createOrSwitchToTab`.

Changes:
- Add parent tab argument to `createOrSwitchToTab`
- Add parent tab argument to `getTabFor(url:)`
  - if url matches, then verify parent tab if given a parent tab
  - use `parentUUID` for zombie tabs
- Pass on parent tab argument when calling `addTab`
- Use currently selected tab in cheatsheet as parent tab

PS:
1) it seems like tab group comes for free just by setting parent property
2) if tab A opens url B and url C, and then tab with url B also opens url C, there will be two tabs with urls because they will have different parents.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
1) Open url in cheatsheet. Verify back button is enabled. Verify that back button behaves with parent tab
2) Open url in cheatsheet. Go back to the first page, open the same url in cheatsheet. Verify that only one tab is created

| Case 1 | Case 2 |
| ------ | ------ |
| <video src="https://user-images.githubusercontent.com/44557675/176306968-4888ad50-d25c-4120-a041-22593a32072f.MOV"> | <video src="https://user-images.githubusercontent.com/44557675/176306985-de75bb53-1575-4b21-a689-f5e8aa8f39b2.MOV"> |


EdwardJTL:
i also plan on working on how parent tab is stored and restored in a separate PR. currently, when the app exits and comes back, back parent is broken

yozuysal (approved):
I feel like the logic here is very unit testable. Do you mind following up with some TabManagerTests after this?

EdwardJTL:
> I feel like the logic here is very unit testable. Do you mind following up with some TabManagerTests after this?

added in #3952

PR #3949, opened 2022-06-28
Use content visibility to show cheatsheet promos

EdwardJTL:
We are aware that the SwiftUI wrapped anchored popover is buggy. It can try to present a view while swiftUI is animating and the parent view on which the popover is anchored is not visible yet. When that happens, the presented view could be  anchored at the wrong location and automatically dismissed right after being presented.

Previously, we had a hardcoded delay as a workaround.
In this change, we schedule the UI changes we wish to perform, and wait until `ContentVisibilityModel` to signal to us that animation has finished.

other changes:
- Make BVC own `CheatsheetPromoModel` instead of using a shared instance
- refactor and bubble up neeva button dependencies to reduce duplicate code.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
tested onboarding flow and recipe onboarding flow
tested opening a page with UGC content from spaces, outside of the app, and switching to a zombie tab.
tested opening a deadlink that triggers UGC promo (verify that promo does not show)


Client/Frontend/Overlay/Content/Cheatsheet/Promo/CheatsheetPromoModel.swift
```diff
@@ -80,6 +80,18 @@ public class CheatsheetPromoModel: ObservableObject {
         case UGC
     }
 
+    struct UIUpdate {
```
yozuysal, line 83: if all updates are essentially showPromo, showBubble, promoType changing, can we make that a struct as VisibilityState use a publisher for that Visibility state instead of using an open ended completion task like this?

You queue up updates to VisibilityState in a publisher of VisibilityState? and then use essentially the same flow. I feel like it will read somewhat better?

EdwardJTL: that makes sense. will do!

yozuysal (approved):

PR #3953, opened 2022-06-29
TabGroup Drag and Drop

evancrow:
Adds the ability to drag a Tab into a TabGroup, within a TabGroup, and out of a TabGroup. Also includes some code cleanup a fixes a bug where the pinned tab Toast would show when removing a Tab from TabGroup, not when pinning/unpinning a Tab.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/176504877-f0aad5b8-d5e0-441d-8743-0bb0d04eda96.mp4


Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -442,9 +442,20 @@ class TabCardModel: CardModel {
...
+
+        // If a Tab is dragged onto the base grid, reset it's rootUUID to remove it from a TabGroup.
+        draggingDetail.tab.rootUUID = draggingDetail.tab.tabUUID
+        super.dropEntered(info: info)
+    }
+
     // MARK: init
     init(manager: TabManager) {
         self.manager = manager
+        super.init()
```
chassu, line 458: I believe it has to do with `CardDropDelegate`... can you explain why is `super.init()` required only when `CardDropDelegate` is introduced? Trying to think of the reason besides "so that it compiles"

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -457,10 +457,20 @@ class TabManager: NSObject {
...
     func rearrangeTabs(fromIndex: Int, toIndex: Int, notify: Bool) {
+        let toRootUUID = tabs[toIndex].rootUUID
+        if getTabGroup(for: toRootUUID) != nil {
+            // If the Tab is being dropped in a TabGroup, change it's,
+            // rootUUID so it joins the TabGroup.
+            tabs[fromIndex].rootUUID = toRootUUID
+        } else {
+            // Tab was dragged out of a TabGroup, reset it's rootUUID.
+            tabs[fromIndex].rootUUID = tabs[fromIndex].tabUUID
+        }
```
chassu, line 468: I realized this is also executed when performing drag and drop between individual tabs. I know it works, but is it intended?

Client/Frontend/Browser/Card/CardModels.swift
```diff
@@ -442,9 +442,20 @@ class TabCardModel: CardModel {
         return getRows(incognito: false)
     }
 
+    public override func dropEntered(info: DropInfo) {
+        guard let draggingDetail = TabCardDetails.dragState?.draggingDetail else {
+            return
+        }
+
+        // If a Tab is dragged onto the base grid, reset it's rootUUID to remove it from a TabGroup.
+        draggingDetail.tab.rootUUID = draggingDetail.tab.tabUUID
```
yozuysal, line 451: a tab's rootUUID is not ever equal to its tabUUID by design. If we wwant to reset this we should just do UUID().uuidString

Client/Frontend/Browser/Card/CardDropDelegate.swift
```diff
@@ -0,0 +1,37 @@
...
+    /// Called when the user drops the item.
+    public func performDrop(info: DropInfo) -> Bool {
+        Self.dragState = nil
+        return true
+    }
+
+    /// Called right when an item is dragged onto another item.
+    /// Should be implemented by the conforming class.
+    public func dropEntered(info: DropInfo) {
+        guard let tabCardModel = Self.dragState?.tabCardModel else {
```
yozuysal, line 25: I dont understand why we need this (or the need for the self referencing static variable here.

chassu: When I did this, I remember I wanted a way for `TabCardDetail` to reference `tabCardModel` without plumbing the reference down to each card.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
evancrow: Yes, because if a user drags a Tab out of a TabGroup but onto a `Card`, not the `TabScrollController`, it won't be removed from the TabGroup without this.

Client/Frontend/Browser/Card/CardDropDelegate.swift
evancrow: Because there will only ever be one Card dragging at a time, the best way to reference that is from a single instance across all the delegates.

yozuysal: > the best way to reference that is from a single instance across all the delegates.

this is a very general statement which is often times not true. (Statics being the best way to reference a single instance)

My comments are more around having a "state" that essentially keeps a reference to a controller class to use later. This should be a part of the constructor for the class and we should be keeping a reference to the TabManager if needed.

This is a pretty convoluted way to inject a dependency.

Client/Frontend/Browser/Card/CardModels.swift
chassu: I see. I was looking at `TabCardDetails` which implements several protocols so IIUC there's no superclass in that case. But delegate works differently which is why `super.init` is needed after the change.

yozuysal (approved):

Issue #3908, opened 2022-06-23
Sometimes the "show tabs" button does nothing (regression?)


darinwf:
## Expected behavior

Tapping the "show tabs" button should show the tab switcher.

## Actual behavior

Sometimes the button press does nothing. See video.

## Screenshots or recordings

https://user-images.githubusercontent.com/79867437/175359831-7565b3ce-d44e-4853-8159-9fe53453fb6e.MP4

## Hardware and software

- Your device: iPhone 12
- OS version: iOS 15.5
- App version: Neeva Browser 1.46.0 (237)

evancrow:
`CardTransitionState` strikes again

darinwf:
Yeah... must somehow be possible to get into a state where we don't realize what state we are in.

evancrow:
Do you have a repro for this?

EdwardJTL:
i found that this can happen when the date rolls over while you are using the app. or you can reproduce this reliably if you open some tabs, go to settings, set the date to some far time in the future, and come back to the app

evancrow:
Related to #3889 I think

Issue #3889, opened 2022-06-18
Tab Unselected When Moved to Yesterday in Background


evancrow:
Repro:
- Select tab
- Background app (don't close)
- Wait for the selected tab to be moved to yesterday 

The tab is unselected but the done button is still enabled. 

@chassu

PR #3957, opened 2022-06-30
Improve Time Based Tab Logic

evancrow:
- When app returns to the foreground, the `selectedTab.lastExecutedTime` is updated
- When closing a tab that was not used today and was selected, show the `CardGrid` and make sure the `selectedTab` is set to `nil`
- Remove the old testing feature flag, changing the date in settings is just as easy and makes the code cleaner
- Assigns the time section based on the difference from now to the last used time, not the day the last used time fell on (i.e. today is within 24 hours instead of being used on the actual date)

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #3908
Fixed: #3889


Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -873,61 +873,21 @@ class TabWebView: WKWebView, MenuHelperInterface {
...
+    let lastExecutedTime = lastExecutedTime / 1000
+    let timeDifference = currentTimestamp - Double(lastExecutedTime)
+    let secondsInADay: Double = 86400
+
     switch byTime {
     case .today:
-        return FeatureFlag[.shortenTimeThresholdForArchivingTabs]
-            ? lastExecutedTime > Int64(startOfOneDayAgo.timeIntervalSince1970 * 1000)
-            : Date.fromTimestamp(lastExecutedTime).isToday()
+        return timeDifference <= secondsInADay
```
chassu, line 883: I actually started the time based switcher with today being defined by "within last 24 hours". But @darinwf thought whether it would be confusing to users if they wakes up the next day and find out there are tabs in today section, while they haven't used the app at all that day.

chassu:
re 2: how is it possible for a tab to be selected and not used today? Shouldn't select tab reset `lastExecutedTime` so it becomes used today?

Shared/FeatureFlags.swift
```diff
@@ -25,8 +25,6 @@ public enum FeatureFlag: String, CaseIterable, RawRepresentable {
     case qrCodeSignIn = "Sign in with QR Code"
     case recommendedSpaces = "Recommended Spaces"
     case reverseChronologicalOrdering = "Sort tabs based on reverse chronological ordering"
-    case shortenTimeThresholdForArchivingTabs =
```
chassu, line 28: Fine with removing this flag, but how are we going to test archived tabs?

Client/Frontend/Browser/Tabs/Tab.swift
yozuysal: +1 I think the prior implementation was a conscious decision after debate, so let's keep that.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -26,15 +26,26 @@ extension TabManager {
             ? incognitoTabs.firstIndex(where: { $0 == tab })
             : normalTabsToday.firstIndex(where: { $0 == tab })
 
+        // Store an instance of the selectedTab in case the current
```
yozuysal, line 29: can we do this only if the IDs match and not for all closures?

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -26,15 +26,26 @@ extension TabManager {
...
+        // tab is the selecteTab and is removed.
+        let selectedTab = selectedTab
+
         addTabsToRecentlyClosed([tab], showToast: showToast)
         removeTab(tab, flushToDisk: true, notify: true)
 
-        guard let index = index else { return }
+        guard let index = index else {
+            if updateSelectedTab {
+                SceneDelegate.getBVC(with: scene).browserModel.showGridWithNoAnimation()
```
yozuysal, line 38: should this be part of the updateSelectedTabAfterRemovalOf call instead? since it is about that logic.

Shared/FeatureFlags.swift
evancrow: Can change the Date & Time in settings

evancrow:
> re 2: how is it possible for a tab to be selected and not used today? Shouldn't selecting the tab reset `lastExecutedTime` so it becomes used today?

If you background the app, and the date switched over

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
evancrow: Found a way to remove it all together

Shared/FeatureFlags.swift
chassu: With that, I think we can only test from today -> yesterday by setting it to 11:59pm. But it will take a full week for a tab to get archived..

evancrow: Nope! You can change it to be any date as well, so can test week, month, over a month.

Screen Shot 2022-06-30 at 9 46 03 AM

Issue #3866, opened 2022-06-15
Regression: browser becomes super slow after navigating a dozen or so links on a page


darinwf:
## Expected behavior

Browser performance shouldn't get slower due to past browsing.

## Actual behavior

Browser becomes super slow after navigating a dozen or so links on a page

## Steps to reproduce the problem

1. Go to https://news.ycombinator.com/
2. Click on the first link, let the page load, then click back.
3. Click on the second link, let the page load, then click back.
4. Repeat about a dozen times.
5. Notice that scrolling hacker news is now getting sluggish.

## Hardware and software

- Your device: iPhone 12
- OS version: iOS 15.5
- App version: Neeva Browser 1.45.3 (236.1)

darinwf:
When the browser gets slow like this, and I then kill it and restart. Upon restart, I notice that it looks like the history DB may have lost some data about links that I previously clicked on. Links on hacker news that I clicked on are marked unvisited. I wonder if that could be related somehow. Maybe the issue is related to the history system getting bogged down somehow.

chung-lai:
I downgraded to 1.44.6 and felt the sluggish was there.

darinwf:
@chassu said he downgraded to a much earlier build and still saw this problem. So it is not a release blocker.

chung-lai:
But also wonder if the side effect of app being slow has anything to do with your larger history DB. Was it from a new install or it's your personal device?

darinwf:
Personal device where I have yeah, been accumulating a large history DB :)

chassu:
After some investigation, I believe we've always had this issue, and it's caused by memory leak in `subscribe` that the `BrowserViewController` implements:

```
let estimatedProgressPub = webView.publisher(for: \.estimatedProgress, options: .new)
        let isLoadingPub = webView.publisher(for: \.isLoading, options: .new)
        estimatedProgressPub.combineLatest(isLoadingPub)
            .forEach(updateGestureHandler)
            .filter { _ in tab === tabManager.selectedTab }
            .sink { [weak self] (estimatedProgress, isLoading) in
                // When done loading, we want to set progress to 1 so that we allow the progress
                // complete animation to happen. But we want to avoid showing incomplete progress
                // when no longer loading (as may happen when a page load is interrupted).
                if let url = webView.url, !InternalURL.isValid(url: url) {
                    if isLoading {
                        self?.chromeModel.estimatedProgress = estimatedProgress
[...]

Issue #3833, opened 2022-06-12
Scrolling is Sticky


evancrow:
I've noticed on a lot of sites that scrolling is sticky and it often pauses or gets stuck for half a second. Makes scrolling very annoying.

darinwf:
I wonder if this could be related to the ad blocker. What happens if you
disable it?

-Darin

evancrow:
Still happens with ad-blocker and Cookie Cutter disabled. While I can recreate it on most websites, its most prevent on ones where data has to load while scrolling: Instagram, Reddit, etc. Does not occur with Safari though so it is something we're doing.

PR #3968, opened 2022-07-05
Fix Gesture Recognizer Name

evancrow:
The check on if the gesture recognizer was already added would always return `false` because the gesture added was using a different name than what we were looking for.

Confirmed that link/image/highlighting context menus still work. 

### How was this tested?
- [x] I tested this manually

## Issues addressed
(not 100% sure these issues are fixed, want to test out this PR to see)
Issue: #3866
Issue: #3833


darinwf (approved):
woohoo!

evancrow:
@darinwf Been using this PR all day and I'm seeing the scrolling issue come back, not solved by this PR.

evancrow:
I notice that the scrolling issue goes away when using an iPad trackpad, so the issue is limited to touch. Maybe the gesture recognizer is still being added somehow?

Issue #3967, opened 2022-07-03
Back Button Disabled After Backgrounding -> Foregrounding


evancrow:
Repro steps:
- Search on Neeva
- Tap result to open in new page
- Back button enabled (correct)
- Background -> Foreground app
- Back button disabled (should be enabled to tap back to Neeva SRP)

@EdwardJTL

darinwf:
I'm not able to repro. Does this require having the "Open links in new tab" option from the Neeva account settings enabled? Is this dependent on the app being killed in the background?

evancrow:
Yes it does require that setting being enable. 

Killing the app will also reproduce the bug, but just quickly backgrounding it without killing it and opening it causes the bug.

PR #3969, opened 2022-07-05
Don't reset state on backgrounding

evancrow:
`canGoBack` would get reset in the background, pretty sure this logic was from the UIKit code.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/177376386-3bc0a994-b7a2-41c7-8ba3-84819ef279fe.mp4

## Issues addressed
Fixed: #3967


darinwf (approved):

chung-lai:
Looks like this did make it to the release. No cherry picking on this

PR #3970, opened 2022-07-05
Update UGC indicator shadow

EdwardJTL:
The default shadow for popover is too big. There is a lighter shadow, which is made up of 3 shadows, specified on figma. 

I tried a few approaches, but I couldn't figure out how to modify the popover and attach all three shadows. As a compromise, I just picked the most obvious of the three shadows. See screenshots

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

| Before | After |
| Simulator Screen Shot - iPhone 13 - 2022-07-05 at 09 44 58 | Simulator Screen Shot - iPhone 13 - 2022-07-05 at 09 43 33 |


EdwardJTL:
@Ria-yu how does this look?

Client/Frontend/Widgets/Popover.swift
```diff
@@ -25,6 +25,7 @@ extension View {
         isPresented: Binding<Bool>,
         backgroundColor: UIColor? = nil,
         useDimmingBackground: Bool = true,
+        useCustomShadow: Bool = false,
```
darinwf, line 28: nit: `useCustomShadow` makes it seem like the flag should enable you to specify a custom shadow value, but this is just about enabling a different shadow mode. maybe better to use a descriptive name for the special shadow? maybe `useLightShadow` or `useSubtleShadow`?

darinwf (approved):

Ria-yu:
> @Ria-yu how does this look?

I found this [SwiftUI tutorial ](https://designcode.io/swiftui-handbook-shadows-and-color-opacity)that mentions how to add multiple shadows, can you take a look and see if it works?

EdwardJTL:
> > @Ria-yu how does this look?
> 
> I found this [SwiftUI tutorial ](https://designcode.io/swiftui-handbook-shadows-and-color-opacity)that mentions how to add multiple shadows, can you take a look and see if it works?

no this popover is presented through a wrapped UIKit :( I tried playing around with CALayers but they never draw correctly.

Ria-yu:
Got it. Then can you use this new shadow I made? It's a single shadow which I updated on the original mock. 
pasted shadow inspect value here: 
box-shadow: 0px 1px 6px 0px #00000014;

EdwardJTL:
> Got it. Then can you use this new shadow I made? It's a single shadow which I updated on the original mock.
> pasted shadow inspect value here:
> box-shadow: 0px 1px 6px 0px #14;

nice :D
here is what those numbers look like in UIKit
Simulator Screen Shot - iPhone 13 - 2022-07-05 at 16 15 22

PR #3971, opened 2022-07-05
Flip Cheatsheet UGC Indicator Flag

EdwardJTL:
Flip default value to true and add migration

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
- delete app data from simulator
- install an older branch
- install new branch
- verify flag


Client/Application/SceneDelegate.swift
```diff
@@ -604,6 +604,10 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
                 }
             }
 
+            if previousVersion.compare("1.47.5", options: .numeric) == .orderedAscending {
```
darinwf, line 607: What if a user skipped a version and never installed `1.47.5`?

Actually why do we need to worry about migrating any users here?

EdwardJTL: the flag is `off` for all users right now (unless they found internal settings). we want to turn "on" the feature for everyone, so we need to migrate the key to `true`.
i changed this to test if `previousVerions <= 1.48.0`. we are on `1.47.5` on app store and `1.48.0` on testflight. so this check should pass with the next build we ship on either channels.

darinwf (approved):

Client/Application/SceneDelegate.swift
EdwardJTL: we sync'd offline and did an experiment. 
Changing the `default:` in the declaration will get picked up if the default value was never written to.

chung-lai:
Looks like this did make it to the release. No cherry picking on this

PR #3975, opened 2022-07-05
add push token to foreground event (for preview users)

chung-lai:
Add push token as attribute to foreground event to capture token. Will remove this once we have the pipeline to add push token in our backend.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


darinwf:
Why is this only be done for signed out users?

chung-lai:
> Why is this only be done for signed out users?

Yes because we do have the backend pipeline stores the token with the user id for signed in users. But we don't have the pipeline for preview users, want to start logging them so we don't lose them. Once we have the preview user pipeline (DB support), will move them there but it takes a little while before we have the pipeline.

darinwf (approved):

Issue #3965, opened 2022-07-01
Close any Neeva home tabs when user signs out


jedireza:
We should close any tabs on Neeva home (maybe also settings or any non-public route)  when the user signs out. It's strange that I can have a tab with Neeva home open, then sign-out and be left with the existing logged in tab.

darinwf:
An alternative option might be to simply blank out these tabs (i.e., clear their saved screenshot). That way tabs and history within the tabs are not lost.

PR #3976, opened 2022-07-05
Clear Neeva Tabs on Sign Out

evancrow:
Went with @darinwf suggestion on this issue. We should clear the tab and its screenshot when a user logs out.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/177415936-d4a88cf5-b29a-4b75-a10f-672ab6eee3af.mp4

## Issues addressed
Fixed: #3965


jedireza (approved):

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -198,6 +198,25 @@ extension TabManager {
...
+            }
+        }
+    }
+
+    /// Used when the user logs out. Clears any Neeva tabs so they are logged out there too.
+    func clearNeevaTabs() {
+        let neevaTabs = tabs.filter { $0.url?.isNeevaURL() ?? false }
+        makeTabsIntoZombies(tabs: neevaTabs)
+        neevaTabs.forEach {
+            $0.setScreenshot(nil)
```
darinwf, line 216: Reading `setScreenshot`, I think passing `nil` will cause `screenshotUUID` to be updated, which will effectively abandon the old screenshot in `TabManagerStore.shared.imageStore`, but I notice that `setScreenshot` will not cause that old image to be discarded. I think that's okay though. The next time we write out tabs, we'll prune that old screenshot. So I think `setScreenshot(nil)` is a good approach here.

It might be good though to either add a new function named `dropScreenshot` that just is implemented w/ `setScreenshot(nil)` to make the code clearer and/or add a comment to `setScreenshot` explaining what passing `nil` will do.

Client/Frontend/Browser/Tabs/TabManager/TabManager+CloseTabs.swift
```diff
@@ -198,6 +198,25 @@ extension TabManager {
         }
     }
 
+    func makeTabsIntoZombies(tabs: [Tab]) {
+        tabs.forEach { tab in
+            if tab == selectedTab {
+                tab.reload()
```
darinwf, line 204: nit: This is not making the tab into a zombie, so this function is not correctly named. One option would be to just move the body of this function inline into `clearNeevaTabs` since no other code needs to call this function. Or you could filter out the selected tab before calling this function. You already have a filter loop.

darinwf (approved):

PR #3977, opened 2022-07-05
Check All Gestures Before Adding a New One

evancrow:
A better version of #3968. Checks both the `ScrollView` and its `SubViews`.

All context menus still work.

### How was this tested?
- [x] I tested this manually

## Issues addressed
(can confirm the issues are now fixed)

Fixed: #3833 
Fixed: #3866


darinwf (approved):
Your expectation is that `allGestures` will not be a long array right?

darinwf:
Can confirm, this seems to fix the bug!

evancrow:
> Your expectation is that `allGestures` will not be a long array right?

It's not, especially after this PR :)

PR #3980, opened 2022-07-06
Add set behaviour to get logging attributes

EdwardJTL:
Motivation:
- some logs don't really need tab counts, but need session and device info.
- getting tab counts require `main` thread access
- I want to reuse code

Change:
- adding a set that can be used to specify what logging attributes we want from the env helper

PS:
there is also a case for making this an `OptionSet` instead of `Set<enum>`. The signatures and util functions will be different, but it will have some benefits.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
logs still look fine


Client/EnvironmentHelper.swift
```diff
@@ -9,6 +9,27 @@ import Shared
 public class EnvironmentHelper {
     public static let shared = EnvironmentHelper()
 
+    public enum Category: Int, RawRepresentable, Hashable, CaseIterable {
+        case tabs
```
evancrow, line 13: Small nit: sort these alphabetically

evancrow (commented):
This is great! Are you going to do a follow up PR to change all logging everywhere, or is this just for going forward?

Client/EnvironmentHelper.swift
```diff
@@ -53,123 +74,139 @@ public class EnvironmentHelper {
...
+        ]
+    ) -> [ClientLogCounterAttribute] {
+        categories.flatMap { getAttributes(category: $0) } + [getSessionUUID()]
+    }
+
+    public func getFirstRunAttributes() -> [ClientLogCounterAttribute] {
+        getAttributes(for: Category.firstRun)
+    }
+
+    public func getSessionUUID() -> ClientLogCounterAttribute {
```
chung-lai, line 90: @macy-n, do you know if we are still using this 30 minutes session id for anything? Wonder if we can drop it now. Note that this 30 min id is still associated with all the events.

Client/EnvironmentHelper.swift
```diff
@@ -53,123 +74,139 @@ public class EnvironmentHelper {
...
+            Defaults[.sessionUUIDExpirationTime] = Date()
+        }
+
+        // session UUID that will rotate every 30 mins
+        return ClientLogCounterAttribute(
+            key: LogConfig.Attribute.SessionUUID, value: Defaults[.sessionUUID])
+    }
+
+    public func getSessionUUIDv2() -> ClientLogCounterAttribute {
+        // session UUID that will rotate every 30 mins
```
chung-lai, line 103: This comment seems wrong, this v2 id doesn't expire 30 minutes, could you please remove it? I know it was there before lol

chung-lai (commented):
LGTM! I assume this PR is to introduce the categories set but no one is using it yet, right?

EdwardJTL:
> LGTM! I assume this PR is to introduce the categories set but no one is using it yet, right?

yes, the goal of this PR is to introduce the new interface but keep the existing behaviour the same.

Client/EnvironmentHelper.swift
macy-n: we can remove this one, this is replaced with the session uuid v2

chung-lai (approved):

Client/EnvironmentHelper.swift
EdwardJTL: removed

PR #3987, opened 2022-07-06
Fix Old Restored Tab Added to Yesterday When Selected

evancrow:
If you deleted a tab that was in the yesterday section and then restored it, it would be selected but still in the yesterday section. The fix was to move the call that updates the `CardGrid` to after the tab selected, not before.

### How was this tested?
- [x] I tested this manually

Fixed:

https://user-images.githubusercontent.com/37356984/177623393-94ad4281-3353-48d8-9c5e-1b6223c949a9.mp4


Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -281,6 +279,7 @@ extension TabManager {
             recentlyClosedTabs.remove(at: index)
         }
 
+        updateAllTabDataAndSendNotifications(notify: true)
```
darinwf, line 282: I'm concerned about the call to `selectTab` above that passes `notify: true`. That means some code may see a selected tab that does not appear to exist as part of the existing tabs.

darinwf:
Also, please add a test :)

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
evancrow: I can make a change where `updateAllTabDataAndSendNotifications` stays where it originally was, but then we make another call to `updateActiveTabsAndSendNotifications` after selecting the tab. Is that better?

Client/Frontend/Browser/Tabs/TabManager/TabManager+AddTab.swift
```diff
@@ -281,6 +281,7 @@ extension TabManager {
             recentlyClosedTabs.remove(at: index)
         }
 
+        updateActiveTabsAndSendNotifications(notify: true)
```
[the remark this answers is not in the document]
evancrow: Following up on what we discussed in our 1:1, refactored `SelectTab`.

evancrow:
Fixed: #4025

darinwf (approved):

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -272,7 +272,7 @@ class TabManager: NSObject {
         selectedTab.lastExecutedTime = Date.nowMilliseconds()
         selectedTab.applyTheme()
 
-        // Tab data needs to be updated when a tab is brought back from archives.
+        // Tab data needs to be updated after the lastExecutedTime is modified.
```
darinwf, line 275: perfect, thanks!

darinwf (approved):

PR #3990, opened 2022-07-06
UGC log false positive promo

EdwardJTL:
when there is no valid reddit data, query bloom filter and log if returns true

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
manually observed log by forcing filter to return false


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -249,6 +249,26 @@ public class CheatsheetMenuViewModel: ObservableObject {
             let ugcDiscussion = UGCDiscussion(backlinks: cheatsheetInfo.backlinks)
             if !ugcDiscussion.isEmpty {
                 results.append(.discussions(ugcDiscussion))
+            } else {
+                if let inputURL = sourcePage?.url,
+                    let canonicalURL = CanonicalURL(
+                        from: inputURL, stripMobile: true, relaxed: true
+                    )?.asString,
+                    let result = BloomFilterManager.shared.contains(canonicalURL),
```
evancrow, line 257: Simpler if you write this as `!(BloomFilterManager.shared.contains(canonicalURL) ?? false)`

macy-n (approved):

PR #4003, opened 2022-07-07
CardStrip Pre-Refactoring

evancrow:
Moved some code from #3929


evancrow:
This is mostly file organization, with some minor code refactoring.

Client/Frontend/Browser/BrowserView/BrowserView.swift
```diff
@@ -11,7 +12,9 @@ import SwiftUI
 struct BrowserContentView: View {
     let bvc: BrowserViewController
     let cardGrid: CardGrid
+    var topBarHeight: CGFloat
 
+    @EnvironmentObject private var browserModel: BrowserModel
```
darinwf, line 17: This may have non-trivial negative performance impact as it means this view will now be invalidated whenever BrowserModel gets signaled. I don't see why this needs to be here in the environment, so you can just delete it right?

Client/Frontend/Browser/BrowserView/BrowserView.swift
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import Combine
```
darinwf, line 5: nit: seems unnecessary

Client/Frontend/Browser/CardGrid/Card/BorderTreament.swift
```diff
@@ -0,0 +1,25 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
```
darinwf, line 1: I don't understand the `Browser/CardGrid/Card/` directory. We also have a `Browser/Card/` directory. How are these different? Why do we need a `Card/` subdirectory under `CardGrid/`?

Client/Frontend/Browser/BrowserView/BrowserTopBarView.swift
```diff
@@ -65,16 +65,11 @@ struct BrowserTopBarView: View {
     var topBar: some View {
         content
             .transition(.opacity)
-            .frame(height: chromeModel.topBarHeight)
```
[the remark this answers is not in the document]
evancrow: This modifier didn't have an effect, so I just removed it.

Client/Frontend/Browser/BrowserView/BrowserView.swift
```diff
@@ -75,49 +78,48 @@ struct BrowserView: View {
...
+                    ZStack(alignment: .top) {
                         // Tab content or CardGrid
-                        BrowserContentView(bvc: bvc, cardGrid: CardGrid(geom: geom))
-                            .environment(\.shareURL, bvc.shareURL(url:view:))
-                            .padding(
-                                UIConstants.enableBottomURLBar ? .bottom : .top,
-                                topBarHeight
-                            )
-                            .background(Color.background)
+                        BrowserContentView(
```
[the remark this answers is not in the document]
evancrow: That's exactly it. When the CardStrip lands, the TopBar needs to be dynamic in height, and when using SwiftUI in general I've found it best to not set fixed heights. Now we can let SwiftUI handle the layout and respond to changes dynamically without us having to program different heights.

Client/Frontend/Browser/CardGrid/Card/BorderTreament.swift
evancrow: `Browser/Card` was renamed to `Browser/CardGrid`. `Card/` was created to organize all of the files that pertain only to `Card`. Here is the new directory:

/Browser
|-- /ArchivedTabs
|-- /BrowserViewController
|-- /CardGrid
     |-- /Card
          // All the Card specific files
     |-- /Grid
          // All the Grid specific files
|-- /Carousel
|-- /CookieCutter
|-- /Content
|-- /ContentScripts
|-- /Spaces
|-- /Tabs

Client/Frontend/Browser/BrowserView/BrowserView.swift
evancrow: Just removed that, not necessary

Client/Frontend/Browser/CardGrid/Card/BorderTreament.swift
evancrow: @darinwf Don't think there are any objections to the revised directory

yozuysal (commented):
should we consider moving archivedtabs under tabs as well?

evancrow:
> should we consider moving archivedtabs under tabs as well?

Makes sense to me!

yozuysal (approved):

PR #4009, opened 2022-07-08
Remove spinner from sharesheet

buracules:
- Removed spinner from share sheet. (Now it will just update the preview when we have the response)
- Also small fix for metatag parser

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Screenshots and Videos

https://user-images.githubusercontent.com/15333612/178068139-48df34ff-ac11-4a92-9ca5-e006d5e813f4.mp4


Extensions/ShareTo/ShareToViewModel.swift
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import Dispatch
```
evancrow, line 5: Is this library needed? Doesn't look like anything that was added falls under this library, since they were elements pre-PR.

evancrow (approved):
LGTM -  one question

Extensions/ShareTo/ShareToViewModel.swift
buracules: I think xcode automatically added this one probably selected wrong item from auto complete menu

jedireza (approved):
I tested with the Internet disabled and it worked as expected.

PR #4012, opened 2022-07-08
Limit try cheatsheet promo to 5 times

EdwardJTL:
After 5 times we log each time when it gets triggered

Changes:
- add a defaults pref to limit it to 5 times
- if the logic to show popover passes, we check if we reach the limit
- when the promo appears, count down limit
- add logging path

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


Client/Frontend/Overlay/Content/Cheatsheet/Promo/CheatsheetPromoModel.swift
```diff
@@ -217,7 +217,18 @@ public class CheatsheetPromoModel: ObservableObject {
             }
 
             if showIntroPopover {
-                self?.scheduleSetDisplayedStateForTryCheatsheet()
+                if Defaults[.tryCheatsheetPopoverCount] > 0 {
+                    self?.scheduleSetDisplayedStateForTryCheatsheet()
+                } else {
+                    DispatchQueue.main.async {
```
darinwf, line 223: Why the `DispatchQueue.main.async` here?

EdwardJTL: logCounter accesses `SceneDelegate.getBVCOrNil()?.incognitoModel.isIncognito`, which queries `UIApplication.shared.connectedScenes` and has to happen on the main thread

EdwardJTL: i have been thinking about a way to make logs on background thread by keep the incognito check somehow...

darinwf: But isn't this code already running on the main thread? I see a `precondition(Thread.isMainThread)` above

EdwardJTL: oh yes you are right! fixed now :)

macy-n (approved):

macy-n:
LGTM!

darinwf (approved):

Issue #4014, opened 2022-07-09
Regression: Navigating back to FastTap via left swipe is broken


darinwf:
## Expected behavior

Swiping left should take you back to FastTap just as hitting the back button would.

## Actual behavior

Swiping left shows the previous page instead.

## Steps to reproduce the problem

1. Start a new search, e.g. "potato salad"
2. Select one of the FastTap URL results and let the page load
3. Try to swipe left; notice it doesn't work
4. Hit the back button; see the FastTap results again (this part is working as expected)
5. Select a different FastTap URL result and let the page load
6. Try to swipe left; notice it shows the URL from step 2

## Screenshots or recordings

https://user-images.githubusercontent.com/79867437/178110673-27e3bd78-8ffc-4170-a4a9-103c1f6946be.MP4

## Hardware and software

- Your device: iPhone 12
- OS version: iOS 15.5
- App version: Neeva Browser 1.48.2 (243.1)

darinwf:
@EdwardJTL @evancrow FYI ^^^

PR #4017, opened 2022-07-11
Swipe Back to Return to Search Suggestions

evancrow:
@darinwf This wasn't a regression, we never actually supported this feature.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/178177515-a2686e9e-0e81-4922-a03f-5c4d327c8cd2.mp4

https://user-images.githubusercontent.com/37356984/178177528-5cfa30ec-c9fa-462b-af8f-1cfe16229996.mp4

## Issues addressed
Fixed: #4014


Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -481,21 +481,35 @@ class Tab: NSObject, ObservableObject {
         return result.truncateTo(length: 100)
     }
 
+    /// Checks if the current navigation was opened from the Suggest UI.
+    /// - Returns: (Bool, String?): Yes/No, Search Query (nil if no query)
+    func canGoBackToSuggestions() -> (Bool, String?) {
```
EdwardJTL, line 486: could this return just an optional String where a `nil` string will represent `false` in the first value.

EdwardJTL (commented):
clicking the back button on a page opened from fast tap will bring back the suggestion UI. The BVC has fall back logic 
```
if self.simulatedSwipeModel.goBack() {
    return
}

self.tabManager.selectedTab?.goBack()
```
where `Tab.goBack()` will check and bring back the suggestion UI.
could we rearrange the code somehow so that we don't need to implement the "back to suggestion check" logic in two places?

EdwardJTL (approved):

darinwf:
> @darinwf This wasn't a regression, we never actually supported this feature.

Ah, thanks! I hadn't realized. Thanks for hooking it up :-)

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -481,21 +481,38 @@ class Tab: NSObject, ObservableObject {
...
+
+        return query.typed
+    }
+
     func goBack() {
-        // Check if user launched the current URL from the Suggestion UI.
+        if let _ = parent {
+            browserViewController?.tabManager.removeTab(self, showToast: false)
+            return
+        } else if let id = parentSpaceID {
```
darinwf, line 499: nit: no need for "else" here since it follows a "return"

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -481,21 +481,38 @@ class Tab: NSObject, ObservableObject {
...
+        else {
+            return nil
+        }
+
+        return query.typed
+    }
+
     func goBack() {
-        // Check if user launched the current URL from the Suggestion UI.
+        if let _ = parent {
```
darinwf, line 496: what happens if the tab has a parent but the user also clicked on a link in that tab? shouldn't the back button first take users back to the previous URL in the tab before closing the tab and showing the parent?

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -466,21 +466,35 @@ class Tab: NSObject, ObservableObject {
...
-        // Check if user launched the current URL from the Suggestion UI.
-        // If true, show the Suggest UI with that query loaded.
+        // If the user opened this tab from another one, close this one to return to the parent.
+        // Else if the user opened this tab from a space, return to the SpaceDetailView.
+        // Else if the user opened this tab from FastTap, return to FastTap.
         // Else just perform a regular back navigation.
-        if let navigation = webView?.backForwardList.currentItem,
-            let query = queryForNavigation.findQueryFor(navigation: navigation),
-            query.location == .suggestion,
+        if let _ = parent, !canGoBack {
```
darinwf, line 485: It is a little confusing that `Tab.canGoBack` maps to `webView.canGoBack` but `Tab.goBack` includes the `parent` logic.

The lack of symmetry here is potentially going to result in some confusion.  hmm....

evancrow: We can rename `canGoBack` to `webViewCanGoBack`?

darinwf: My question was more about which responsibilities should live in `Tab` versus at a higher level. If the responsibility for handling back navigation from child `Tab` to parent `Tab` is the responsibility of `Tab.goBack`, then we should also make it so that `Tab.canGoBack` corresponds to that.

Of course this is putting a lot of complexity inside of `Tab`. I wonder if there is a design that better separates concerns or maybe it is right for `Tab` to know about all of this. Hmm...

evancrow: Inside of `Tab`, could have a separate `NavigationModel`, which removes the goBack code from both `Tab` and `SimulatedSwipeModel`.

evancrow: Didn't create a separate model, just condensed all the logic I could into `Tab`. The goForward logic I couldn't move out of `SimulatedSwipeModel` since that needs to be at a higher level.

evancrow:
Looks like I messed up a merge commit, fixing that

Client/Frontend/Browser/BrowserModel.swift
```diff
@@ -66,9 +66,7 @@ class BrowserModel: ObservableObject {
     }
 
     func showSpaces(forceUpdate: Bool = true) {
-        cardTransitionModel.update(to: .hidden)
-        contentVisibilityModel.update(showContent: false)
-        showGrid = true
+        showGridWithNoAnimation()
```
darinwf, line 69: Any concern about the addition of the `gridModel.scrollToSelectedTab()` call? Do you see a brief animation?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1196,11 +1196,12 @@ extension BrowserViewController: TabDelegate {
                 // Catch history pushState navigation, but ONLY for same origin navigation,
                 // for reasons above about URL spoofing risk.
                 navigateInTab(tab: tab, webViewStatus: .url)
+
+                updateGestureHandler()
```
darinwf, line 1200: Why does this need to happen again? Didn't it already happen at line 1182?

Client/Frontend/Browser/Carousel/SimulatedSwipeController.swift
```diff
@@ -51,6 +51,8 @@ class SimulatedSwipeController:
             )
             self.progressView.view.backgroundColor = .clear
         }
+
+        self.view.isHidden = true
```
darinwf, line 55: Why set `isHidden` to true here? What bug is this solving?

evancrow: This call was moved from outside the `SimulatedSwipeController` after it was initialized. Made more sense to bring this call into the `init` of the actual class instead of having another file set it.

Client/Frontend/TabChrome/TabChromeModel.swift
```diff
@@ -8,8 +8,14 @@ import Shared
 import SwiftUI
 
 class TabChromeModel: ObservableObject {
-    @Published var canGoBack: Bool
-    @Published var canGoForward: Bool
+    var canGoBack: Bool {
```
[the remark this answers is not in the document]
evancrow: These are `Published` variables from `TabManager` which are just passed along. They will update any views observing them just like any `Published` variables without `TabChromeModel.objectWillChange.send()`. I tested it, and we have test coverage on these back/forward buttons, everything works fine.

Client/Frontend/Browser/BrowserModel.swift
evancrow: Nope! Users also won't see the tabs at all since it will switch directly to Spaces.

Client/Frontend/Browser/BrowserViewController.swift
evancrow: Oh, you're right, not necessary.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1196,11 +1196,12 @@ extension BrowserViewController: TabDelegate {
...
             }
             .store(in: &tab.webViewSubscriptions)
 
         webView.publisher(for: \.title, options: .new)
-            .forEach(updateGestureHandler)
```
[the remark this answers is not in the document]
evancrow: It felt unnessecary but I thought about it and I should probably just add it back. Better safe than sorry.

Client/Frontend/TabChrome/TabChromeModel.swift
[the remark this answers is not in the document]
darinwf: Hmm, if a SwiftUI view is observing `TabManager`, then it would work yeah. But if they switched to observing `TabChromeModel` and not also `TabManager`, then it would not work, right? There could also be flakiness here...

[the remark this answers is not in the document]
darinwf: For example, consider `TabToolbarButtons.BackButton`. That's a SwiftUI view that depends on the value of `TabChromeModel.canGoBack` but does not observe `TabManager`. I'm not sure why this doesn't manifest as a bug in your testing, but it seems like it is a potential bug waiting to happen.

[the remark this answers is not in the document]
evancrow: Nope, works without observing `TabManager`. `OverflowMenuView` for example, only observers `ChromeModel`, and the Forward button relies on `canGoForward` updating properly. Tested that and it updates just fine. The same thing applies to `TabToolbarButtons.BackButton`.

Here's a screen recording of me navigating around:

https://user-images.githubusercontent.com/37356984/179097871-b40fb43d-c761-423b-9651-bba7be96aaba.mp4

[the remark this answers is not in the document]
darinwf: Perhaps some higher level view in the stack is observing `TabManager` and that is causing it to work, but if we ever removed that, then it would break. The way SwiftUI works, you need to make sure each view is watching all of the models it depends on or else you can have subtle bugs like this when you change something in a different part of the code. It seems really fragile to rely on something unrelated to `BackButton` to ensure that its state is kept up-to-date properly.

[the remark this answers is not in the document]
evancrow: Since `canGoBack` is a computed variable based on a `Published` variable, it's also technically a `Published` variable. The `Published` property will call `objectWillChange.send()` on `ChromeModel`. I'm pretty sure this is safe, but do you still have a suggested change here?

I did a bit of poking around, read [this](https://stackoverflow.com/a/62995401) SO thread which gave a good explanation.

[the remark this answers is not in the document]
darinwf: I believe it calls objectWillChange on TabManager and not TabChromeModel.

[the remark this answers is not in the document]
evancrow: `TabManager` also isn't an `ObservableObject`, and there aren't any views that observe it. 

As a test, I took the top-most level view (BrowserView), made `ChromeModel` an `ObservedObject`, and then watched `canGoForward`. Updates properly. 

I did experiment with using a subscription in `ChromeModel` to update `canGoBack/Forward` but it's a lot of extra code and likely not as efficient as the computed variable. Would prefer to not have to do that :)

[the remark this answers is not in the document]
darinwf: I think you need to do more to prove to me that this is how SwiftUI works... doesn't match my understanding of how it works!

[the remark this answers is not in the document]
evancrow: Will do! Could be my understanding is off as well. @EdwardJTL You're knowledgeable on Combine, any thoughts on this?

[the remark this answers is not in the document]
darinwf: Here's a simple SwiftUI program to demonstrate my point:

```
import SwiftUI

class Foo: ObservableObject {
    init() {
        textColor = .black
    }
    @Published var textColor: UIColor
}

class Bar: ObservableObject {
    init(foo: Foo) {
        self.foo = foo
    }
    let foo: Foo
    var textColor: UIColor {
        return foo.textColor
    }
}

struct TestView: View {
    @EnvironmentObject var bar: Bar;

    var body: some View {
        Text("hello world")
            .foregroundColor(Color(uiColor: bar.textColor))
            .padding()
    }
}

struct ContentView: View {
    let bar = Bar(foo: Foo())
    @State var forceUpdate: Int = 0
    var body: some View {
        TestView()
            .environmentObject(bar)
            .id(forceUpdate)
        Button() {
            bar.foo.textColor = .red
//            forceUpdate += 1
        } label: {
            Text("Change Color")
        }
    }
}
```

Just copy / paste that into a stock Xcode project. Try uncommenting that `forceUpdate += 1` line above to see.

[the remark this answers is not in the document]
darinwf: My hypothesis is that there is something akin to `forceUpdate +=1` that is masking the bug of not observing the selected `Tab` (or making the fields on `TabChromeModel` actual properties that we update) in `BackButton` when trying to base UI on `canGoBack`.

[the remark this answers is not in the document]
evancrow: Ah thank you so much for that! Makes sense to me now. I have a solution that I will push!

Client/Frontend/TabChrome/TabChromeModel.swift
```diff
@@ -140,6 +133,18 @@ class TabChromeModel: ObservableObject {
 
             self.isPage = Self.isPage(url: url)
             self.isErrorPage = Self.isErrorPage(url: url)
+
+            self.navigationSubscriptions = []
+            self.canGoBack = tab?.canGoBack ?? false
+            self.canGoForward = tab?.canGoForward ?? false
+
+            tab?.$canGoBack.sink { value in
```
darinwf, line 141: nit: Could use `.assign` here instead...

```
tab?.$canGoBack
    .assign(to: \.canGoBack, on: self)
    .store(in: &self.navigationSubscriptions)
```

darinwf (approved):

PR #4021, opened 2022-07-11
Update appearance of simulated back swipe

EdwardJTL:
Changes:
- Use system background instead of always using white
- Add image view to give preview of the back target
  - Implemented to show parent tab screenshot when appropriate
- Slight refactoring of layout methods and when they are called

Note:
- when removing a tab with parent tab, `updateSelectedTabAfterRemovalOf` does not prioritize selecting the parent tab. as a result, it may jump to a different tab. I am not sure I agree with the decision in `TabManager.selectParentTab(afterRemoving tab)` to not select the parent tab if it's not the most recently used tab.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
- open tab with target blank. swipe back
- open tab from spaces. verify that back swipe picks up on sys background colour

| Preview of parent tab | Sys background colour (opening from spaces) |
| -------- | -------- |
| <video src="https://user-images.githubusercontent.com/44557675/178336704-3851f064-8d2c-47a1-8016-939b7e34cd8c.MOV"> | <video src="https://user-images.githubusercontent.com/44557675/178336735-23d93107-447f-4187-86d7-d9f8df47a932.MOV"> |


darinwf (approved):
LGTM

Issue #4022, opened 2022-07-11
Regression: top and bottom bars no longer hide on scroll


darinwf:
## Expected behavior

When scrolling a page, the top and bottom bars should hide.

## Actual behavior

Top and bottom bars no longer hide.

## Screenshots or recordings

https://user-images.githubusercontent.com/79867437/178342075-432d7cd7-abec-4218-905f-d3b94340ed10.MP4

## Hardware and software

- Your device: iPhone 12 
- OS version: iOS 15.5
- App version: Neeva Browser 1.49.1 (245)

I believe it was fine in 1.48.3, so this is a recent regression on `main`

evancrow:
Going to the CardGrid and back seems to fix this.

darinwf:
> Going to the CardGrid and back seems to fix this.

Interesting! Looks like just creating a new tab through tapping on the URL bar will also resolve the issue.

evancrow:
The issue is caused by the `WebView` being created after the `selectedTabPublisher` is sent, so the `webView.scrollView` is `nil` when `setupDelegates` is set.

PR #4023, opened 2022-07-11
Fix Hide Toolbars on Scroll

evancrow:
Cause by the `WebView` being `nil` when the `selectedTabPublisher` is sent.

### How was this tested?
- [x] I tested this manually

Fixed: #4022


darinwf:
> Cause by the WebView being nil when the selectedTabPublisher is sent.

Hmm, why not fix `selectedTabPublisher` to ensure that the WebView is non-nil then? Instead of working around it. Maybe other code will be broken as well?

Is it hard to make the WebView non-nil when `selectedTabPublisher` runs?

PR #4029, opened 2022-07-12
Revert bundle identifiers for safari extension

darinwf:
Fixes release build


darinwf:
This is restoring to the values from before 7ce3afbae44eef49f3ccd531d5eaeb01f4389050

Client.xcodeproj/project.pbxproj
```diff
@@ -8869,7 +8869,7 @@
 					"-framework",
 					SafariServices,
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = "co.neeva.app.ios.browser-dev.neeva-for-safari.safariExtension";
+				PRODUCT_BUNDLE_IDENTIFIER = co.neeva.app.ios.browser.SafariExtension;
```
evancrow, line 8872: `"` shouldn't be removed

evancrow (approved):
LGTM - with one change

Client.xcodeproj/project.pbxproj
darinwf: interesting! i just did a revert using the diff... i wonder, maybe those quotes don't matter. i'll add them in

PR #4038, opened 2022-07-12
revert space sort feature

EdwardJTL:
As discussed on #4036

the feature was introduced in #3946 and #3884. this revert is not very clean because #3946 has parts that can be kept and #3884 has unrelated changes.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


yozuysal:
this lgtm overall but let's get the tests and build green first. OK that it is not a clean revert, we can revert this PR landing the feature back. @buracules FYI

EdwardJTL:
> this lgtm overall but let's get the tests and build green first. OK that it is not a clean revert, we can revert this PR landing the feature back. @buracules FYI

it's failing on `testPinSpace`. It's probably because #3884 removed the refresh calls that would've otherwise brought the pinned ones to the top. It's the same issue i had with data not updating in #4036. Do you want to revert the entirety of #3884, which will also reopen #3596?

EdwardJTL:
> > > this lgtm overall but let's get the tests and build green first. OK that it is not a clean revert, we can revert this PR landing the feature back. @buracules FYI
> > 
> > 
> > it's failing on `testPinSpace`. It's probably because #3884 removed the refresh calls that would've otherwise brought the pinned ones to the top. It's the same issue i had with data not updating in #4036. Do you want to revert the entirety of #3884, which will also reopen #3596?
> 
> I don't think that is the reason why. I wrote this test in [a separate PR](https://github.com/neevaco/neeva-ios/pull/3902/files). Might just be flaky, I'll take a look.

the test works. I verified manually that on this PR, the pinned space does not rise to the top, and on main before #3884, it moves to the top.

yozuysal (approved):

PR #4043, opened 2022-07-14
Fix Cancel With No Selected Tab ZeroQuery

evancrow:
Fixes a bug where opening ZeroQuery from a shortcut item when no tab is selected, would result in a blank ZeroQuery page.

### How was this tested?
- [x] I tested this manually

Before:

https://user-images.githubusercontent.com/37356984/178866909-67bebeb8-8f49-427c-8a93-6dc27e1a5fa9.mp4

After:

https://user-images.githubusercontent.com/37356984/178866913-e7be86e5-3fac-4ae0-9212-043dde857011.mp4


darinwf (commented):
Is this something you can write a test for?

evancrow:
Don't think so, since this requires going to the app and long pressing on the icon. Unless if XCUITests has the capability for that.

darinwf:
Hmm, I see. Is there another way perhaps to hit this code path so we can exercise via tests? Maybe we'd have to use UI Tests instead?

evancrow:
Yeah, we can use UITests to manually `BVC` code. Will add those.

Client/Application/QuickActions.swift
```diff
@@ -106,17 +105,11 @@ class QuickActions: NSObject {
         }
     }
 
-    fileprivate func handleOpenNewTab(
+    func handleOpenNewTab(
         withBrowserViewController bvc: BrowserViewController, isIncognito: Bool
     ) {
-        // Wait for animations that run when returning to foreground to finish
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
```
darinwf, line 113: Love getting rid of `asyncAfter` calls, but does the problem this was added to address no longer exist?

evancrow: Yeah, I've been testing this a lot while working on the automated tests, and this issue is long gone! It probably was fixed when we did the SwiftUI transition but who knows.

UITests/ShortcutTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        bvc = SceneDelegate.getBVCOrNil()
+        webRoot = SimplePageServer.start()
+    }
+
+    // Runs the code path that a shortcut from the homescreen would call.
+    private func simulateShortcutCase(incognito: Bool, openURL: Bool) {
+        QuickActions.sharedInstance.handleOpenNewTab(
+            withBrowserViewController: bvc, isIncognito: incognito)
+        tester().waitForView(withAccessibilityLabel: "Cancel")
+        sleep(1)
```
darinwf, line 25: Hmm, is this `sleep(1)` needed? Doesn't this code run on the main thread of the app?

evancrow: Removed it and just saw this :), was some debugging code

PR #4052, opened 2022-07-15
Go Back to Parent Tab From ZeroQuery Cancel

evancrow:
From ClickUp. Allows the user to use the back button to navigate to its parent tab when the child was opened from the address bar.

### How was this tested?
- [x] I added automated tests

### Manual test cases
- Run search A. 
- Click on an item N. 
- Run search B from bar. 
- SRP in new tab. 
- Click an item M. 
- Try to use back button to get to N, can't get there. 

https://user-images.githubusercontent.com/37356984/179311597-2da3fece-0295-40d0-8924-03c0e92c9504.mp4


Client/Frontend/Browser/Tabs/TabManager/TabManager+CreateOrSwitchToTab.swift
```diff
@@ -15,18 +15,21 @@ extension TabManager {
         for url: URL,
         query: String? = nil, suggestedQuery: String? = nil,
         visitType: VisitType? = nil,
-        from parentTab: Tab? = nil
+        from parentTab: Tab? = nil,
+        shouldCreateTabGroupWithParent: Bool = true
```
EdwardJTL, line 19: could we name this `keepInParentTabGroup`? since the parent tab may already be in a group

Client/Frontend/Browser/Tabs/TabManager/TabManager+CreateOrSwitchToTab.swift
```diff
@@ -40,6 +43,10 @@ extension TabManager {
                 visitType: visitType
             )
 
+            if let _ = parentTab, !shouldCreateTabGroupWithParent {
```
EdwardJTL, line 46: the `addTab` call tree will end up in `insertTab`, where it examines tabs to insert the new tab and assign its parent relationship. besides, all tab objects are created with a new UUID. To avoid unnecessary assignments, i'd suggest we bubble down the "tab group with parent" property (whatever you end up naming it), so we take care of this association in `insertTab`.

XCUITests/SuggestionBackButtonTests.swift
```diff
@@ -59,4 +59,22 @@ class SuggestionBackButtonTests: BaseTestCase {
         app.buttons["Back"].tap()
         testAddressBarContains(value: "example.com")
     }
+
+    func testReturningToParentTab() {
```
EdwardJTL, line 63: nit: name it `testReturningToParentTabFromURLBar`

EdwardJTL (approved):

PR #4053, opened 2022-07-17
Fix Blurred Reddit Pages

evancrow:
A side effect of #4001. Certain Reddit threads would be blurred out when you tried to view them, with no way to dismiss the blur.

### How was this tested?
- [x] I tested this manually

Before:

After:


jedireza:
This was already in the release branch.

PR #4058, opened 2022-07-18
Cache Cheatsheet Results

EdwardJTL:
To improve hub-n-spoke experience of using cheatsheet

Cheatsheet Changes:
- Refactor page parsing routine
- Skip scheme matching on neeva domain
- Add injection layer for testable data flow
  - ViewModel -> CheatsheetDataService that abstracts network requests and caching

Cache Implementation:
- Wrapper around NSCache
- Key with hashable types
- Value is wrapped with time-to-live expiration date
- Date provider protocol for injecting TTL functionality
- `TimeInterval` backed datetime provider

Parameters:
- Two caches, one for each API call
- 15 items in each cache, TTL is 60 minutes.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
- Test that reader mode still works
- Test that neeva pages are still blocked
- Test that repeated calls and hub-n-spoke flow have expected behaviour
- Test that item count and TTL work (by temporarily setting a low limit)


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetDataStore.swift
```diff
@@ -0,0 +1,412 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+import Shared
+
+private protocol ExpirableDate {
```
yozuysal, line 8: Expirable? since it is not necessarily a date at this point

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetDataStore.swift
```diff
@@ -0,0 +1,412 @@
...
+    static let infoTTL: TimeInterval = .minutes(60)
+    static let searchTTL: TimeInterval = .minutes(60)
+    // count of info results to cache
+    static let infoBudget: Int = 15
+    // count of search results to cache
+    static let searchBudget: Int = 15
+
+    // MARK: - Private Properties
+    private let service: CheatsheetService
+    private let infoCache: TTLCache<InfoParams, InfoResult, TimeIntervalDeadlineProvider>?
```
yozuysal, line 209: do we not always need both calls for both params? Since all parameters above are the same as well, I wonder if we need the two separate caches and keeping two param types. can we have an InfoParams with url, title, query and InfoResult as it is right now and use a single infoCache?

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetDataStore.swift
```diff
@@ -0,0 +1,412 @@
...
+    private let searchCache: TTLCache<SearchParams, SearchResult, TimeIntervalDeadlineProvider>?
+
+    // MARK: - Init
+    init(
+        service: CheatsheetService,
+        infoTTL: TimeInterval? = CheatsheetDataStore.infoTTL,
```
yozuysal, line 215: I think it would be good to avoid the complexity of these being set to nil if we are not going to use that right now. With the caches combined, the below logic would be just a single instantiation without checks.

yozuysal (commented):
a few high level questions about defaults, whether we need two caches to begin with.  I think it would be good to redice complexity here for devs looking at this three months down the line and asking which parts of the flexibility was needed or added just in case.

EdwardJTL:
to answer why we need two caches: the calls are not always connected for a couple of reasons
1) getCheatsheetInfo on two urls return the same memorized query. those would hit the same search result
2) getCheatsheetInfo on a url did not return a query. The fallback methods get a search query for which the result was already cached
3) getCheatsheetInfo has a cached result that did not return a memorized query. however, two calls on two separate tabs have different fall back queries, so they need different search results.
4) getCheatsheetInfo is public but search is not. so if the search call fails, we'd need to populate the search result in the combined cache with blanks. But they can sign in later and should see more results.

I don't really see the benefit of combining the caches. Having separate caches to store results from separate API calls makes sense to me.

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetDataStore.swift
```diff
@@ -0,0 +1,412 @@
...
+        precondition(ttl != nil)
+
+        let deadline = dateProvider.getDeadline(in: ttl!)
+        cache.setObject(
+            WrappedObject(value: value, deadline: deadline),
+            forKey: WrappedKey(key)
+        )
+    }
+
+    func value(for key: Key) -> Value? {
```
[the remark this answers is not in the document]
EdwardJTL: the nil only gets set when it is accessed. Since the value is wrapped in discardable, setting it to nil will give the system permission to clean up that memory even when it's not under pressure.

For all entries in the cache, when the system is under memory pressure, it will start evicting this cache. The TTLCache could have already released the cached entry after a few hours if the system was under memory pressure.

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetDataStore.swift
EdwardJTL: okay i think it makes sense to make the default value non-nil-able.

yozuysal:
> to answer why we need two caches: the calls are not always connected for a couple of reasons
> 
> 1. getCheatsheetInfo on two urls return the same memorized query. those would hit the same search result
> 2. getCheatsheetInfo on a url did not return a query. The fallback methods get a search query for which the result was already cached
> 3. getCheatsheetInfo has a cached result that did not return a memorized query. however, two calls on two separate tabs have different fall back queries, so they need different search results.
> 4. getCheatsheetInfo is public but search is not. so if the search call fails, we'd need to populate the search result in the combined cache with blanks. But they can sign in later and should see more results.
> 
> I don't really see the benefit of combining the caches. Having separate caches to store results from separate API calls makes sense to me.

I am not saying that it will save us memory. In all the above cases you are storing a CheatsheetResultData inside different caches. And I am basically saying, let the keys be different in these cases and cache these separate values inside the same cache. The separate API calls here are not by an optimized design after all. In a perfect world, we would have a single getCheatSheetInfo(CheatSheetInfoInput) that returns a CheatsheetResultData and all the url to query conversion would happen on the backend. I am saying let's not carry this implementation detail split further into the client business logic and merge things as early as we can.

AFAICT the cases you mention above can be covered with nullable's inside the input struct. If there is anything tab specific that is not covered, then that means we need to add more fields (3 I think?) to the input struct rather than relying on these models being owned by separate tabs. (whatever could make them have separate fallback queries should be clear from the input. Something like `fastTapQuery: String`)

The perspective I am coming from is a few months down the line, if a new dev was to refactor something here, it would be pretty hard to tell what is a must have and what can be changed. (Because the current structure suggests that these two pieces are and should stay separate whereas we know that they are not)

(Also I know we would like to get this in on a timeline, so happy to continue this discussion as a followup PR. Not necessarily saying this is a blocking concern, but I think it is better to hide this business split logic as early as we can on the client side.)

EdwardJTL:
> > to answer why we need two caches: the calls are not always connected for a couple of reasons
> > 
> > 1. getCheatsheetInfo on two urls return the same memorized query. those would hit the same search result
> > 2. getCheatsheetInfo on a url did not return a query. The fallback methods get a search query for which the result was already cached
> > 3. getCheatsheetInfo has a cached result that did not return a memorized query. however, two calls on two separate tabs have different fall back queries, so they need different search results.
> > 4. getCheatsheetInfo is public but search is not. so if the search call fails, we'd need to populate the search result in the combined cache with blanks. But they can sign in later and should see more results.
> > 
> > I don't really see the benefit of combining the caches. Having separate caches to store results from separate API calls makes sense to me.
> 
> I am not saying that it will save us memory. In all the above cases you are storing a CheatsheetResultData inside different caches. And I am basically saying, let the keys be different in these cases and cache these separate values inside the same cache. The separate API calls here are not by an optimized design after all. In a perfect world, we would have a single getCheatSheetInfo(CheatSheetInfoInput) that returns a CheatsheetResultData and all the url to query conversion would happen on the backend. I am saying let's not carry this implementation detail split further into the client business logic and merge things as early as we can.
> 
> AFAICT the cases you mention above can be covered with nullable's inside the input struct. If there is anything tab specific that is not covered, then that means we need to add more fields (3 I think?) to the input struct rather than relying on these models being owned by separate tabs. (whatever could make them have separate fallback queries should be clear from the input. Something like `fastTapQuery: String`)
> 
> The perspective I am coming from is a few months down the line, if a new dev was to refactor something here, it would be pretty hard to tell what is a must have and what can be changed. (Because the current structure suggests that these two pieces are and should stay separate whereas we know that they are not)
> 
> (Also I know we would like to get this in on a timeline, so happy to continue this discussion as a followup PR. Not necessarily saying this is a blocking concern, but I think it is better to hide this business split logic as early as we can on the client side.)

i flipped service and store. service now managers both the cache and the network requests. This allows us to separate the discussion of managing the cache and managing the requests.
the two network requests are different; they require different credentials, return different types, and we handle the errors differently. the perfect world where `getCheatsheetInfo` always return a good query is not on our to-dos. So, I don't agree with the statement "we know these pieces are not separate".
we could try merging the cache piece later, but we will need to discuss what the key types and value types are, and how to represent them properly. The unified `CheatsheetResult` type was created just to make my life easier in the Client project. I could have left them in their original `Shared.API` types. But most crucially, only getCheatsheetInfo returns a valid query that can be used for the second search, the data returned by two APIs do not overlap at all, and the four use cases require the value to be separated and not combined. These are additional challenges that need to be addressed if you want to move to a unified cache.

yozuysal (approved):
Thanks for the refactor. Agree that this is cleaner. lgtm so that we can move forward with landing this.

Happy to follow up in zoom about the higher level comments on the design. Again, these are not meant to be for "let's change this now before we move forward":

"the two network requests are different": One of the APIs is called getCheatsheetInfo which is literally "The things we could glean without calling the existing Search API". And the second API is literally "The exact graphQL API that every single SRP uses", so is quite a bit more involved to modify and disentangle. That is why they are separate right now. Not because of an initial design choice we made. Again, all the logic around first getting a query and then using it in Search was a best effort attempt by @macy-n to put together a logical flow that can give us the links and data we need. They indeed give different things, similar to how different entity types in the Search API call give different things. That doesn't point to them belonging to different APIs.

"the perfect world where getCheatsheetInfo always return a good query is not on our to-dos." <- This is exactly a great reason to introduce an abstraction on the client side that can merge two data pieces as early as we can. Ideally all these different conditions you are mentioning are separable through a data structure that makes it clear where things came from. I shouldn't need to look in multiple different places to understand how/why two tabs have different fallback queries. "The perfect world" was getCheatSheetInfo returning a single result struct that contained everything we would need. I am not saying we should prioritize building that. I am just saying at some level on the client, the dev tweaking one of the UI blocks only shouldn't need to care through which API the information came from and how things came to be.

"The unified CheatsheetResult type was created just to make my life easier in the Client project" <- I applaud making your life easier! I am basically saying a similar unification on the input side might make yours and any other devs life easier as well.

jedireza:
FYI: I had to cherry pick this commit into the build. This PR introduces `Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift` and the other PRs https://github.com/neevaco/neeva-ios/pull/4068 and https://github.com/neevaco/neeva-ios/pull/4071 both modified that file, which didn't exist in the build branch.

PR #4060, opened 2022-07-18
Open in App Preferences

evancrow:
Adds the ability to remember a user preference, so the popup only shows once per host. I also set up code to support app names in the popup, but I'm trying to find a good list online of app hosts and corresponding names. If you know of one, lmk :)

### How was this tested?
- [x] I tested this manually

_unfortunately not something I can add tests for (as of now) since it requires having the third-party apps installed_

https://user-images.githubusercontent.com/37356984/179635438-6a761834-66c9-4df9-bce5-ea3b35667b40.MP4

Updated popup:


darinwf:
Would be good to get design review on this. I added it to this week's iOS design open hours. @Ria-yu FYI

evancrow:
After discussion, have decided to not move forward with this. We risk creating a loop where the user can't foreground the Neeva app. No other app has this functionality, so we are fine with the current implementation.

PR #4064, opened 2022-07-19
[i18n] clean up unused strings from Strings class

chung-lai:
* Ran unused.rb script to find all the unused strings in Strings.swift so we don't translate those strings
* Ran abandonedStrings tool to find all the unreferenced string from strings resource file

Most strings are inherited from Firefox repo

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


evancrow (approved):
Looks like there's a lot! Does this mean we're going to have to move all of the new strings around the app into this file?

chung-lai:
> Looks like there's a lot! Does this mean we're going to have to move all of the new strings around the app into this file?

You for sure get this right :)

Planning to have a smaller PR which introduce `NeevaString` which will eventually contain all the strings in the app and they will be defined with `NZLocalizedString` instead of hardcoding with `Text("String") in swiftUI. After that, will start moving all the SwiftUI strings into that file gradually or do it once (there are around ~800 Text() ) in the app. The goal is to have one string file so that we can centralized file to have all the string resource. I just don't feel it's scalable to have hardcoded strings everywhere in different SwiftUI view. Will welcome everyone's feedback on that initial small PR :)

chung-lai:
@evancrow would be great if you can review the `strings` files change too. Turns out we should clean that up as well otherwise the exported localized string file will have them.

evancrow (approved):
Still LGTM! We never really used the string files so deleting them and starting from scratch is fine.

chung-lai:
> Still LGTM! We never really used the string files so deleting them and starting from scratch is fine.

We use it in some places :( i guess there will be follow up work to convert them to use the new strings framework we will be building.

chung-lai:
Will merge after branch cut so we have more baked time.

PR #4066, opened 2022-07-19
Flip Restore App UI Flag

evancrow:
Flips the restore app UI flag. This has been in testing for a bit, and no problems have been reported!

### How was this tested?
- [x] I tested this manually


Client/Application/SceneDelegate.swift
```diff
@@ -60,7 +61,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
         updateCurrentVersion()
 
-        if !urlHandledOnLaunch {
+        if !urlHandledOnLaunch && !testIsRunning {
```
darinwf, line 64: Instead of this var, how about using `AppConstants.IsRunningTest || AppConstants.IsRunningPerfTest` just like `main.swift`?

darinwf (approved):

PR #4067, opened 2022-07-19
Fixes for Xcode beta 3

EdwardJTL:
They fixed the issue with SVG files. We just need to fix the decorated property. But swift-format doesn't like underscores :/ 

it builds and runs


evancrow (approved):

EdwardJTL:
> FYI you can `swift-format` ignore on an as-needed basis: https://github.com/apple/swift-format/blob/main/Documentation/IgnoringSource.md

tries those, they didnt work :/

PR #4080, opened 2022-07-22
[i18n] Enable compiler extract localized string flag and misc fixes

chung-lai:
First pass on app localization:

* Enable "Use compiler to extract swift string":
```
When enabled, the Swift compiler will be used to extract Swift string literal and interpolation LocalizedStringKey and LocalizationKey types during localization export.
```
* Removed and cleaned up a bunch of unused `strings` file
* Added `localizable.strings` for client to add keys for some of the string variables
* Marked all the `WalletCore` strings not to localize them by using `Text(verbatim:)` so it won't appear in exported string file (fyi, @yozuysal)
* Various fixes for some of the frequent used screens (tab switcher, settings, browser, cookie cutter, first run..)

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Client/Frontend/Browser/CookieCutter/CookieCutterOnboardingView.swift
```diff
@@ -32,12 +32,9 @@ struct CookieCutterOnboardingView: View {
...
                 ) {
                     VStack(alignment: .leading, spacing: 8) {
-                        Text("Your Cookie Cutter just declined a ")
-                            + Text("cookie popup ")
-                            .bold()
-                            + Text("and blocked ")
-                            + Text("\(trackingStatsViewModel.numTrackers) trackers.")
-                            .bold()
+                        Text(
+                            "Your Cookie Cutter just declined a **cookie popup** and blocked \(trackingStatsViewModel.numTrackers) **trackers**"
```
darinwf, line 36: This only works in iOS 15 and later right? I guess that is okay. Have you checked out what happens on an iOS 14 device / simulator?

darinwf (approved):
LGTM, just a couple things:

Client/Frontend/Browser/CookieCutter/AdBlockerPromoView.swift
```diff
@@ -26,13 +26,14 @@ public struct AdBlockerPromoView: View {
                 Spacer()
             }
             VStack(alignment: .leading, spacing: 0) {
-                Text("Block Intrusive Ads")
+                Text(verbatim: "Block Intrusive Ads")
```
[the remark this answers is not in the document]
chung-lai: This promo modal only shows for existing users so it shouldn't apply for new users who download the latest app.

Client/Frontend/Browser/CookieCutter/CookieCutterOnboardingView.swift
chung-lai: Interesting, it works though i'll need to file separate issue for cookie cutter not functioning in iOS 14 :P :

Screen Shot 2022-07-25 at 10 57 44 AM

darinwf (approved):

Client/en.lproj/Localizable.strings
```diff
@@ -39,3 +39,18 @@
...
+"Today" = "Today";
+
+/* section header to group tabs created yesterday */
+"Yesterday" = "Yesterday";
+
+/* section header to group tabs created in the last 7 days */
+"Past 7 Days" = "Past 7 Days";
+
+/* section header to group tabs created in the last 30 days */
+"Past 30 Days" = "Yesterday";
```
darinwf, line 53: Actually, this looks like a typo

darinwf (changes requested):

Client/en.lproj/Localizable.strings
chung-lai: Thanks for catching that!

darinwf (approved):
Approving so you aren't blocked.

PR #4087, opened 2022-07-26
Setup Neeva Cookies Opt-in Backend

evancrow:
Added support for Neeva cookies opt-in on the Cookie Cutter side, now we just need to implement the front end.

### How was this tested?
- [x] I tested this manually

### DIFF
--- a/Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
+++ b/Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
@@ -51,39 +51,40 @@ class CookieCutterHelper: TabContentScript {
                     cookieCutterModel.flagSite(domain: domain)
                 }
             case .getPreferences:
-                do {
-                    if let domain = domain,
-                        let escapedEncoded = String(
-                            data: try JSONEncoder().encode([
+                if let domain = domain, let webView = currentWebView {
+                    if currentWebView?.url?.isNeevaURL() ?? false,
+                        cookieCutterModel.optIntoNeevaCookies
+                    {
+                        // Pass false so Cookie Cutter does not block these cookies.
+                        // The user has already consented to these cookies on first run.
+                        sendResponse(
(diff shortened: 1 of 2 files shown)


chung-lai (approved):
Awesome! LGTM. @jlynch630, please also take a look as you are more familiar with the flags :)

chung-lai:
Please hold onto this change for now as we will have a discussion whether we want to do this.

Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
```diff
@@ -51,39 +51,40 @@ class CookieCutterHelper: TabContentScript {
...
+                        cookieCutterModel.optIntoNeevaCookies
+                    {
+                        // Pass false so Cookie Cutter does not block these cookies.
+                        // The user has already consented to these cookies on first run.
+                        sendResponse(
+                            data: [
+                                "cookieCutterEnabled":
+                                    TrackingPreventionConfig.trackersPreventedFor(
+                                        domain, checkCookieCutterState: true),
+                                "marketing": false,
```
jlynch630, line 65: The engine expects true for allowed and false for disallowed/blocked -- do you do some other transformation before these prefs reach the engine?

jlynch630 (commented):
LGTM but should address the true/false debacle first

Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
evancrow: Well shit, I thought we had discussed that it worked similar to our ad blocker where true was passed to block the Cookie, sounds like a major miscommunication on my end. Have a hotfix to upload coming real quick.

chung-lai:
So we decided to revisit this solution at a later time. We can close this for now and reopen later. Thanks for looking into it 🙏

PR #4089, opened 2022-07-26
Cookie Cutter Hotfix

evancrow:
Cookie Cutter flags were being passed incorrectly. 

### How was this tested?
- [x] I tested this manual


Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
```diff
@@ -58,9 +58,9 @@ class CookieCutterHelper: TabContentScript {
                                 "cookieCutterEnabled":
                                     TrackingPreventionConfig.trackersPreventedFor(
                                         domain, checkCookieCutterState: true),
-                                "marketing": !cookieCutterModel.marketingCookiesAllowed,
-                                "analytic": !cookieCutterModel.analyticCookiesAllowed,
-                                "social": !cookieCutterModel.socialCookiesAllowed,
+                                "marketing": cookieCutterModel.marketingCookiesAllowed,
```
darinwf, line 61: nit: We should probably consider updating the fields of the JSON blob to use names like "marketingAllowed" to make the sense of the flag more clear.

darinwf (approved):

Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
darinwf: Probably shouldn't be done as part of this quick fix though.

jlynch630 (approved):

PR #4092, opened 2022-07-26
Update Cookie Cutter JSON Names

evancrow:
Follow up from #4089


Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/CookieCutterHelper.js
```diff
@@ -2,14 +2,14 @@
...
-var cookiePreferences = { marketing: false, analytic: false, social: false }; 
+var cookiePreferences = { marketingAllowed: false, analyticAllowed: false, socialAllowed: false }; 
 var isSiteFlagged = false
 
 function setPreference(preferences) {
     if (preferences["cookieCutterEnabled"]) {
-        cookiePreferences["marketing"] = preferences["marketing"];
-        cookiePreferences["analytic"] = preferences["analytic"];
-        cookiePreferences["social"] = preferences["social"];
+        cookiePreferences.marketingAllowed = preferences["marketingCookiesAllowed"];
```
darinwf, line 10: super minor nit but might be nice to pick one name and use throughout... i'd probably use the shorter form `marketingAllowed` instead of the longer `marketingCookiesAllowed`.

darinwf (approved):
Thanks for following up with this

PR #4094, opened 2022-07-27
Add usage data collection consent in welcome screen

chung-lai:
Exclude the current app server for cookie notice:

Screen Shot 2022-07-28 at 11 37 34 AM

Welcome Screen:

| iPhone 12 | iPhone 8 | iPad |
| --- | --- | --- |
| Screen Shot 2022-07-28 at 10 40 41 AM | Screen Shot 2022-07-28 at 10 43 33 AM | Screen Shot 2022-07-28 at 10 45 19 AM |

Setting:

Screen Shot 2022-07-28 at 10 31 21 AM

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


chung-lai:
@hwiatneeva would love any feedback on the layout. Couple things in particular:

* I tried to do the padding in figma but that push the primary button too high and the subsequent screen's buttons are a bit too high when we try to keep the button at the same position.
* The setting screen copy

Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
```diff
@@ -52,19 +52,24 @@ class CookieCutterHelper: TabContentScript {
                 }
             case .getPreferences:
                 do {
-                    if let domain = domain,
-                        let escapedEncoded = String(
+                    if let domain = domain {
+                        var cookieCutterEnabled = TrackingPreventionConfig.trackersPreventedFor(
+                            domain, checkCookieCutterState: true)
+                        // not to dismiss cookie consent on neeva.com
+                        if let webView = currentWebView, webView.url?.isNeevaURL() ?? false {
```
evancrow, line 59: Shouldn't need to unwrap the WebView. Can just do `currentWebView?.url?.isNeevaURL() ?? false`.

Client/Frontend/Browser/ContentScripts/CookieCutterHelper.swift
```diff
@@ -52,19 +52,24 @@ class CookieCutterHelper: TabContentScript {
                 }
             case .getPreferences:
                 do {
-                    if let domain = domain,
-                        let escapedEncoded = String(
+                    if let domain = domain {
+                        var cookieCutterEnabled = TrackingPreventionConfig.trackersPreventedFor(
+                            domain, checkCookieCutterState: true)
+                        // not to dismiss cookie consent on neeva.com
```
evancrow, line 58: The comment would make a bit more sense written as `Don't dismiss the cookie consent on neeva.com`

Client/Frontend/DefaultBrowserOnboarding/DefaultBrowserInterstitialWelcomeView.swift
```diff
@@ -12,6 +12,16 @@ struct DefaultBrowserInterstitialWelcomeView: View {
 
     @State private var switchToDefaultBrowserScreen = false
```
evancrow, line 14: nit: remove extra line here

Client/Frontend/DefaultBrowserOnboarding/DefaultBrowserInterstitialWelcomeView.swift
```diff
@@ -12,6 +12,16 @@ struct DefaultBrowserInterstitialWelcomeView: View {
...
 
+    @State private var collectUsageStatsCheckbox: Bool = true
```
evancrow, line 15: nit: No need to specify this a `Bool`

Client/Logging/ClientLogger.swift
```diff
@@ -82,34 +82,51 @@ public class ClientLogger {
...
 
+    public func flushLoggingQueue() {
+        for clientLog in loggingQueue {
+            performLogMutation(clientLog)
+        }
+        loggingQueue.removeAll()
+    }
+
+    private func performLogMutation(_ clientLog: ClientLog) {
+        if let shouldCollectUsageStats = Defaults[.shouldCollectUsageStats],
```
evancrow, line 103: Could also write this as ` if Defaults[.shouldCollectUsageStats] ?? false {}`

Client/Logging/ClientLogger.swift
```diff
@@ -82,34 +82,51 @@ public class ClientLogger {
...
+                switch result {
+                case .failure(let error):
+                    print("LogMutation Error: \(error)")
+                case .success:
+                    if let counter = clientLog.counter,
+                        let path = counter?.path
+                    {
+                        log.info("logging sent for counter_path: \(path)")
+                    }
+                    break
```
evancrow, line 127: Don't need the break statement here

Client/Frontend/Settings/Privacy/PrivacySettingsSection.swift
```diff
@@ -11,6 +11,9 @@ struct PrivacySettingsSection: View {
 
     @Default(.closeIncognitoTabs) var closeIncognitoTabs
 
+    @State var usageCollectionCheckbox =
```
evancrow, line 14: Instead of making this a `State` can make it a `Default`: `@Default(.shouldCollectUsageStats) var shouldCollectUsageStats`.


This way, you can remove the `onChange` from the `Toggle` and just use the `Default` value directly.

chung-lai: But `Default(.shouldCollectUsageStats)` is an optional Bool and Toggle doesn't like `Binding<Bool?>`. Any better way to deal with this?

darinwf: Yeah, I'm not sure of a better solution either. Maybe just add a comment about how this can't just use `@Default` because the preference is of type `Bool?`

darinwf (approved):

chung-lai:
@hwiatneeva merging this now but please feel free to leave any feedback and I'll address them accordingly! thanks

hwiatneeva:
Overall LGTM nothing blocking. TY Chung! 

Just casing follow-up:
* In the settings:   Help Improve Neeva ('I' should be caps)

Side note:
* T&C and PP: there's some inconsistency i created among screens. so I'll come back to it for some polishes later. For now, all good as-is.

PR #4095, opened 2022-07-27
Xcode Troubleshooting Guide + Easier To Remove DerivedData

evancrow:
This adds a doc with some common problems people have run into and ways we've found to solve them! Also, when running `bootstrap --force`, added code to delete the DerivedData folder.


darinwf (approved):
Thanks!!

Issue #4084, opened 2022-07-26
Bottom bar pushed up and cannot be restored.


EdwardJTL:
## Expected behavior

Bottom bar stay on the bottom

## Actual behavior

<img width="206" alt="Screen Shot 2022-07-25 at 6 45 50 PM" src="https://user-images.githubusercontent.com/44557675/180904679-a39bf9f3-ad34-4979-a17e-21a3f2c0474f.png">
Screenshot from external folk.

## Steps to reproduce the problem

have not found a reliable way to reproduce it

## Hardware and software
seems to only be a problem when running on iOS 16. I assume the external folk installed the app through app store/testflight instead of building and compiling themselves.

evancrow:
I see the bug here, the BottomBar is appearing while the keyboard is shown when returning to the foreground from the background.

PR #4098, opened 2022-07-27
Fix BottomBar Appearing W/ Keyboard

evancrow:
Caused by incorrect updating of the keyboard height when returning from the background.

### How was this tested?
- [x] I tested this manually

Fixed: #4084


EdwardJTL (approved):

Client/Frontend/Browser/KeyboardListener.swift
```diff
@@ -12,14 +12,33 @@ struct KeyboardListener: ViewModifier {
...
+                    NotificationCenter.default.publisher(
+                        for: UIResponder.keyboardWillShowNotification)
+                ) { _ in
+                    keyboardVisible = true
+                }
+                .onReceive(
+                    NotificationCenter.default.publisher(
+                        for: UIResponder.keyboardDidHideNotification)
+                ) { _ in
+                    keyboardVisible = true
```
darinwf, line 39: shouldn't this set `keyboardVisible` to false?

darinwf (commented):
There's the `KeyboardReadable` helper in `SwiftUIHelpers.swift` that might be handy here.

Client/Frontend/Browser/KeyboardListener.swift
evancrow: Yes, yes it should. Thank you for catching that!

PR #4103, opened 2022-07-28
Improve Cache Control in Apollo GraphQL

EdwardJTL:
Motivation:
- WebKit bug leaks cookies into HTTPCookieStorage.shared
- I want to add cookies to requests
- multithread bug in setting `anonymous` in `GraphQLAPI`
  - network requests can happen concurrently
  - setting the variable is also not thread guarded

Main changes:
- Use interceptors to edit the cookie storage on the underlying `URLSession`
- Add a shared anonymous instance of `GraphQLAPI`
  - this uses an `ephemeral` url session, that does not save/read cookies and cache to file system, which avoids leaks
  - this uses an interceptor that removes login cookies before every request goes out
- Flip code signature for making graphQL calls
  - Before: `Query/Mutation.perform` calls the default instance of `GraphQLAPI` and asks it to perform the query
  - Now: get access to a `GraphQLAPI` instance and use it to make the query. `QueryController` class methods accept a parameter for `GraphQLAPI`
- In spaces, use the anonymous instance on anonymous calls.
- `GraphQLAPI` instances are no longer created lazily
- url session call backs are no longer on main

More details:
- New subclasses added to override behaviours
- New interceptors added to edit cookie storage

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually tested and observed network traffic for:
- get user info
- cheatsheet
- spaces interactions


EdwardJTL:
adding @yusuf, @jonathangarelick to verify spaces behaviour, adding @chung-lai to verify logging and other tokens, adding @jedireza to verify IAP

if you do not have the bandwidth, feel free to slack me a list of things to test

Shared/API/Network.swift
```diff
@@ -77,8 +60,204 @@ public class GraphQLAPI {
...
         }
     }
+
+    // MARK: - Private Methods
+    private init(
+        urlSessionConfig: URLSessionConfiguration,
+        callbackQueue: OperationQueue? = nil,
+        anonymous: Bool = false
+    ) {
+
```
darinwf, line 70: nit: can delete this new line

Shared/API/Network.swift
```diff
@@ -77,8 +60,204 @@ public class GraphQLAPI {
...
+    ) {
+        self.client = client
+        self.shouldInvalidateClientOnDeinit = shouldInvalidateClientOnDeinit
+        self.store = store
+        self.anonymous = anonymous
+    }
+
+    deinit {
+        if self.shouldInvalidateClientOnDeinit {
+            self.client.invalidate()
```
darinwf, line 162: Do we care what thread this runs on?

darinwf (commented):
This makes sense and looks good! I do have some questions about threading though.

darinwf:
@EdwardJTL - Let's find some time to sit down and discuss all of the threading considerations. It is not very clear which code is used on multiple threads, etc.

darinwf:
@EdwardJTL - For starters, can you add a comment above any function that might run on background threads? That would help a lot.

darinwf (approved):
LGTM, thanks for the extra comments and playing it safe :-)

EdwardJTL:
> LGTM, thanks for the extra comments and playing it safe :-)

I am going to hold off merging this. I am tied up with other work, and I want to make sure I have the bandwidth to address any problems if we notice them in testflight.

Issue #4085, opened 2022-07-26
Tab selection animation dropping frames


EdwardJTL:
## Expected behavior

Smooth animation?

## Actual behavior

See recording

## Steps to reproduce the problem

This does not happen with the current app store build (which is a few weeks behind). The screen recording is from building from main on debug.

## Screenshots or recordings

https://user-images.githubusercontent.com/44557675/180906066-3b052351-7d0b-42d0-a1a1-feef947642c5.MP4

evancrow:
Something must have happened to the lazy WebView. Are we sure this is iOS 16 specific?

PR #4104, opened 2022-07-28
Fix Poor CardTransitionModifier Frame Rate

evancrow:
Removing the `transition` modifier seemed to do the trick. Also, added a guard against < 0 width in `updateCardSize` to prevent a negative width SwiftUI runtime error.

### How was this tested?
- [x] I tested this manually

Before:

https://user-images.githubusercontent.com/37356984/181649579-2509d0c6-5d09-4ae6-aca7-9710b22874ca.mp4

After:

https://user-images.githubusercontent.com/37356984/181649643-36553195-cbe9-43f2-b532-fd60ea791de3.mp4

## Issues addressed
Fixed: #4085


darinwf:
Why does this work? Would be nice to see before / after of the same thing, e.g., both landscape or both portrait mode

darinwf:
Is there anything broken by removing that `transition(.identity)`? Can you figure out from blame why it was added in the first place?

evancrow:
Was actually correlation not causation here (so this is not the fix). This issue only occurs when the `Card` is tapped, not the `Done` button. I only used the Done button when `transition` was commented out. That said, I have a better fix which I will push.

darinwf:
I'm seeing the bad animation when tapping a card too. I feel like that must be a regression. It might also be valuable to identify the PR that caused the regression? That might also be good input to figuring out the best fix.

evancrow:
I've narrowed down the issue to be only when details.onSelect is called.

evancrow:
We have two functions that are called when a Card is selected. `Details.onSelect`, and `selectionCompleted`.  `selectionCompleted` works fine by itself for tabs, but SpaceCards need Details.onSelect. Maybe it makes sense to consolidate everything to use `selectionCompleted` and only use `Details.onSelect` as needed at a higher level?

evancrow:
And the culprit is... `updateAllTabDataAndSendNotifications`. Think it makes sense to update this after the animation, just like the `WebView`.

darinwf:
@evancrow - that makes sense!

evancrow:
After:

https://user-images.githubusercontent.com/37356984/181656801-eece0fb5-dee6-461f-a06c-e982acccdbba.mp4

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -272,14 +272,16 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
-        updateAllTabDataAndSendNotifications(notify: notify)
-
         if notify {
             sendSelectTabNotifications(previous: previous)
             selectedTabWebViewPublisher.send(selectedTab.webView)
-        } else if selectedTab.shouldCreateWebViewUponSelect {
+        }
+
+        if selectedTab.shouldPerformHeavyUpdatesUponSelect {
             updateWebViewForSelectedTab(notify: false)
```
darinwf, line 281: Hmm, should you forward the `notify` parameter here?

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -316,6 +318,14 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
         }
     }
 
+    func updateSelectedTabDataPostAnimation() {
+        // Tab data needs to be updated after the lastExecutedTime is modified.
+        updateAllTabDataAndSendNotifications(notify: true)
+        updateWebViewForSelectedTab(notify: true)
+
+        selectedTab?.shouldPerformHeavyUpdatesUponSelect = true
```
darinwf, line 326: nit: recommend setting this flag before calling out to foreign code, which is what the above two functions can do.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -272,14 +272,16 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
         selectedTab.lastExecutedTime = Date.nowMilliseconds()
         selectedTab.applyTheme()
 
-        // Tab data needs to be updated after the lastExecutedTime is modified.
-        updateAllTabDataAndSendNotifications(notify: notify)
-
         if notify {
             sendSelectTabNotifications(previous: previous)
```
darinwf, line 276: I'm anxious about not updating tabs data when generating notifications here. that seems potentially problematic since the notifications may indicate one thing but the tabs data may indicate another. they could be out-of-sync. do you still get good results if you limit honoring the `shouldPerformHeavyUpdatesUponSelect` flag to the case where notifications are not sent?

I'm thinking all of this notification suppression logic is super complicated and fragile!!!

evancrow: I most certainly agree about notifications being fragile! I am not worried for a couple of reasons:

- We have had a lot of regressions where notifications broke things, and a lot of those cases now have plenty of test coverage
- This notification only updates the `CardGrid` logic, so if it's delayed the user won't even know because it will be updated when they try to open the `CardGrid` again.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
evancrow: Moved the code around a bit, and added comments, but basically, we should pass the inverse of `notify`. If notify is true, the `WebView` notifications will be sent from a separate caller in `SelectTab`. If notify in `SelectTab` is false, we should pass `true` to `updateWebViewForSelectedTab` so it sends the notifications.

darinwf (approved):
Thanks for giving this another pass. I agree its good that we have a lot of test coverage in this area, but it still makes me nervous how fragile this system is.

PR #4113, opened 2022-07-29
Test Fix For iPadOS 16 Stage Manager Crash

evancrow:
According to the crash stack, the problem is with calling `getBVC` and `getTabManager`. This should hopefully fix the crash.


Client/Frontend/Browser/Tabs/Tab/Tab.swift
```diff
@@ -513,10 +513,10 @@ class Tab: NSObject, ObservableObject {
     }
 
     func updateCanGoForward() {
-        if let webView = webView, SceneDelegate.getTabManager(for: webView).selectedTab == self {
+        if browserViewController?.tabManager.selectedTab == self {
```
EdwardJTL, line 516: nit: could add `let bvc = browserViewController` so we wont need `?? false` later

EdwardJTL (approved):

PR #4114, opened 2022-07-29
Remove Parent When A Tab's Parent Is Removed

evancrow:
Fixes a bug where after removing a tab`s parent you could still use the back button to close the tab to attempt to return to the parent (even though it no longer existed and the tab was closed for nothing).

### How was this tested?
- [x] I added automated tests


EdwardJTL:
o i was gonna do this as a part of another change 😅

evancrow:
Oh haha! Do you want me to keep this PR or just use yours?

EdwardJTL:
we can go ahead with this. i may revisit the implementation when i get a chance to do my change

darinwf (approved):

Issue #4112, opened 2022-07-29
Regression: Tab group no longer formed when revisiting homepage


darinwf:
1. Start with no tabs
2. Open a site like https://news.ycombinator.com/
3. Click on any link
4. Tab on the URL bar and navigate to https://news.ycombinator.com/ again

Expect to see a tab group formed from the two tabs with the HN homepage in the left hand spot.

Instead two separate tabs are created.

Observed using 1.52.0 (250) and at e49f6ddf3ee1a0066b6df75e17bb6ae52f7d146d.

## Screenshots or recordings

https://user-images.githubusercontent.com/79867437/181839494-f59e2bfd-18ca-4514-867d-48693029d95c.mp4

darinwf:
This is the nytimes case, and `TabGroupTests.swift` should provide coverage. Not sure what is going wrong.

PR #4115, opened 2022-07-29
Fix NYTimes Case When Using Address Bar

evancrow:
Since the change was made to set the parent tab whenever a tab is opened via the address bar, this prevented the code in `insertTab` from creating the tab group for the NYTimes case. I streamlined the logic to make sure almost all of it is run, regardless of if the parent tab is set.

### How was this tested?
- [x] I added automated tests

## Issues addressed
Fixed: #4112


darinwf (approved):
Makes sense, LGTM

Issue #4093, opened 2022-07-27
Back swiping a page shows a black screen instead of the previous tab


jonathangarelick:
TestFlight v1.51.0

Can't see prev page when back swiping, it is all black. The animation also seems slower.

https://user-images.githubusercontent.com/41495718/181153551-76dad6cc-49d9-4474-8d68-0f75c937bea5.MP4

evancrow:
@jonathangarelick Safe to close?

evancrow:
Ah I misread your earlier comment. Thought you meant that the latest version was working fine. This issue has not been fixed then.

PR #4116, opened 2022-07-30
Hide SimulatedBackswipe If WebView Can Go Back

evancrow:
Bug was caused by our back swipe interfering with WebKit's back swipe. Default to WebKit if possible.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/181861949-26b69801-624e-4cac-9a8a-c35013c2c254.mp4

## Issues addressed
Fixed: #4093


darinwf (approved):
nit: Please add a comment explaining this extra check.

evancrow:
Yup! This was caused by #4017. `SimulatedSwipeModel.isHidden` was set on if the `Tab` could always go back, even if the `WebView` had a native swipe back available. When users went to swipe back, they would always see our swipe gesture. Our swipe back can only show a screenshot for the parent tab, which is really only when this feature should be used. Now it still worked correctly because it would call `Tab.goBack` which called `WebView.goBack`.

Anyways, the fix is to hide our swipe when there is a `WebView` swipe available, which is whenever the `WebView.canGoBack` is true.

PR #4124, opened 2022-08-01
Simplify the default browser promo card display logic

chung-lai:
We decided to simplify the promo card display logic and show a persistent default browser promo card for the first three days for the user 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


jedireza (approved):
Code looks good. I didn't test drive.

PR #4138, opened 2022-08-03
Add Tab Count Info to Feedback + Settings

evancrow:
This change should make it easier to debug performance feedback. Includes tab counts in feedback. Note I didn't add a toggle for this, don't know if we need to discuss adding one, but this isn't private information.

### How was this tested?
- [x] I tested this manually

Debug Settings
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-08-02 at 18 53 40

Example Feedback:
https://github.com/neevaco/neeva/issues/72673


Shared/UI/Feedback/SendFeedbackView.swift
```diff
@@ -13,6 +13,8 @@ public struct SendFeedbackView: View {
     let onDismiss: (() -> Void)?
     let screenshot: UIImage?
     let query: String?
+    /// Active Tabs (no zombies), Zombie Tabs, Archived Tabs
+    let tabCounts: (Int, Int, Int)?
```
darinwf, line 17: nit: While it is a bit more typing, I think it would be worth it to create a struct here with named members. That way we minimize risk of the elements getting out of sync.

You might even consider making this a struct defined in `TabManager.swift` with a getter on `TabManager`. Maybe call the struct `TabStats` and then just have `tabManager.stats` return that struct. This way you avoid code duplication between the feedback form and the settings UI, and you get the benefit of named fields.

evancrow: I have to declare `TabStats` in the Shared library so that it can be used by the Feedback UI but other than that, SGTM

darinwf: Oh that seems a bit unfortunate. Should we instead parameterize the feedback code w/ a `String` that is just extra info to include in the feedback report? We could have `Client/` code convert `TabStats` into that `String`. Maybe in the future we will have other data we want to add here, so having a single `String` param would make sense instead of having to teach `Shared/` about the details.

Client.xcodeproj/project.pbxproj
```diff
@@ -5001,6 +5003,7 @@
 				E4E0BB161AFBC9E4008D6260 /* Supporting Files */,
 				C9C18DCE26558FF2004710CD /* SystemColors.swift */,
 				E650758E1E37F7AB006961AC /* SystemUtils.swift */,
+				204EFC2C289AE41300600122 /* TabStats.swift */,
```
darinwf, line 5006: Looks like this can be removed

Shared/TabStats.swift
```diff
@@ -0,0 +1,5 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
```
darinwf, line 5: Can be deleted

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -1358,6 +1358,32 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
         store.clearArchive(for: SceneDelegate.getCurrentScene(for: nil))
     }
 
+    // MARK: - TabStats
+    struct TabStats {
+        /// Number of active tabs (excluding zombies).
+        let numberOfActiveTabs: Int
+        let numberOfZombieTabs: Int
```
darinwf, line 1365: nit: spell out the names instead?

```
numberOfActiveNonZombieTabs
numberOfActiveZombieTabs
```

darinwf (approved):

Client.xcodeproj/project.pbxproj
```diff
@@ -7177,9 +7177,8 @@
 				4A42C4B527E58487003B9826 /* TourPromptViewController.swift in Sources */,
 				4A42C4B627E58487003B9826 /* WebViewContainer.swift in Sources */,
 				4A42C4B727E58487003B9826 /* TrackingMenuView.swift in Sources */,
-				4A42C4B827E58487003B9826 /* TabManager+TabEventHandler.swift in Sources */,
 				205267682877AA0D000550A1 /* (null) in Sources */,
-				205267682877AA0D000550A1 /* CardModels.swift in Sources */,
+				205267682877AA0D000550A1 /* (null) in Sources */,
```
darinwf, line 7181: 👀

evancrow: Ah! Will fix.

evancrow: #4156

PR #4139, opened 2022-08-03
Don't Open CardGrid On Show Tabs Button Longpress

evancrow:
Fixes a bug where long pressing on the Show Tabs button to open the `ContextMenu` would also sometimes open the `CardGrid` which isn't intended. Also includes a tiny bit of refactoring.

### How was this tested?
- [x] I tested this manually


darinwf (approved):
Good catch

Shared/UI/Util/LongPressButton.swift
```diff
@@ -22,20 +22,25 @@ public struct LongPressButton<Label: View>: View {
...
+            LongPressGesture(minimumDuration: 0.3).onEnded { _ in
                 if let longPressAction = longPressAction {
                     longPressAction()
                     Haptics.longPress()
-                    didLongPress = true
+                }
+
+                didLongPress = true
+
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.4) {
```
darinwf, line 42: This deserves a comment. I'm not sure why we need a delay or why it needs to be 0.4. More explanation please :)

darinwf (approved):
LGTM, thanks

PR #4145, opened 2022-08-03
Update LocationModel WebView Whenever SelectedWebView Changes

evancrow:
Calls `updateSecureListener` whenever the selectedWebView is changed, not just whenever navigation occurs in a `Tab`. This caused a small bug where sometimes entering or exiting reader mode would cause the lock icon to show the wrong state. 

### How was this tested?
- [x] I tested this manually


darinwf (approved):

darinwf:
Possible to write a test for this? Maybe using badssl.com?

Client/Frontend/TabChrome/TopBar/LocationView/LocationLabel.swift
```diff
@@ -20,8 +20,13 @@ struct LocationLabel: View {
         )
         .lineLimit(1)
         .frame(height: TabLocationViewUX.height)
-        .allowsHitTesting(false)
+        .if(isSecure != nil) {
+            $0.accessibilityIdentifier(
+                isSecure! ? "locationLabelSiteSecure" : "locationLabelSiteNotSecure"
```
darinwf, line 25: could it have worked to put these identifiers on lines 61 and 63? on the `Symbol` views? it seems like you are having to duplicate the logic here which could be fragile and i also see a lot more checks in `LocationLableAndIcon` that needs to be true for `Symbol` views to added.

evancrow: No because we set `accessibilityHidden` to true (line 28). Unless you think it's fine to remove that modifier.

darinwf: hmm, i'm a bit confused... we set `accessibilityHidden(true)` after `accessibilityIdentifier` here too... why does it matter if it is within `LocationLabelAndIcon` then?

evancrow: It seems to apply to subviews. I tried your original suggestion of applying labels to the Symbols, but `accessibilityHidden` seemed to block that.

darinwf: hmm, interesting

darinwf (approved):
OK

PR #4152, opened 2022-08-04
Pinned Tab Section

evancrow:
Creates a new pinned tab section and sets up for pinned tab improvements. 

(All behind flag)

### How was this tested?
- [x] I tested this manually


Client/Frontend/Browser/CardGrid/Card/CardModels.swift
```diff
@@ -71,19 +73,24 @@ class TabCardModel: CardDropDelegate, CardModel {
...
-            timeBasedNormalRows[.lastMonth] = buildRows(incognito: false, byTime: .lastMonth)
+            timeBasedNormalRows[.lastMonth] = buildRows(incognito: false, for: .lastMonth)
         } else if archivedTabsDuration == .forever {
-            timeBasedNormalRows[.lastMonth] = buildRows(incognito: false, byTime: .lastMonth)
-            timeBasedNormalRows[.overAMonth] = buildRows(incognito: false, byTime: .overAMonth)
+            timeBasedNormalRows[.lastMonth] = buildRows(incognito: false, for: .lastMonth)
+            timeBasedNormalRows[.overAMonth] = buildRows(incognito: false, for: .overAMonth)
         }
 
+        // TODO: in the future, we might apply time-based treatments to incognito mode.
```
darinwf, line 91: Oh, can we have pinned incognito tabs?? Should there be a pinned section for incognito? Hmm... probably some good questions to review w/ @Ria-yu. (I hadn't considered incognito tabs.) Maybe we should not support pinning incognito tabs since they are not really meant to be kept around.

evancrow: We cannot have incognito pinned tabs since the `ContextMenu` with that option is disabled for incognito tabs.

darinwf (approved):

Client/Frontend/Browser/CardGrid/Card/CardModels.swift
Ria-yu: From what I can recall, we made the decision not to support pinned tabs/archived tabs etc. for the reason Darin listed above - incognito tabs are generally not supposed to be kept. While I think there could be interesting use cases for pinned tabs in incognito, I'd not invest in doing it unless we get strong evidence that it'll benefits large number of users.

PR #4154, opened 2022-08-04
Move generation of the tabs state archive to a background thread

darinwf:
We were previously taking care to write to storage on a background
thread, but it turns out the tabs state archive encoding can also
be fairly expensive (10s of milliseconds) with 500+ tabs, and since
this code runs when a tab is selected, it is best to not have it running
on the main thread.


Storage/BackgroundFileWriter.swift
```diff
@@ -25,6 +25,26 @@ private class WriteDataTask: Task {
...
+    private let writer: (Data) -> Void
+
+    init(dataProvider: @escaping () -> Data?, writer: @escaping (Data) -> Void) {
+        self.dataProvider = dataProvider
+        self.writer = writer
+    }
+
+    func isDuplicate(of otherTask: Task) -> Bool {
+        false
+    }
```
EdwardJTL, line 39: should we still try to implement this since `BackgroundTaskProcessor` uses this check. maybe the data provider should provide a `Comparable` data that can be executed synchronously with less cost?

darinwf: It could probably compare `[SavedTab]` array objects, but I don't think this optimization would kick in very much. The pattern I see is that we only reach this code once per selectTab call and then upon navigation completing. There is a lot of time between those events so I don't think this anti-coalescing code will be utilized.

At one point in time, I think we were a bit spammier with the tabs state archiving calls, which is why I had added this system of avoid duplication. I think it is probably an optimization that can just be removed to help simplify the code though.

darinwf: Also, if `isDuplicate` becomes a bit more expensive (such as having to walk over all of the properties of a large `[SavedTab]`), then it starts to defeat the purpose of moving the workload off the main thread.

EdwardJTL (approved):

PR #4144, opened 2022-08-03
Open Neeva.com in New Tab After Sign-in

evancrow:
### Screenshots and screen recordings
https://user-images.githubusercontent.com/37356984/182672989-2580a3cd-0778-4b0f-8c04-01d6320e3575.mp4

## Issues addressed
Fixed: #4143

Issue #3614, opened 2022-05-17
Set login token for okta login flow


chung-lai:
Currently, we rely on navigating to Neeva home to set the proper login token/cookie for Okta login; however, we don't want to navigate to Neeva home if the user is currently browsing. Couple approaches we may take:

* Look into the SafariViewController on how to pass the token around to set the token properly in our web view
* Navigate to neeva home in a phantom tab, extract the cookie/token from the webUI and set the cookie

PR #4155, opened 2022-08-04
New welcome flow

jedireza:
This PR implements the new welcome flow.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Tested in the simulator. I plan on testing on my device also before landing.

So many possibilities to screenshot. Test driving on different devices is recommended.

## Issues addressed
N/A


Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
```
chung-lai, line 1: Love this infra piece. Is the plan to switch `IntroViewModel` to use this AuthStore for authentication given it's mostly the same logic?

jedireza: I was thinking that we could retire the whole `IntroViewModel` when we do a new "Sign Up/In" flow using the new stuff.

Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
...
+import CryptoKit
+import Defaults
+import Shared
+
+private let log = Logger.auth
+
+class AuthStore: NSObject, ObservableObject {
+    public static let shared = AuthStore()
+
+    private var bvc = SceneDelegate.getBVC(for: nil)
```
evancrow, line 15: I don't think we should use `getBVC(for: nil)` here, especially since this view interacts with UI elements. Either should have it passed directly, or find a way to pass in/out only what's needed (my recommendation).

Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
...
+                .registerDeviceTokenWithServer(deviceToken: notificationToken)
+        }
+
+        let httpCookieStore = self.bvc.tabManager.configuration.websiteDataStore.httpCookieStore
+
+        httpCookieStore.setCookie(NeevaConstants.loginCookie(for: token))
+    }
+
+    private func openURLFromAuth(_ url: URL) {
+        if let selectedTab = self.bvc.tabManager.selectedTab,
```
evancrow, line 34: Please update this logic to use the change I made in #4144

Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
...
+        } else {
+            self.bvc.openURLInNewTab(url)
+            DispatchQueue.main.async {
+                self.bvc.hideCardGrid(withAnimation: false)
+            }
+        }
+    }
+
+    // MARK: - Auth Initializers
+
```
evancrow, line 50: nit: remove the extra line here

Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
...
+        else { return }
+
+        self.onError = onError
+        self.onSuccess = onSuccess
+
+        let session = ASWebAuthenticationSession(
+            url: authURL,
+            callbackURLScheme: NeevaConstants.neevaOAuthCallbackScheme()
+        ) { [self] callbackURL, error in
+
```
evancrow, line 90: nit: remove extra line

Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
...
+            guard error == nil, let callbackURL = callbackURL else { return }
+            let queryItems = URLComponents(string: callbackURL.absoluteString)?.queryItems
+            let token = queryItems?.filter({ $0.name == "sessionKey" }).first?.value
+            let serverErrorCode = queryItems?.filter({ $0.name == "retry" }).first?.value
+
+            if let errorCode = serverErrorCode {
+                var errorMessage = "Some unknown error occurred"
+
+                switch errorCode {
+                case "NL003":
```
evancrow, line 105: Should these codes be hard coded in a shared location somewhere? Make them an `Enum` with `String` literals, so it's easier to read/reuse.

Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
...
+                self.handleOauthWithProvider(token: cookie)
+            }
+        }
+
+        session.presentationContextProvider = self
+        session.start()
+    }
+
+    // MARK: - Auth Handlers
+
```
evancrow, line 139: Nit: remove extra line

Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
...
+                    .registerDeviceTokenWithServer(deviceToken: notificationToken)
+            }
+        }
+
+        SpaceStore.shared.refresh(force: true)
+    }
+}
+
+// MARK: - Apple Auth Support
+
```
evancrow, line 187: ditto

Client/Authentication/AuthStore.swift
```diff
@@ -0,0 +1,454 @@
...
+            if Defaults[.signedInOnce] {
+                ClientLogger.shared.logCounter(
+                    .SignInWithAppleSuccess,
+                    attributes: EnvironmentHelper.shared.getFirstRunAttributes()
+                )
+            }
+
+            self.handleSignUpWithApple(
+                identityToken: identityTokenStr, authorizationCode: authorizationCodeStr)
+            break
```
evancrow, line 235: Can remove the `break` here. `break` is only needed if there is no code to execute for a `case`.

Client/Authentication/AuthStore.swift
chung-lai: Ah, makes sense. forgot we are redoing that flow as well. Exciting!

Client/Frontend/Overlay/OverlayManager.swift
```diff
@@ -102,6 +103,22 @@ class OverlayManager: ObservableObject {
         show(overlay: .fullScreenModal(content), animate: animate, completion: completion)
     }
 
+    public func presentFullScreenCover(
```
evancrow, line 106: The `FullScreenModal` is a full-screen cover, so `FullScreenCover` shouldn't be necessary. On iPad, it won't take up the full screen, but this is intentional to follow iPad Design Guidelines.

Client/Frontend/WelcomeFlow/WelcomeFlowPlansView.swift
```diff
@@ -0,0 +1,286 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Shared
+import SwiftUI
+
+struct WelcomeFlowPlansView: View {
+    @ObservedObject var model: WelcomeFlowModel
+
+    var premiumBullets = [
```
chung-lai, line 11: Should convert all these `String` type to `LocalizedStringKey` so they will be localized. Still in the middle of writing the doc about this :)

chung-lai: Ah, saw you are casting them to localized key but you will still need to add them to `Localizable.strings` file

Client/Authentication/AuthStore.swift
jedireza: This function might be going away in favor of letting the consumer choose what to do. In the welcome flow we don't want to open any tabs after authenticating. However, in a future sign up/in flow, we may indeed want to open a tab. Will revisit and ping you.

Client/Authentication/AuthStore.swift
jedireza: I expect this file to be the only location for these values, so I'm ok with them here like this.

Client/Frontend/Overlay/OverlayManager.swift
jedireza: > On iPad, it won't take up the full screen, but this is intentional to follow iPad Design Guidelines.

Yep and that's why I implemented this so we have full control when we need the overlay to be edge to edge.

Client/Authentication/AuthStore.swift
jedireza: 😅 I heard your voice when I put this line of code in. This isn't a view. If needed, we'll figure out a better way to get the bvc in here.

evancrow: Haha, guess my voice wasn't very persuasive :). I think the BVC code is only used here for opening a new tab from sign-in, which shouldn't be a problem. I don't see the user having multiple windows when signing in. Also, hopefully, we can remove the need to open a new tab soon as well.

If we keep this, it should probably be a `let` instead of a `var`.

Client/Authentication/AuthStore.swift
evancrow: I figured we would remove this once we could inject the cookies natively without navigating to Neeva.com. Think the only hold-up was that OKTA was configured to work like this. (am I remembering correctly @chung-lai?)

Client/Frontend/Overlay/OverlayManager.swift
evancrow: Right, I'm just thinking we probably shouldn't use it because it goes against what Apple has designed for iPad. Also, if you want a full-width view, you can configure `OverlayPopoverView` via `OverlayStyle` to fill the view. This just adds another redundancy.

It's kind of confusing, would probably be good to have a quick doc explaining all the options (not asking you to write one, I should probably do that).

Client/Authentication/AuthStore.swift
evancrow: Also would change it to `getBVCOrNil()`, because it could be a bit more accurate in finding the active window if there are multiple.

Client/Authentication/AuthStore.swift
chung-lai: That's right..For OKTA flow, we need to open neeva.com in order to have the right user token set. https://github.com/neevaco/neeva-ios/issues/3614

I think it's because OKTA login was done in the SafariWebViewController and they have their own cookie jar and it doesn't communicate with our web view.

Client/Frontend/WelcomeFlow/WelcomeFlowModel.swift
```diff
@@ -0,0 +1,72 @@
...
+                         LogConfig.PromoCardAttribute.defaultBrowserInterstitialTrigger,
+                     value: trigger.rawValue
+                 )
+             ]
+         )
+         */
+    }
+
+    func complete() {
+        Defaults[.introSeen] = true
```
chung-lai, line 68: Before, `introSeen` was set when user saw the welcome flow. In the new welcome flow, looks like we set this only when user completes the default browser flow or in SceneDelegate's `handleSignInToken` This makes sense to make sure user runs through the entire welcome flow. just curious if that's on purpose?

Client/Frontend/Premium/PremiumStore.swift
```diff
@@ -0,0 +1,201 @@
...
+
+    init() {
+        Task {
+            self.loadingProducts = true
+            do {
+                self.products = try await Product.products(for: [
+                    PremiumPlan.monthly.rawValue, PremiumPlan.annual.rawValue,
+                ])
+            } catch {
+                // TODO: log
```
chung-lai, line 40: Do we plan to add the logging to the same PR? I would suggest we do another PR just for logging. Otherwise, it's a bit hard to see all the logging related changes.

chung-lai (commented):
Left a few comments. Overall, the change looks good and love the refactoring! Would be great if we can break this huge PR into multiple smaller PRs to make it easier to digest but I guess it may be a little hard to pull them apart :)

Client/Frontend/WelcomeFlow/WelcomeFlowModel.swift
jedireza: Yes that was intentional.

Client/Frontend/Premium/PremiumStore.swift
jedireza: We can do that if you prefer it.

Client/Frontend/WelcomeFlow/WelcomeFlowIntroView.swift
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                )
+                .buttonStyle(.neeva(.secondary))
+            }
+
+            Spacer()
+
+            HStack {
+                Button(action: { collectUsageStats.toggle() }) {
+                    collectUsageStats
```
chung-lai, line 92: We should call this when the user completes the flow

```
                Defaults[.shouldCollectUsageStats] = collectUsageStatsCheckbox
                if Defaults[.shouldCollectUsageStats] == true {
                    ClientLogger.shared.flushLoggingQueue()
                }
```

chung-lai: This will determine we should enable logging for the entire app

chung-lai:
Also, we may want to disable the first run test as we are changing the flow now.

Client/Authentication/AuthStore.swift
jedireza: Ok this has been reworked and the bvc is passed in explicitly. Thank you for calling this out.

Client/Frontend/WelcomeFlow/WelcomeFlowPlansView.swift
```diff
@@ -0,0 +1,375 @@
...
+                            .foregroundColor(.brand.white)
+                            .frame(maxWidth: .infinity)
+                    }
+                )
+                .buttonStyle(.neeva(.primary))
+
+                if !NeevaUserInfo.shared.hasLoginCookie() {
+                    Button(
+                        action: {
+                            model.logCounter(.SignInClick)
```
chung-lai, line 242: We should log the screenName as an attribute for `SignInClick` logging so we can tell which screen's sign in button was tapped on.

chung-lai (approved):
One change on the SignInClick to add screen name. Other than that, LGTM with the logging and the usage stats change!

Client/Frontend/WelcomeFlow/WelcomeFlowPlansView.swift
chung-lai: nvm, all the events will get the screenName as well

PR #4157, opened 2022-08-04
fix cookie consent not show up

chung-lai:
Fix one of the tracking protection rules to handle `$ping` type properly

IMG_8443

Will have a follow up to fix any other $ping resource type

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


darinwf (approved):

darinwf:
These json files were generating using some tool, right? Do we need to do something so that if someone runs those tools again, it will generate the changes you have made here?

chung-lai:
> These json files were generating using some tool, right? Do we need to do something so that if someone runs those tools again, it will generate the changes you have made here?

Yes! plan is to fix the tool so it generates those rules properly.

PR #4168, opened 2022-08-05
Fix TabGroups Disappearing

evancrow:
Cause by a logic error in #4152. TabGroups were not being shown if they weren't in the today section because `tab.isIncluded(in: section)` returned false. The fix was to evaluate TabGroups separately with `tabGroup.isIncluded(in: section)`.

### How was this tested?
- [x] I tested this manually
- [x] I added automated tests


Client/Frontend/Browser/CardGrid/Card/CardModels.swift
```diff
@@ -258,7 +258,7 @@ class TabCardModel: CardDropDelegate, CardModel {
             let includedInSection = tab.isIncluded(in: section)
             let includedInSearch = tabIncludedInSearch(tabCard)
 
-            guard incognitoMatches, includedInSection, includedInSearch else {
+            guard incognitoMatches, includedInSearch else {
```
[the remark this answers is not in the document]
evancrow: Yeah, I'll edit it to make more sense

Client/Frontend/Browser/CardGrid/Card/CardModels.swift
```diff
@@ -251,26 +251,29 @@ class TabCardModel: CardDropDelegate, CardModel {
...
+        // Get all the tabs matching the section,
+        // incognito state, and search filter (if active).
         var allDetailsFiltered = allDetails.filter { tabCard in
             let tab = tabCard.tab
             let incognitoMatches = tab.isIncognito == incognito
             let includedInSection = tab.isIncluded(in: section)
             let includedInSearch = tabIncludedInSearch(tabCard)
 
-            guard incognitoMatches, includedInSection, includedInSearch else {
+            guard incognitoMatches, includedInSearch else {
```
chassu, line 262: want to make sure I'm following along - can you explain why removing `includedInSection` fixes the logic error? I'm trying to understand why TabGroups that aren't in the Today Section wasn't showing. I'm thinking whether this check is the actual problem...

chassu: I have a feeling that this is related to the `lastExecutedTime` of a TabGroup being defined as the lastest executed child, and not the last executed time of the representative tab?

chassu: If the `lastExecutedTime` of the last executed child is different from the `lastExecutedTime` of the representative tab, then `includedInSection` and `tabGroup.isIncluded(in: section)` won't be true at the same time.

evancrow: > If the `lastExecutedTime` of the last executed child is different from the `lastExecutedTime` of the representative tab, then `includedInSection` and `tabGroup.isIncluded(in: section)` won't be true at the same time.

That was the bug! The `lastExecutedTime` of the `TabCard` would be different than the actual `TabGroup` so the Cards wouldn't show.

chassu: yup, in #4152, the only check to filter section is `tabGroup!.wasLastExecuted(byTime!)` and that would only place tab group in one section. That extra check `includedInSection` must have caused the regression.

chassu: I wonder if we can put comments about this so we remember what we're doing?

evancrow: I added a bunch, do you think I need more?

chassu: I think existing comments LG, but in addition I think it's better to have a one-liner to remind us the `LastExecutedTime` is defined a bit differently for individual tabs and groups so we can keep in mind for future changes.

chassu:
cc @darinwf for one more pass

evancrow:
@darinwf Marked this for Cherry-Pick since people will likely have hidden tabs without it.

Client/Frontend/Browser/CardGrid/Card/CardModels.swift
```diff
@@ -251,26 +251,32 @@ class TabCardModel: CardDropDelegate, CardModel {
...
             if let tabGroup = manager.getTabGroup(for: tab) {
-                // Make sure the tabCard is the first in the TabGroup.
+                // Make sure tabCard is the first in the TabGroup.
                 let isRepresentativeTab = isRepresentativeTab(tab, in: tabGroup)
+                /* Check if the most recently used tab in TabGroup is
+                 * a part of the section. Or if a tab in the TabGroup is pinned.
+                 * This can differ from tabCard since it uses the most recently used
+                 * tab in the TabGroup. */
                 let tabGroupIncludedInSection = tabGroup.isIncluded(in: section)
                 return isRepresentativeTab && tabGroupIncludedInSection
```
darinwf, line 275: I'm confused by this. What if a representative tab is very old? Won't this cause it to stick around in the section if any tab in its group belongs to the section?

The original code that excludes all tabs not included in the section still seems right to me. Are you sure this part of the change is needed?

darinwf: To be honest, I find the code just really hard to understand, and I'm not sure if the concern I'm raising is valid.

evancrow: I'm also kinda confused. I'm not sure what the point of `isRepresentativeTab` is. I'm seeing what happens if I remove it. Hopefully, it's fine to do away with it.

evancrow: It's actually important, go figure. It prevents the tab group from being added multiple times. Will add a comment.

darinwf: yeah, i know the algorithm that follows wants to process the entire tab group as a chunk, and so it is important that allDetailsFiltered only have one tab from any tab group that needs to be included. my suggestion is to try reverting the changes to how allDetailsFiltered is generated, and see if the bug you were trying to fix is still fixed.

evancrow: Yeah, that was the root cause of the bug. What happened was that the `representativeTab` was older than the most recently used tab in the tab group. So no tab could be added to the filteredList and no group was added to the grid. 

We could leave `includedInSection` in the guard statement (what this PR is removing) and make the `representativeTab` the most recently used. That would require another sort statement though, I don't think that's worth it.

Also, I don't think the old code removed any older tabs from the group, at least that never worked for me. It would show the entire tab group in the section where the most recently used tab was.

darinwf: Yes, reading thru the code, I can see that it is not letting tabs age out of groups as it is supposed to be doing. We can fix that another time.

darinwf (approved):

PR #4185, opened 2022-08-10
Pinned Tab Improvements

evancrow:
(Behind flag)

- Open a new placeholder for the pinned tab when navigation occurs
- Go back to pinned tab (and set WebView) when navigating back
- No forward navs in a pinned tab.

### How was this tested?
- [x] I tested this manually

New placeholder tab opened:

https://user-images.githubusercontent.com/37356984/183962732-0899d096-5d40-4f68-836a-089c9136ac0c.mp4

Returning to parent (zombie):

https://user-images.githubusercontent.com/37356984/183962699-f636c5e7-34fe-4192-af9b-813efd6b8405.mp4

Returning to parent (active WebView):

https://user-images.githubusercontent.com/37356984/183962545-ce5057f2-8e93-42bd-b8ce-566f148a4cc3.mp4

Returning to parent (large nav stack):

https://user-images.githubusercontent.com/37356984/183961831-2a4f7691-102d-4971-9f3c-fe63fc982ced.mp4

Opening multiple children and returning:

https://user-images.githubusercontent.com/37356984/183961806-72541a9d-389f-47d5-831c-0ccd01ab3310.mp4


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -433,7 +433,9 @@ extension WKNavigationAction {
 
 extension BrowserViewController: WKNavigationDelegate {
     func webView(_ webView: WKWebView, didStartProvisionalNavigation navigation: WKNavigation!) {
-        if tabManager.selectedTab?.webView !== webView {
+        guard let selectedTab = tabManager.selectedTab, let url = webView.url,
```
darinwf, line 436: nit: `url` appears to be unused. You could use it down below perhaps? What was the reason for requiring a non-nil `webView.url` here?

Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -1180,6 +1180,20 @@ extension BrowserViewController: TabDelegate {
...
             .filter { tab.url?.origin == $0?.origin }
             .sink { [self] url in
+                // Handle pinned tab navigation.
+                let options: [URL.EqualsOption] = [
+                    .normalizeHost, .ignoreFragment, .ignoreLastSlash, .ignoreScheme,
+                ]
+
+                if let selectedTab = tabManager.selectedTab,
+                    selectedTab.isPinned,
+                    !(selectedTab.url?.equals(url, with: options) ?? false),
```
darinwf, line 1190: nit: Maybe move part of this to a helper function named `treatAsNavigationFromPinnedTab(tab, url)`

Also is it necessary to restrict to tabs that are selected? What about navigations that occur in a background pinned tab? Maybe just use `tab` here instead of looking up `selectedTab`?

Client/Frontend/Browser/SessionData.swift
```diff
@@ -67,9 +69,14 @@ class SessionData: NSObject, NSCoding {
     }
 
     var currentUrl: URL? {
-        // TODO: We should probably unwrap this if it is a session restore internal URL.
         let index = urls.count - 1 + currentPage
-        return 0..<urls.count ~= index ? urls[index] : nil
+        let url = 0..<urls.count ~= index ? urls[index] : nil
+
+        if let nestedUrl = InternalURL.unwrapSessionRestore(url: url) {
```
darinwf, line 75: Glad to see this change, but let's be super careful that we are updating all callsites of `currentUrl` as necessary now. Maybe some are trying to `unwrapSessionRestore` unnecessarily now? Please audit all callers.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
```diff
@@ -1367,6 +1358,51 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
+        // Create a new placeholder tab to represent the pinned tab.
+        // Should be a zombie with the same session data as
+        // the original pinned tab before it navigated.
+        let newPinnedTab = addTab(atIndex: tabIndex, flushToDisk: true, zombie: true)
+        let savedTab = tab.saveSessionDataAndCreateSavedTab(isSelected: false, tabIndex: tabIndex)
+        savedTab.configureTab(newPinnedTab, imageStore: store.imageStore)
+        newPinnedTab.updateCanGoBackForward()
+
+        // Reset the navigated tab (the original pinned tabs) data.
+        tab.tabUUID = UUID().uuidString
```
darinwf, line 1387: The child tab should still be in the same tab group as the pinned tab.

darinwf (commented):
Please add some tests :-D

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
evancrow: Ah, this is where I originally had the placeholder tab code. I can remove this change.

Client/Frontend/Browser/Tabs/TabManager/TabManager.swift
evancrow: Yeah, it is. The `rootUUID` is kept the same. The setter: `tab.rootUUID = newPinnedTab.rootUUID`, is actually redundant because `newPinnedTab.rootUUID` is the same as `tab` since they were created from the same data.

darinwf: Whoops, I just misread this line of code :-P

evancrow:
> Please add some tests :-D

Happy to add tests now, but is it fine if we wait till the flag is enabled? Either way is good with me.

darinwf:
> > Please add some tests :-D
> 
> Happy to add tests now, but is it fine if we wait till the flag is enabled? Either way is good with me.

You are saying that because it is hard to enable feature flags in tests? Maybe we should make that easy to do. I bet we could do it via LaunchArguments.

evancrow:
We do already have a LaunchArgument for enabling the flag. I'll just use that.

darinwf:
> We do already have a LaunchArgument for enabling the flag. I'll just use that.

Might need to plumb it thru... seems straightforward to do.

darinwf (approved):

PR #4186, opened 2022-08-10
iPad Improvements

evancrow:
- Adds pointer effects to toolbar buttons
- Adds window titles

### How was this tested?
- [x] I tested this manually

Window Titles:
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-08-10 at 10 47 48

Pointer Effects:

https://user-images.githubusercontent.com/37356984/183981459-75de3456-6fa4-4b87-8735-b2d6b65a861b.mp4

A small bug I noticed is that when tapping a button with a `hoverEffect`, it lags a bit when disappearing, which is odd. Will take a look.


EdwardJTL (approved):

evancrow:
The weird button sticking issue isn't an issue on an actual device, just simulator lag.

PR #4188, opened 2022-08-10
Cheatsheet log end of journey

EdwardJTL:
Added a new log and observed logs

Note:
- because of our overlay lifecycle, if cheatsheet is dismissed by opening a url, `onDisappear` log will be called on the newly created tab's cheatsheet model, which will then log with a nil journey ID. I think this can be remediates in SQL reasonably easily


macy-n (approved):

darinwf:
Marking `cherry-pick-done` since I see this on `Build-255-release-branch`

PR #4199, opened 2022-08-11
TabGroupPanel Overhaul

evancrow:
- Lots of refactoring
- Design modification (to review with @Ria-yu)
    -  Brings it in line with History. Happy to revert this part of the PR if needed.

### How was this tested?
- [x] I tested this manually

Different variations of the panel:
 

## Issues addressed


evancrow:
Date headers (went into the future for this one 😎) :

Client/Frontend/Browser/Tabs/Tab/Tab.swift
```diff
@@ -89,7 +89,7 @@ class Tab: NSObject, ObservableObject {
 
     @Published var favicon: Favicon?
 
-    var lastExecutedTime: Timestamp?
+    var lastExecutedTime: Timestamp = Date.nowMilliseconds()
```
evancrow, line 92: @chassu @darinwf Made `lastExecutedTime` a non-optional variable now, with the current time as default.

evancrow:
Duration label added:

Client/Frontend/Browser/Tabs/ArchivedTabs/ArchivedTabsPanelModel.swift
```diff
@@ -6,59 +6,98 @@ import Combine
...
     func loadData() {
-        archivedTabs = tabManager.archivedTabs
-        archivedTabGroups = tabManager.archivedTabGroups
+        groupedRows = DateGroupedTableData<ArchivedTabRow>()
+
+        // Once all the tab groups are added, no need to also add duplicate tabs.
+        // This prevents any tab with a RootID already added from being inserted into the rows again.
+        var handledRootIDs: [String] = []
+
+        // Add all the TabGroups in `section` to `groupedRows`.
```
darinwf, line 59: Hmm, it looks like this may be re-computing what `TabManager.archivedTabGroups` already provides.

Client/Frontend/Browser/Tabs/ArchivedTabs/ArchivedTabsPanelRowView.swift
```diff
@@ -0,0 +1,57 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Shared
+import SwiftUI
+
+struct ArchivedTabsRowView: View {
+    @Environment(\.selectionCompletion) private var selectionCompletion: () -> Void
+
+    let tab: Tab
```
darinwf, line 11: Should this be observing `tab`? The view appears to depend on `Tab.url`, which is a published property. I worry SwiftUI might continue to use a stale instance of `ArchivedTabsRowView` that might show a favicon for a different site, for an old value of `tab.url`.

Client/Frontend/Browser/Tabs/ArchivedTabs/ArchivedTabsPanelRowView.swift
```diff
@@ -0,0 +1,57 @@
...
+    let tabManager: TabManager
+    private let padding: CGFloat = 4
+
+    var tabURL: URL? {
+        if let url = tab.url ?? tab.sessionData?.currentUrl {
```
darinwf, line 16: I feel like this should really be a helpful function on `Tab` to get the url, or we should take it as a follow-up PR to figure out how to make sure `tab.url` always returns the right thing, hiding details about zombie tabs and tabs with error pages, etc.

Client/Frontend/Browser/Tabs/ArchivedTabs/ArchivedTabsPanelRowView.swift
darinwf: Or maybe it is fine since this UI should only be a view on a snapshot of the archived tabs data anyways, and whenever that data changes the whole view would get re-created or not be visible at that time.

Client/Frontend/Browser/Tabs/Tab/Tab.swift
darinwf: I think this makes sense. This helps cleanup consumer code which was anyways assuming this as the default.

darinwf (approved):

Client/Frontend/Browser/Tabs/ArchivedTabs/ArchivedTabsPanelRowView.swift
evancrow: This was fixed in another PR, just need to merge those changes..

Client/Frontend/Browser/Tabs/ArchivedTabs/ArchivedTabsPanelRowView.swift
evancrow: That's exactly it, it will all get reloaded when the view becomes visible again. The user can't take any tab-changing action without leaving the view.

Issue #3632, opened 2022-05-19
Neeva icons for a tab flicker in card grid as card is animated


darinwf:
## Expected behavior

Icons should animate smoothly with the card as it goes through its animation transition

## Actual behavior

The size of the neeva icon seems to jump / flicker as the tab transition animation runs

## Screenshots or recordings

https://user-images.githubusercontent.com/79867437/169217633-55e07cce-d0ed-43b4-b1ae-38cb051dc8d6.MP4

## Hardware and software

- Your device: iPhone 12
- OS version: 15.4.1
- App version: Neeva Browser 1.41.3

Issue #3551, opened 2022-05-10
History favicon jitters while typing into the search box


darinwf:
https://user-images.githubusercontent.com/79867437/167669132-edc2a0e4-4b7b-4ca3-841c-92a8ad5ee0f7.MP4

@evancrow FYI

Issue #4159, opened 2022-08-04
Tab Grid icon animated separately during transition


EdwardJTL:
If you look at it frame by frame, the text and the screenshot are animated at the same time, but the icon is not. Towards the end of the animation, there is a noticeable bounce animation on the icon where the bottom part of it gets occluded.

## Screenshots or recordings

https://user-images.githubusercontent.com/44557675/182942343-40e79192-4a4a-4a78-b958-769c75db7509.MP4

## Hardware and software

- Your device: iPhone 13
- OS version: 16 beta 3
- App version: Neeva Browser 1.51.1 (249.1)

darinwf:
On iOS 15, I also observe the favicon appearing to not be animating in perfectly alongside the text label. Perhaps that is a clue as to how to resolve this.

evancrow:
Resolved by #4250

PR #4202, opened 2022-08-12
Add caching to FaviconResolver to avoid flickering of favicons

darinwf:
The system frequently creates and recreates FaviconView objects,
and the FaviconResolver has an async DB lookup to find the URL.
As a result, each time a FaviconView gets created it would show
the fallback content momentarily before showing the right image.
This can result in a fair bit of flicker.

The simple solution here is to maintain an in-memory cache of
site URLs to favicon URLs. This way we can synchronously know
the favicon URL to use and avoid showing fallback content in
most cases.

It's necessary to update this cache whenever FaviconHandler
learns of a new mapping.

This also seems to fix the issue with the favicon of a card
animating out of sync with the rest of the card during the
card transition animation. This is probably some side-effect
of animating the view while it is transitioning from one view
to another. Maybe tickling a bug in SwiftUI or something, but
the cache helps avoid this issue too.

Fixes https://github.com/neevaco/neeva-ios/issues/3632, https://github.com/neevaco/neeva-ios/issues/3551


evancrow:
Does this also resolve #3551?

EdwardJTL:
The clipping animation issue on iOS 16 is still there
https://user-images.githubusercontent.com/44557675/184457182-f9b6e2f9-301c-4c13-80c1-a129c829f890.MOV

Client/Frontend/Favicons/FaviconResolver.swift
```diff
@@ -4,8 +4,28 @@
...
+/// favicon to the site-specified favicon.
+class FaviconResolverCache {
+    static var shared = FaviconResolverCache()
+
+    // Sized so this works well for history UI, archived tabs, etc.
+    private let cache = LRUCache<URL>(maxEntries: 128)
+
+    subscript(url: URL) -> URL? {
+        get {
+            cache[url] as! URL?
```
EdwardJTL, line 21: could we make this not a force casting? feels unnatural since the return type is optional as well

Shared/LRUCache.swift
```diff
@@ -0,0 +1,121 @@
...
+                tail = entry.prev
+            }
+            entry.removeFromList()
+            insertAsHead(entry: entry)
+        }
+
+        return entry.value
+    }
+
+    private func store(for key: KeyType, value: Any?) {
```
EdwardJTL, line 78: i think these methods should be synchronized or use atomic operations. There can data errors or nil accesses if the caller(s) cause a race condition

Shared/LRUCache.swift
```diff
@@ -0,0 +1,121 @@
...
+    var count: Int {
+        map.count
+    }
+
+    public init(maxEntries: Int) {
+        self.maxEntries = maxEntries
+    }
+
+    /// Store and retrieve keyed objects from the cache.
+    public subscript(key: KeyType) -> Any? {
```
EdwardJTL, line 52: is it possible to generic version of this function that will do the casting for the caller?

Client/Frontend/Favicons/FaviconResolver.swift
darinwf: > The clipping animation issue on iOS 16 is still there (#4159)
> 
>  Upload.from.GitHub.for.iOS.MOV

Too bad, I had high hopes! I will keep that issue open :-(

Shared/LRUCache.swift
```diff
@@ -0,0 +1,121 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+
+/// A fixed-size, in-memory LRU cache.
```
EdwardJTL, line 7: maybe we should leave a note stating that
1) this cache keeps a strong reference to the inserted value
2) it does not make a copy of the key or the value
3) it does not respond to memory warnings
4) it is not thread safe (if we decide to not address my other comment about threading)

darinwf:
> The clipping animation issue on iOS 16 is still there (#4159)
> 
>  Upload.from.GitHub.for.iOS.MOV

Too bad, I had high hopes! I was seeing a very similar animation issue on my device, which this PR seemed to eliminate. The other idea I had was to replace `WebImage` with a `UIImage` that we update manually. That might be the answer for iOS 16.

Client/Frontend/Favicons/FaviconResolver.swift
darinwf: Hmm, but we know the type has to be `URL?` given how we have restricted access to `cache`. I can play around to see if there is another way.

Shared/LRUCache.swift
darinwf: Oh, I'd rather just add assertions to prevent multi-threaded access. We don't need to add the overhead of synchronization unless / until we have a need for multi-threaded access.

Shared/LRUCache.swift
darinwf: Hmm, I'm not sure how to do that w/ `subscript`.

Shared/LRUCache.swift
darinwf: Those are good call outs. Thanks!

Shared/LRUCache.swift
darinwf: Adding another template parameter to the class seems to be the way to go.

Shared/LRUCache.swift
darinwf: I think the 2nd bullet point is redundant with the 1st, so I'm just gonna mention the 1, 3 and 4.

Shared/LRUCache.swift
darinwf: I went with just restricting access to the main thread. More ideal would be to restrict to the thread the class is instantiated on, but this will do for now. We can be fancier or add proper support for multiple threads in the future if the need comes up.

EdwardJTL (approved):

PR #4204, opened 2022-08-12
Switch to building for web and pull patched page-metadata-parser

### LINKED ISSUE #4203: metadata parser not injected properly into sites; does not load site specified favicons
## Expected behavior

Go to a website that specifies a favicon that is different than the one we have baked into the product, e.g., youtube.com. Expect the app to pick up the site-specific favicon and use that going forward.

## Actual behavior

Site-specified favicons are not being loaded. Some users who have existing DBs from long ago would still have these site specified favicons because they are getting it out of this old history DB.

The problem appears to be due to the generated `MainFrameAtDocumentEnd.js`, which includes the file `url-utils.js` from the `page-metadata-parser` package. That file has as its first line `if (global.URL !== undefined) {`. The dereferencing of `global.URL` triggers an exception since `global` is not defined and you can't access properties of `undefined` in JavaScript.

As a result, the rest of `MainFrameAtDocumentEnd.js` does not load, including the rest of the `page-metadata-parser` scripts. It happens to be the last set of scripts packaged as part of `MainFrameAtDocumentEnd.js`, so other features are not impacted.

When I go back to Firefox and build this same script, I don't see this problem. `global.URL` check is not emitted directly in the output. Maybe `webpack` or something is adding some extra code to handle this properly. Maybe `global` is supposed to be replaced with either `window` or whatever the global object is for the script context. This `url-utils.js` script appears to be designed to work in either a browser context or a Node.js context.

darinwf:
Fixes #4203


evancrow (approved):
LGTM - Can you make sure Cookie Cutter, Reader Mode are still working :)

darinwf:
> LGTM - Can you make sure Cookie Cutter, Reader Mode are still working :)

Those seem to work properly in my testing

Issue #3907, opened 2022-06-23
Neeva Scope Sindow - Sign Up or Join Promt


barisertgrul:
## Expected behavior

When user click on neeva scope button, user should see recommendations

## Actual behavior

when user freshly install 1.46.2 (238.1), user can see sign up or join promt on neeva scope window

## Steps to reproduce the problem

1. user fresh install 1.46.2 (238.1)
2. user select one of the recommendations after skipping default browser set up
3. after page opens, user can see sign up or join to neeva promt after clicking on neeva scope button4. 

user's first action should be search on search bar to see scope recommendations on a web site. until user search anything, scope window shows sign up or join to neeva promt
## Screenshots or recordings

![IMG_7507](https://user-images.githubusercontent.com/17303291/175357306-c9e79583-4cdd-427e-8e67-daf469aea7d9.PNG)

## Hardware and software

- Your device:  iphone 13
- OS version: 15.5
- App version: Neeva Browser 1.46.2 (238.1)

PR #4211, opened 2022-08-12
Explicit Preview Cookie Management for Cheatsheet

EdwardJTL:
We saw on mode there is a decent chunk of preview requests
It's a known issue that preview mode only works because of a WKWebkit Bug

This change introduces an explicit control of the preview cookie used for cheatsheet.
if use is not signed in, we fall back to a preview flow. In this flow, we check for a preview cookie in the keychain. If there is no cookie in the keychain, we ping neeva.com/preview to fetch a new one. Then, we use a separate GraphQLAPI instance with a separate cookie storage to make the search request.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
on real device:
- No sign in, make cheatsheet request without visiting neeva.com first
- Then sign in, make another cheatsheet request
- Then sign out, make another cheatsheet request

- manually set an expired cookie in keychain, then make a request

Observe the header in the request and response to make sure that the cookies are correct.

## Issues addressed
#3907


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
```diff
@@ -76,41 +86,82 @@ public final class CheatsheetServiceProvider: CheatsheetDataService {
...
+        } else if (try? NeevaConstants.getPreviewCookie()) != nil {
+            NeevaScopeSearch.SearchController.getRichResult(
+                using: self.previewClient,
+                query: query,
+                completion: completion
+            )
+        } else {
+            self.cookieFetcher.fetch { [self] result in
+                // return to main to set cookie and initiate request
+                DispatchQueue.main.async {
```
darinwf, line 155: nit: How about putting this `DispatchQueue.main.async` call inside the implementation of `cookieFetcher`? That way users don't have to think about threads.

darinwf (approved):
LGTM, I think @macy-n should also take a look

macy-n (approved):

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
EdwardJTL: I wanted the caller to enforce it because it is the caller who needs main thread to do the work here. main thread is not an intrinsic requirement for the fetch callback

EdwardJTL:
decided to add a retry logic to handle expired cookie case.
```
reads preview cookie
try with that one
if fails, re-fetch preview cookie
try one last time
```

also I am now removing leaked preview cookie in shared graphQL client on each request.

@darinwf requesting a review for those changes 🙏

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
darinwf: I see. The advantage of doing it in the implementation is that it then allows callers to not need to think about multi-threading. As is, you need to make sure there are bright blinking neon lights on the definition of `fetch` telling people that the callback might happen on a random background thread. Otherwise it is a big footgun for unsuspecting callers. The only reason to allow the footgun is because it is some significant optimization. Is there such here?

Shared/API/Network.swift
```diff
@@ -348,6 +348,11 @@ class LogInCookieIntercept: ApolloInterceptor {
             // Else, let this request fail with an authentication error.
             client.cookieStorage?.removeLogInCookies()
         }
+
+        // WKWebKit can leak cookies into URLSessions if both are using persistent
+        // cookie storages on physical devices. Remove these preview cookies
```
darinwf, line 353: Do we have a webkit bug or radar issue we can reference here?

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
EdwardJTL: i allowed it because this class is private and for this use case, the accesses are synced on on `main` due to the keychain accesses before and after.

I could add synchronization in the fetcher too so that this class may be safely used eventually. i will still prefer to have the fetcher work on a background by default, and the caller can decide if they want to reschedule the work on main in the completion. How does that sound?

Shared/API/Network.swift
EdwardJTL: they have a few related bugs on webkit bugs
https://bugs.webkit.org/buglist.cgi?quicksearch=HTTPCookie
i can't find radar anymore...

darinwf (commented):
Looks good. Nice catch on the need for synchronization here.

As discussed, please add more documentation around the synchronization, why it matters, what someone should have in their head when reading this code, etc. Also would be nice to add a DispatchQueue parameter to fetch so that callers can specify that, and it helps someone not think they don't need to worry about what thread the callback will run on.

PR #4214, opened 2022-08-13
[Xcode 14] Disable Bitcode

EdwardJTL:
Apple finally gave up on bitcode. With Xcode 14, if you build with bitcode, you can't upload to app store connect.


darinwf (approved):
Is there any downside here?

EdwardJTL:
> Is there any downside here?

I think it will be worth to check that we can still build and upload via Xcode 13 with this change.
Bitcode support is deprecated and expected to be dropped completely in the future. You cannot upload with Xcode 14 if you build with bitcode. See apple's [release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes)

PR #4224, opened 2022-08-15
welcome flow; fix possible index out of range exception

### LINKED ISSUE #4223: First Run Crash @ Indexing activeTabs
On main.
Screen Shot 2022-08-15 at 10 05 46 AM

I believe this is because UserInfo is initialized with an optimistic "true" value, which allows this code to force index into 0

Steps:
- Clean install onto simulator with blank keychain and no app in app group
- don't sign in, pick free plan, pick do it later for default browser

jedireza:
Fixes #4223

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

N/A

## Issues addressed
N/A


Client/Frontend/WelcomeFlow/WelcomeFlowModel.swift
```diff
@@ -56,7 +56,7 @@ class WelcomeFlowModel: ObservableObject {
         self.onCloseAction?()
         self.onCloseAction = nil
 
-        if NeevaUserInfo.shared.isUserLoggedIn {
+        if authStore.bvc.tabManager.activeTabs.indices.contains(0) {
```
EdwardJTL, line 59:
```
if let tab = authStore.bvc.tabManager.activeTabs.first {
   authStore.bvc.tabManager.selectTab.selectTab(tab, notify: false)
}
```

Client/Frontend/WelcomeFlow/WelcomeFlowModel.swift
```diff
@@ -56,7 +56,7 @@ class WelcomeFlowModel: ObservableObject {
...
             authStore.bvc.tabManager.selectTab(
                 authStore.bvc.tabManager.activeTabs[0], notify: false)
```
EdwardJTL, line 61: is there a reason this does not notify? we have UI that depends on the selected tab publisher/call backs

jedireza: Yes. Because it would interrupt the welcome flow experience.

EdwardJTL (approved):
My opinion is that we should still send the tab update at some point. Perhaps the welcome flow model should be reworked so that the `complete` method is the last method called to complete the flow, such that notifying the tab selection won't "disrupt the flow". 

Seeing that this PR is aimed to just fix the crash, and that the `notify: false` behaviour is already in the code, we can proceed with this for cherry picking.

PR #4228, opened 2022-08-15
FeatureFlag HoverEffects

evancrow:
Right now, the `HoverEffect` experience kinda sucks. I have an open question on the [Apple Dev Forums](https://developer.apple.com/forums/thread/712412?answerId=724148022#724148022), but for now, I see it best to put this behind a flag until it can work better.

In an attempt to debug this, I created a `HoverEffectButton`, and figured I might as well leave it for now.


Shared/UI/Util/HoverEffectButton.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    let effect: HoverEffect
+    let isTextButton: Bool
+    let action: () -> Void
+    let label: Label
+
+    var button: some View {
+        Button {
+            if UIDevice.current.userInterfaceIdiom == .pad, useHoverEffect {
+                useHoverEffect = false
```
EdwardJTL, line 18: i think this is giving me `Client[976:84011] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.` every time i hover over a button

EdwardJTL: maybe just leave a comment for this so we will consider that when we fix up the feature?

EdwardJTL (approved):

Shared/UI/Util/HoverEffectButton.swift
evancrow: Just gonna nuke that code since I don't like using the delay

PR #4229, opened 2022-08-15
Fix ad blocker rules and update easy_priacy and easy_list rules

chung-lai:
* Turns out all of our current easy list rules are blocking cookies instead of request. Updated to the latest easy list and change the rule from `block-cookies` to `block` so it blocks the actual request
* Updated the conversion script to convert the rules properly (happen on different repo)
* Updated the easy_privacy rules with the latest
* Add support of the `ping` resource type

With the conversion script update, we should start updating the rules more regularly. That would be the next step.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


darinwf:
Oh, good catch! We'll have a much improved ad blocker now :-D

darinwf (approved):

PR #4231, opened 2022-08-15
welcome flow; fix collection toggle persistence

jedireza:
Fixes the persistence of the value

### How was this tested?
N/A

### Manual test cases
N/A

N/A

## Issues addressed
N/A


Client/Frontend/WelcomeFlow/WelcomeFlowIntroView.swift
```diff
@@ -103,7 +103,7 @@ struct WelcomeFlowIntroView: View {
                 }
                 .buttonStyle(PlainButtonStyle())
                 .onChange(of: collectUsageStats) { newValue in
-                    Defaults[.shouldCollectUsageStats] = collectUsageStats
+                    Defaults[.shouldCollectUsageStats] = newValue
```
chung-lai, line 106: hmm...this will set it to the new value but the bug is more about we should set a default value after user passes the first screen if it hasn't got set. Otherwise, `Defaults[.shouldCollectUsageStates]` will always be nil

chung-lai (approved):

PR #4234, opened 2022-08-15
Add Pin Tab Option in ShowTabsMenu

evancrow:
Idea from @jedireza in Design Weekly.

@chung-lai I also updated the casing of some `localizedStrings`. Not sure if that's okay.

### How was this tested?
- [x] I added automated tests

https://user-images.githubusercontent.com/37356984/184755492-f4193ab2-f20c-4faf-a868-31269ba7161f.mp4


Client/Frontend/Browser/MenuBuilder.swift
```diff
@@ -212,6 +212,27 @@ class ContextMenuActionsBuilder {
...
+
+    // MARK: - Pin Tab
+    struct TogglePinnedTabAction: View {
+        let tabManager: TabManager
+        let tab: Tab
+        var isPinned: Bool
+
+        var body: some View {
+            Button { [self] in
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.8) { [self] in
```
EdwardJTL, line 224: `self` is a struct here so we could omit the capturing.

EdwardJTL, line 224: could we get a comment on what problem this delay is remediating?

Client/Frontend/Browser/MenuBuilder.swift
```diff
@@ -291,8 +312,13 @@ class MenuBuilder {
                     bvc.openLazyTab(openedFrom: .newTabButton)
                 }
 
-                ContextMenuActionsBuilder.CloseTabAction {
-                    tabManager.removeTab(tabManager.selectedTab)
+                if let selectedTab = tabManager.selectedTab {
```
EdwardJTL, line 315: this `if let` check is new right? I'm trying to understand why this check was not needed before...

EdwardJTL (approved):
just some nits

PR #4242, opened 2022-08-16
Fix PinnedTab "Twitter Case"

evancrow:
The bug was caused by the Twitter profile URL making its way into the `SessionData` URL stack without the URL being set on the `Tab` yet. This prevents this by removing the URL if it hasn't yet been officially loaded in the `Tab`.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Steps to repro "Twitter Case" prior this PR:

- Sign in to twitter
- Go to twitter.com/home
- Pin tab
- Tap a profile
- Go back to pinned tab placeholder
- The profile is loaded instead of the home page


Client/Frontend/Browser/Tabs/Tab/Tab.swift
```diff
@@ -793,13 +793,23 @@ class Tab: NSObject, ObservableObject {
...
+            var navigationStackIndex = webView?.backForwardList.navigationStackIndex ?? 0
+            var queries = navigationList.map {
                 queryForNavigation.findQueryFor(navigation: $0)
             }
 
+            // When creating SessionData for a pinned tab's placeholder, a
+            // URL will be added to SessionData even if it hasn't yet been set for the tab.
+            // This prevents the URL from being included in the SessionData
+            // since it's not meant to be used in the placeholder.
+            if urls.last != url {
```
darinwf, line 807: couldn't this have other side-effects though?

evancrow: Yes, it definitely could. Going to let tests run a few times and test them myself. I think it should be safe in all different cases though, `setURL` is called before we create the `SessionData`.

evancrow: Improved the function caller, only trim the URL if working, making `SessionData` for the pinned tab.

darinwf (approved):
Good call! Nice to scope this down to just this flow.

Issue #4247, opened 2022-08-16
Close button on tab should animate with the tab


darinwf:
The close button should not appear as a disjoint UI element. It should appear glued to the top right corner of the tab.

https://user-images.githubusercontent.com/79867437/184996512-89e578ea-830c-4011-9776-98a20a2215ad.mp4

Observed with 1.53.4

PR #4250, opened 2022-08-17
Animate Close Button With Tab

evancrow:
### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/185193330-516a06d5-94be-4642-abf8-d28f408520ac.mp4

## Issues addressed
Fixed: #4247


Client/Frontend/Browser/CardGrid/Card/Card.swift
```diff
@@ -283,4 +267,47 @@ struct Card<Details>: View where Details: CardDetails {
...
+                    .opacity(animate && !showGrid ? 0 : 1)
+                    .transition(.identity.combined(with: .opacity))
+                    .animation(CardTransitionUX.animation)
+            }.accessibilityHidden(true)
+        }
+
+        func body(content: Content) -> some View {
+            if details.closeButtonImage != nil {
+                Group {
+                    if #available(iOS 15.0, *) {
```
darinwf, line 298: Is this iOS 15 method material to the solution?

I'm wondering if we can fix the bug without introducing the iOS 15+ specific code branch.

darinwf (commented):
Thanks for jumping on this. It seems like there is some hysteresis effect on the close button as the card transition animation is going from small to large, but otherwise this seems like a huge improvement. Can you say more about how the solution works? The transition animation seems like the key bit, right?

evancrow:
Two solutions:

- Yes, the `animation` and `transition` modifiers made this possible. Before, the opacity would just go from 0 to 1 (and vice versa) whenever the `CardTransition` was completed. Now it animates with it, using the same animation.
- Removing the `.if` modifier. It wasn't the main solution, but the animation was certainly smoother without it.

darinwf (approved):

PR #4252, opened 2022-08-18
auth store qr code support + integrate into welcome flow

jedireza:
This PR adds sign-in with QR code functionality to the `AuthStore` and implements the feature (still behind a flag) into to the welcome flow sign-in screen.

I also extended the welcome flow model to support a stack of history. This helps the user get back to where they started. Something I noticed specifically in this QR code support.

I left comments in the code, but we really need to invest in not requiring a browser tab to fully authenticate.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator and also on my physical device.

Screen Shot 2022-08-17 at 20 18 12

## Issues addressed
N/A


Client/Frontend/WelcomeFlow/WelcomeFlowSignInView.swift
```diff
@@ -81,12 +81,22 @@ struct WelcomeFlowSignInView: View {
...
                     }
                 ) {
-                    model.prevScreen = nil
+                    model.clearPreviousScreens()
                     model.changeScreenTo(.defaultBrowser)
                 }
             }
 
-            // TODO: sign-in with QR code
+            if FeatureFlag[.qrCodeSignIn] {
```
jedireza, line 89: To test drive this, you can simply make a change like this and then run the build.

```suggestion
            if FeatureFlag[.qrCodeSignIn] || true {
```

Client/Frontend/WelcomeFlow/WelcomeFlowModel.swift
```diff
@@ -13,13 +13,14 @@ enum WelcomeFlowScreen: String, CaseIterable {
...
     case signUpEmail
     case signIn
+    case signInQRCode
 }
 
 class WelcomeFlowModel: ObservableObject {
     var onCloseAction: (() -> Void)?
     var authStore: AuthStore
     @Published var currentScreen: WelcomeFlowScreen = .intro
-    @Published var prevScreen: WelcomeFlowScreen?
```
chung-lai, line 22: nit: `prevScreens` since it's an array of screens now?

chung-lai (approved):
LGTM. would be great if @yasminyin96 can take a look at the QR sign in flow

yasminyin96 (approved):
LGTM, thank you

PR #4253, opened 2022-08-18
Move archived tabs out of TabManager.tabs

darinwf:
With this change, operations on `tabs` is now independent of the
number of archived tabs, which should improve performance.

Introduces `ArchivedTab` type and `GenericTab` as a protocol shared by
`Tab` and `ArchivedTab`. `ArchivedTab` is a thin wrapper around `SavedTab`.

Doesn't change the way we persist tabs, including archived tabs,
but does pave the way for that as a follow-up.

Also flattens out the directory structure under `Browser/Tabs` to be more
manageable on the command line :-)

A side-effect of this PR is that the createOrSwitchToTab method will no longer
consider archived tabs. That's a performance win and paves the way for these
not needing to be kept in memory, but it is also a change in behavior. The plan
is to experiment with this behavior change and see how it works out.


Client/Frontend/Browser/Tabs/ArchivedTab.swift
```diff
@@ -0,0 +1,26 @@
...
+    var tabUUID: String { savedTab.UUID ?? "" }
+    var rootUUID: String { savedTab.rootUUID ?? "" }
+
+    let isPinned = false  // Archived tabs cannot be pinned
+    let parentSpaceID: String? = nil  // Archived tabs cannot have a parent
+
+    var id: String { tabUUID }
+
+    init(savedTab: SavedTab) {
+        self.savedTab = savedTab
```
evancrow, line 24: Instead of doing calculated variables, could be better to `map` on `init`.

Client/Frontend/Browser/Tabs/ArchivedTab.swift
```diff
@@ -0,0 +1,26 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+import Shared
+import Storage
+
+class ArchivedTab: NSObject, GenericTab {
+    let savedTab: SavedTab
```
evancrow, line 10: Maybe it's best to save all tabs as `ArchivedTabs` and restore them to regular `Tabs`. Removes the need for `SavedTab`.

Client/Frontend/Browser/Tabs/GenericTab.swift
```diff
@@ -0,0 +1,43 @@
...
+    var displayTitle: String { get }
+    var lastExecutedTime: Timestamp { get }
+    var isPinned: Bool { get }
+
+    var tabUUID: String { get }
+
+    /// All tabs with the same `rootUUID` are considered part of the same group.
+    var rootUUID: String { get }
+
+    var parentSpaceID: String? { get }
```
evancrow, line 20: Don't all tabs also have a `Parent`, and `ParentUUID`?

Client/Frontend/Browser/Tabs/GenericTab.swift
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func shouldBeArchived(basedOn lastExecutedTime: Timestamp) -> Bool {
+    func isIncluded(in section: TabSection) -> Bool {
+        section.includes(isPinned: false, lastExecutedTime: lastExecutedTime)
+    }
+
+    switch UserDefaults.standard[.archivedTabsDuration] {
```
evancrow, line 28: nit: Use `Defaults[.archivedTabsDuration]`

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -206,26 +195,8 @@ class Tab: NSObject, ObservableObject {
...
-            return
-                !(isIncluded(in: .today)
-                || isIncluded(in: .yesterday)
-                || isIncluded(in: .lastWeek)
-                || isIncluded(in: .lastMonth))
-        case .forever:
-            return false
-        }
+    var shouldBeArchived: Bool {
+        !isPinned && Client.shouldBeArchived(basedOn: lastExecutedTime)
```
evancrow, line 199: Think you call this function without saying `Client.`

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -509,47 +508,47 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
-            + normalTabs.filter {
-                return !$0.isArchived
-            }
-
-        if notify {
-            tabsUpdatedPublisher.send()
-        }
-    }
-
+    // TODO: make this function private
```
evancrow, line 511: :)

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -509,47 +508,47 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
-
-        archivedTabs = normalTabs.filter {
-            return $0.isArchived
-        }
-
-        activeTabGroups =
-            activeTabs
-            .reduce(into: [String: [Tab]]()) { dict, tab in
+        // TODO: how should we handle the case of the selectedTab being old?
+        assert(!(selectedTab?.shouldBeArchived ?? false))
```
evancrow, line 514: Unless this is fixed in this PR, this is probably going to cause some crashes when the app is opened after being in the background overnight.

Client/Frontend/Browser/Tabs/TabManagerStore.swift
```diff
@@ -168,18 +164,23 @@ class TabManagerStore {
...
 
         for savedTab in savedTabs {
-            // Provide an empty request to prevent a new tab from loading the home screen
-            let tab = tabManager.addTab(
-                flushToDisk: false, zombie: true, isIncognito: savedTab.isIncognito, notify: false)
-            savedTab.configureTab(tab, imageStore: imageStore)
+            let shouldBeArchived =
+                !savedTab.isPinned
+                && shouldBeArchived(
+                    basedOn: savedTab.lastExecutedTime ?? Date.nowMilliseconds())
```
evancrow, line 173: Should make `SavedTab.lastExecutedTime` non-optional (not necessarily this PR)

Client/Frontend/Browser/Tabs/ArchivedTab.swift
darinwf: Tell me more about what you are thinking. How would you use `map` here? I figure these computed properties are super cheap and likely inlined away by the compiler.

Client/Frontend/Browser/Tabs/ArchivedTab.swift
darinwf: Could do, but `SavedTab` is the "plain old data" structure we use for serialization. I was trying to not mess with the serialization of a tab for now. `ArchivedTab` gives us a way to express just what we need at runtime to represent an archived tab. `SavedTab` has a lot more going on, including the ability to mutate properties. `ArchivedTab` gives us a more narrow type.

Client/Frontend/Browser/Tabs/GenericTab.swift
darinwf: Yeah, but I didn't need to expose these. This is just what is needed to support `GenericTabGroup` and other code in this PR. We can add more properties to `GenericTab` as we need them.

Client/Frontend/Browser/Tabs/GenericTab.swift
darinwf: Thanks, will do

Client/Frontend/Browser/Tabs/Tab.swift
darinwf: Actually, the compiler complained and suggested this solution :-/

Client/Frontend/Browser/Tabs/TabManager.swift
darinwf: Yeah, I'm worried about it to. What it means is that the selected tab will get put in the archive and we will be left in a bad state, with `selectedTab` pointing to a `Tab` that is no longer part of `tabs`. I'm not sure if this can happen in practice though. The way I hit this scenario was with the debug settings `archiveAllTabs` command. I'll spend some more time on this and see what I should do.

Client/Frontend/Browser/Tabs/TabManagerStore.swift
darinwf: Can't really change the definition of `SavedTab` since it corresponds to what is persisted in storage. A user upgrading from an older version of the app won't have this field in their `SavedTab` instance. I mean, maybe we can do something to encapsulate the defaults. I'm a little torn on how rich to make `SavedTab` though since it is kind of just "plain old data".

evancrow: I'm just saying you could do something similar to what I did with `Tab.lastExecutedTime` (`var lastExecutedTime: Timestamp = Date.nowMilliseconds()`). Just move the default value to be on `init` so you don't have to call it outside of `SavedTab`.

Client/Frontend/Browser/Tabs/Tab.swift
evancrow: Huh, alright.

Client/Frontend/Browser/Tabs/ArchivedTab.swift
evancrow: Oh, that's true. We would need to create some sort of migration code. Fine to leave this as is, for now, can look into it later if want.

Client/Frontend/Browser/Tabs/TabManagerStore.swift
darinwf: Right, but I think the storage type would need to still be `Timestamp?` because of the `encode` call. I don't want to change the object type that is used for persistence. If the member stays optional, then callers still have to deal with it being an optional. I agree we could probably find a way to encapsulate all of this and simplify callers. I don't think that is trivial so I'd rather add / explore as part of a separate PR.

Client/Frontend/Browser/Tabs/ArchivedTab.swift
evancrow: I shouldn't have said `map` not what I meant actually. I meant it could be simpler to set all the variables once on `init` (`self.x = x`). That way, it's set once, and we don't have to calculate it multiple times. Because it's just wrapping the variables it's fine to leave it as is too, I agree it's lightweight.

darinwf: OK.. yeah, plus if the underlying `SavedTab` were to change for some reason (not that we do that), it would not be great to have to update these fields.

Client/EnvironmentHelper.swift
```diff
@@ -199,7 +199,7 @@ class EnvironmentHelper {
         TabManager.all.forEach { tabManager in
             // In the future, we may also want to log archived tab groups.
             numOfTabGroups += tabManager.activeTabGroups.count
-            numOfChildTabs += tabManager.childTabs.count
+            numOfChildTabs += tabManager.activeTabGroups.values.flatMap(\.children).count
```
EdwardJTL, line 202: suggestion:
`tabManager.activeTabGroups.values.map(\.children.count).reduce(0, +)`

Client/Frontend/Browser/Tabs/ArchivedTab.swift
```diff
@@ -0,0 +1,26 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+import Shared
+import Storage
+
+class ArchivedTab: NSObject, GenericTab {
```
EdwardJTL, line 9: nit:
could we separate conformance to an extension? like
```
class ArchivedTab: NSObject {
let savedTab: SavedTab
init(){...}
}

extension ArchivedTab: GenericTab {
...
}
```

Client/Frontend/Browser/Tabs/ArchivedTab.swift
```diff
@@ -0,0 +1,26 @@
...
+    let savedTab: SavedTab
+
+    var url: URL? { savedTab.url ?? savedTab.sessionData?.currentUrl }
+    var displayTitle: String { savedTab.title ?? "" }
+    var lastExecutedTime: Timestamp { savedTab.lastExecutedTime ?? Date.nowMilliseconds() }
+    var tabUUID: String { savedTab.UUID ?? "" }
+    var rootUUID: String { savedTab.rootUUID ?? "" }
```
EdwardJTL, line 16: when you restore these tabs, would the fallback `""` string cause the tabs to be grouped? should we create a new UUID if one is not found in savedtab?

Client/Frontend/Browser/Tabs/ArchivedTab.swift
```diff
@@ -0,0 +1,26 @@
...
+import Shared
+import Storage
+
+class ArchivedTab: NSObject, GenericTab {
+    let savedTab: SavedTab
+
+    var url: URL? { savedTab.url ?? savedTab.sessionData?.currentUrl }
+    var displayTitle: String { savedTab.title ?? "" }
+    var lastExecutedTime: Timestamp { savedTab.lastExecutedTime ?? Date.nowMilliseconds() }
+    var tabUUID: String { savedTab.UUID ?? "" }
```
EdwardJTL, line 15: since this is used for `Identifiable` conformance, maybe it should also not default to the same empty string. else it may crash swiftUI

Client/Frontend/Browser/Tabs/GenericTab.swift
```diff
@@ -0,0 +1,43 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Defaults
+import Foundation
+import Shared
+import Storage
+
+protocol GenericTab: Identifiable {
```
EdwardJTL, line 10: both `ArchivedTab` and regular `Tab` use `tabUUID` to conform to `Identifiable`. should we provide that as a default implementation of `GenericTab`? or somehow force that to be the case?

Client/Frontend/Browser/Tabs/Tab.swift
EdwardJTL: you need to make that a static function maybe?

Client/Frontend/Browser/Tabs/TabManager.swift
darinwf: OK, I did some testing. On a cold start, we already do the right thing. `TabManager.restoreTabs` takes care to update the selected tab as needed before calling `updateAllTabDataAndSendNotifications`.

Client/EnvironmentHelper.swift
darinwf: that's a much more performant suggestion, thanks!

darinwf:
@EdwardJTL - re: "you need to make that a static function maybe?", only methods declared on a type can be made static. I agree the `Client.` prefix is unusual and would be nice to avoid. I also like using this name for it and the same name for the `Tab` method. Hmm...

Client/Frontend/Browser/Tabs/GenericTab.swift
darinwf: Good thinking. However, ArchivedTab is really just a wrapper around SavedTab, but hmm... let me give this some thought.

Client/Frontend/Browser/Tabs/ArchivedTab.swift
darinwf: Yes... good thinking. It should not be possible for tabs to lack an ID. I guess if they do, we should just make one up. Might require some extra work though as we should probably update the `SavedTab` ID in that case. Will think on it...

Client/Frontend/Browser/Tabs/ArchivedTab.swift
darinwf: OK... yeah, I guess this is the more canonical way with Swift? :)

darinwf:
@EdwardJTL @evancrow - OK, I decided to bite the bullet and make `tabUUID` and `rootUUID` non-optional on `SavedTab`. It looks like it is calling the same `encode` method, so this should hopefully not impact the serialization format. Don't want to screw things up too badly for people who might want to migrate back to an older version of the app.

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -508,47 +507,51 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
-
-        if notify {
-            tabsUpdatedPublisher.send()
-        }
-    }
-
+    // TODO: make this function private
     internal func updateAllTabDataAndSendNotifications(notify: Bool) {
-        updateActiveTabsAndSendNotifications(notify: false)
+        // Handle the case of the selected tab needing to be archived. This can happen
```
darinwf, line 512: @evancrow - This is how I solved the problem of a selected tab being old. Let me know what you think. I considered making no selected tab, but the result is that you can still see the old selected tab in the task switcher on iPhone, but when you switch to the app, it seems wrong to suddenly see that tab disappear. So I think this is the better solution. It is less jarring for the user.

Client/Frontend/Browser/CardGrid/Card/CardModels.swift
```diff
@@ -72,6 +72,8 @@ class TabCardModel: CardDropDelegate, CardModel {
     private(set) var tabsDidChange = false
 
     private func updateRows() {
+        needsUpdateRows = false
```
darinwf, line 75: This was needed to make sure we didn't call `updateRows` without a prior call to `updateAllTabDataAndSendNotifications`. What was happening is that the flag was being set when the selected tab changed, but it wouldn't be necessarily cleared the next time `updateRows` was called. It would only be called if `updateIfNeeded` is called, which happens when clicking on the "show tabs" button. It makes sense to clear this flag whenever `updateRows` is called since that should satisfy the intent of `needsUpdateRows`.

This makes it so that if the selected tab has an old age, we won't rebuild the CardDetails before getting a chance to revise the lastExecutedTime of the selected tab.

Client/Frontend/Browser/Tabs/TabManager.swift
evancrow: Yes, this makes sense. Should we always do this for the `selectedTab` even if `shouldBeArchived` is false?

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -508,47 +507,51 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
+        for tab in tabsToArchive {
+            tab.closeWebView()
+            // Discard tabIndex at this point. When restored, we'll simply append to activeTabs.
+            archivedTabs.append(
+                ArchivedTab(
+                    savedTab: tab.saveSessionDataAndCreateSavedTab(isSelected: false, tabIndex: nil)
+                ))
+        }
+        if !tabsToArchive.isEmpty {
+            tabs = tabs.filter { $0.isIncognito || !$0.shouldBeArchived }
```
evancrow, line 531: Think this is easier to read as `tabs.removeAll { tabsToArchive.contains($0) }`

Client/Frontend/Browser/Tabs/TabManager.swift
darinwf: Well, that's a good question. It could make sense. Otherwise the selected tab could become a yesterday tab, which might be right or might be confusing. Hmm...

Client/Frontend/Browser/Tabs/TabManager.swift
darinwf: It is easier to read, but it is also O(N x M) vs. O(N). If helpful to not repeat the code for the filter expression, I could move that into a helper function and then just invert it for this one.

Client/Frontend/Browser/Tabs/TabManager.swift
darinwf: @evancrow - Thinking about this some more, I'd like to leave it as-is for now and then explore what should happen with the selected tab moving from today to yesterday separately. The code added here is important because without it, the selected tab would be removed from `tabs`, which would create a broken state (non-nil `selectedTab` pointing to a `Tab` that is not part of `tabs`). The other issue is interesting but not a new issue created with this PR, so it can be punted to another PR. This PR is already really big.

Client/Frontend/Browser/Tabs/TabManager.swift
darinwf: I went with defining a shared condition (callback) and used it for both the filter call to generate `tabsToArchive` as well as in a `removeAll` call as you suggested. PTAL.

Client/Frontend/Browser/Tabs/GenericTab.swift
darinwf: Done by updating `SavedTab` to provide the fallbacks.

Client/Frontend/Browser/Tabs/Tab.swift
darinwf: Unfortunately can't have static functions at file scope.

Client/Frontend/Browser/Tabs/TabManagerStore.swift
darinwf: Just to update, I did change `SavedTab` to have non-optional `rootUUID` and `tabUUID`, but I couldn't change `lastExecutedTime` similarly. Reason is that for a regular `Tab` the right fallback timestamp is `Date.nowMilliseconds` so it will become a `.today` tab, but for archived tabs, the right fallback timestamp is `0` (distant past), so that it 1) remains an archived tab and 2) is in the oldest section. So I kept the fallback code for `lastExecutedTime` in `Tab` and `ArchivedTab` respectively.

Client/Frontend/Browser/Tabs/TabManager.swift
evancrow: Yes, okay. This function looks good, and this way we focus on finding a proper solution. 

I remember a little while ago, I added code that would move the `selectedTab` from yesterday into today when the app entered foreground but I can't seem to find it anymore.

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -583,28 +588,12 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
-        // Write newly created tab group names into dictionary
-        tabGroups.forEach { group in
-            let id = group.key
-            if Defaults[.tabGroupNames][id] == nil {
-                Defaults[.tabGroupNames][id] = group.value.displayTitle
-            }
-        }
-
-        // Garbage collect tab group names for tab groups that don't exist anymore
+    private func updateTabGroupNames() {
```
evancrow, line 591: Not for this PR, but I wonder why we don't assign names directly onto the `TabGroup` type.

evancrow (approved):
This LGTM!!

Client/Frontend/Browser/Tabs/ArchivedTab.swift
EdwardJTL: It makes it more clear which part of the class is used to conform to the protocol, and which part is actually needed to the class itself. When we change/remove the protocol, it will also be easier to modify the code here.

Client/Frontend/Browser/Tabs/Tab.swift
EdwardJTL: oh i think its because there is a conflict in the symbols. there is a `shouldBeArchived` in this class, so in this scope the compiler prioritizes the instance method.

Client/Frontend/Browser/Tabs/TabManager.swift
darinwf: I think that would be a better API for sure. The current approach has to do, I think, with allowing users to customize a tab group's name. So it needs to be serialized. Also, TabGroup objects get recreated over and over again (which is probably not ideal for performance).

PR #4255, opened 2022-08-18
add easylist to content blocking stats reporting

### LINKED ISSUE #4254: Ad blocking rules are not being reported in the content blocking stats
Currently, we hardcode easyprivacy list for content blocking stats reporting. We should add easy list in the case user enables ad blocking.

chung-lai:
Fix #4254

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


content-blocker-lib-ios/src/ContentBlocker.swift
```diff
@@ -50,6 +50,18 @@ enum BlocklistFileName: String, CaseIterable {
         }
         return fileList
     }
+
+    static func statsListsForMode(strength: BlockingStrength) -> [BlocklistFileName] {
+        var blockListFiles = BlocklistFileName.listsForMode(strength: BlockingStrength(rawValue: Defaults[.contentBlockingStrength]) ?? .easyPrivacy)
```
darinwf, line 55: swift-format?

darinwf (approved):

content-blocker-lib-ios/src/ContentBlocker.swift
chung-lai: Looks like swift-format only look at the `Client` directory but not `content-blocker-lib-ios` 😆

chung-lai: Let me add the directory to the `swift-format-dirs.txt` and lets see how it goes

darinwf (approved):

Issue #4183, opened 2022-08-09
Avoid using `@Default` outside of SwiftUI `View` implementations


darinwf:
Avoid using `@Default` outside of SwiftUI `View` implementations

If you just need to read/write from `UserDefaults`, then use this shorthand provided by the Defaults package:

```
UserDefaults.standards[.keyName] = value
let otherValue = UserDefaults.standard[.otherKeyName]
```

That way we avoid setting up all of the observer notifications machinery setup by `@Default`.

darinwf:
See https://github.com/neevaco/neeva-ios/pull/4181 for an example of this kind of change.

PR #4256, opened 2022-08-18
Remove @Defaults

evancrow:
Some callers have to stay since the `ObservedObject` wrapper is important.

## Issues addressed
Fixed: #4183


darinwf (approved):
Thanks! I see that `Defaults[key]` is better than `UserDefaults.standard[key]`. Will use that from now on.

PR #4258, opened 2022-08-18
Enable PinnedTabImprovment Flags

evancrow:
Enables:
- PinnedTabImprovments
- PinnedTabSection

### How was this tested?

Approved by design


darinwf (commented):
I think pinned tab section is ready to go, but can we do some more live testing of `pinnedTabImprovments` first? I'd like to feel the impact of the bug fix you landed, etc. and make sure there aren't other oddities.

evancrow:
Yeah, sure. We can wait on pinnedTabImprovments for another week.

darinwf (approved):

Issue #4266, opened 2022-08-20
Regression: cancel button does not animate with the url bar properly


darinwf:
See video, slow down the playback speed and focus on the "cancel" button:
https://user-images.githubusercontent.com/79867437/185753954-c4b91453-3bdd-43f8-a9f7-b045c19f213a.MP4

Tap on the URL bar
Notice the "cancel" button snap into view
Then notice the input field change size
The share button also does not animate properly

The UI elements should feel like they are connected. As the cancel button comes in it should appear to squish down the input field and the share button should not move out of sync with the input field.

## Hardware and software

- Your device: iPhone 12
- OS version: iOS 15.6.1
- App version: Neeva Browser 1.55.1 (260)

darinwf:
Appears this was caused by 0d46450b339649d8bb60f7bff394f9de306d5c70

@evancrow FYI

PR #4268, opened 2022-08-22
Fix TopBarButton Animation

evancrow:
Removed unnecessary modifiers.

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #4266


darinwf (approved):

darinwf:
Thanks for the quick fix!

PR #4269, opened 2022-08-22
Refactor LongPressButton Gesture

evancrow:
Improves reliability and get's rid of the delay.

### How was this tested?


darinwf (approved):
Thanks for getting rid of the `asyncAfter` :-D

PR #4271, opened 2022-08-22
Disable Card Close Button When Animate is False

evancrow:
Fixes a bug where the close button would animate while scrolling the CardGrid. Should only animate with the `CardTransitionModel`.

### How was this tested?
- [x] I tested this manually


darinwf:
Can you add a description to this PR? What bug is this fixing? Thanks!

evancrow:
Updated!

darinwf:
How do I repro the bug being fixed?

evancrow:
Trying to get a screen recording, flakey, but here's the repro (latest TestFlight):

- Lots of tabs in CardGrid
- Scroll top/bottom & vice versa:
- Close button for every tab will sometimes slide in

darinwf (approved):
Thanks, makes sense.

PR #4273, opened 2022-08-22
Filter "removed" and "deleted" reddit posts

EdwardJTL:
If a post does not have comments, and the OP's post body is now "[removed]" or "[deleted]", we won't show that entry on the front end because there is really nothing to see there.


darinwf (approved):
Is the bloom filter construction also taking this into account?

PR #4274, opened 2022-08-22
Fix GroupedOverlayStyle PopoverView

evancrow:
Improves padding of GroupedOverlayStyle.

### How was this tested?
- [x] I tested this manually

Before:
before

After:
after


darinwf (approved):
The rest of the dialogs that use this style work well too?

evancrow:
Yup, `OpenInApp`, `SpaceFilter`, etc. all good!

PR #4279, opened 2022-08-22
Enable Drag and Drop of Pinned Tabs

### LINKED ISSUE #2975: support drag-n-drop for pinned tabs
- [ ] onDrag not showing preview image

Would be nice to have:

- [ ] animation support
- [ ] scroll the scroll view if item is dragged above/below scroll view
- [ ] UX indicator that shows which position a tab is getting inserted (now only show preview result)

evancrow:
- Enables dragging of pinned tabs in `CardGrid` and `CardStrip`.
- Ensures singular pinned tabs appear before pinned tabs in a `TabGroup` in `CardStrip`.

### How was this tested?
- [x] I tested this manually

Dragging:

https://user-images.githubusercontent.com/37356984/186036167-8b692741-aa80-4c16-a4e3-70ef41e811d6.mp4

Strip ordering correction:
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-08-22 at 16 18 14

Fixes #2975


Client/Frontend/Browser/CardGrid/Strip/CardStripCard.swift
```diff
@@ -94,17 +94,14 @@ struct CardStripCard<Details>: View where Details: TabCardDetails {
 
     var body: some View {
         card
+            .onDrop(of: ["public.url", "public.text"], delegate: details)
```
darinwf, line 97: Moving this up here helps avoid wonkiness with the same gesture being used for context menu activation?

darinwf (approved):

Client/Frontend/Browser/CardGrid/Strip/CardStripCard.swift
evancrow: Nope, it was for debugging, and I meant to return it to the original location. Thx for flagging.

Shared/Extensions/BoolExtensions.swift
```diff
@@ -0,0 +1,11 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+
+extension Bool {
+    public var binaryValue: Int {
```
EdwardJTL, line 8: nit: could you rename this `asInt`?

PR #4280, opened 2022-08-22
Add some more use of the private keyword in TabManager

darinwf:
Also moves test-only functions of TabManager into a dedicated extension and apply a `ForTesting` suffix.


jedireza (approved):
Looks reasonable. Build and tests are passing 👍

PR #4282, opened 2022-08-23
Probable fix for history DB being closed

darinwf:
See investigation in the issue, but it looks like we can end up with
activeSceneCount being more than 1 on an iPhone *and* have a closed
history DB at the point when starting that next scene. The end result
is a browser that cannot access the history DB. No search suggestions
on the zero-query and a History UI that doesn't load.

This should fix the issue. Adding more logging to try to better
understand what is going on.

Issue #3651


jedireza (approved):
Cool! Nice detective work. 🤞

Issue #4281, opened 2022-08-23
Crash @ [ PromoStateStorage.provision(url:) ]


darinwf:
Looks like it is not safe to write to Defaults from a background thread. The KVO system appears to be running code that is not thread-safe. Not sure if this is unique to our app and our use of the `Defaults` package or what. Apple docs say that `UserDefaults` is thread-safe, but clearly not whatever this observer happens to be doing.

```
#0	(null) in objc_release ()
#1	(null) in -[NSObject(NSKeyValueObservingPrivate) _changeValueForKeys:count:maybeOldValuesDict:maybeNewValuesDict:usingBlock:] ()
#2	(null) in -[NSObject(NSKeyValueObservingPrivate) _notifyObserversOfChangeFromValuesForKeys:toValuesForKeys:] ()
#3	(null) in -[CFPrefsSource forEachObserver:] ()
#4	(null) in -[CFPrefsSource _notifyObserversOfChangeFromValuesForKeys:toValuesForKeys:] ()
#5	(null) in ___CFPrefsDeliverPendingKVONotificationsGuts_block_invoke ()
#6	(null) in __CFDictionaryApplyFunction_block_invoke ()
#7	(null) in CFBasicHashApply ()
#8	(null) in CFDictionaryApplyFunction ()
#9	(null) in _CFPrefsDeliverPendingKVONotificationsGuts ()
#10	(null) in -[_CFXPreferences _deliverPendingKVONotifications] ()
#11	(null) in __108-[_CFXPreferences(SearchListAdditions) withSearchListForIdentifier:container:cloudConfigurationURL:perform:]_block_invoke ()
#12	(null) in normalizeQuintuplet ()
#13	(null) in -[_CFXPreferences withSearchListForIdentifier:container:cloudConfigurationURL:perform:] ()
#14	(null) in -[_CFXPreferences setValue:forKey:appIdentifier:container:configurationURL:] ()
#15	(null) in _CFPreferencesSetAppValueWithContainerAndConfiguration ()
#16	(null) in -[NSUserDefaults(NSUserDefaults) setObject:forKey:] ()
#17	0x00000001065e2cd4 in NSUserDefaults._set<A>(_:to:) at /Users/dfisher/Library/Developer/Xcode/DerivedData/Neeva-eqkexqatdwccugfburzfzimnqdmf/SourcePackages/checkouts/sindresorhus-Defaults/Sources/Defaults/UserDefaults.swift:66
[...]

PR #4284, opened 2022-08-23
Don't write to `Defaults` on background thread in ugc promos

EdwardJTL:
Comment out general hit stats because they are writing on background. We haven't been looking at those stats as much anyways.

Keeping "promoClear" log because we need it. Enforcing writing to that on main

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually tested on pages with promos

## Issues addressed
#4281


Client/Frontend/Overlay/Content/Cheatsheet/Promo/CheatsheetPromoModel.swift
```diff
@@ -38,28 +38,30 @@ private struct PromoStateStorage {
...
         guard let result = bloomFilterManager.contains(canonicalURL)
         else {
-            Defaults[.numOfUGCNoResult] += 1
+            //            Defaults[.numOfUGCNoResult] += 1
             return
         }
 
         cache[url] = result ? .hit : .missed
-        Defaults[.numOfUGCHits] += result ? 1 : 0
+        //        Defaults[.numOfUGCHits] += result ? 1 : 0
```
darinwf, line 60: nit: just tidy up the commenting-out of code.. maybe remove the extra spaces between the `//` and the commented out expression? also would be good to add a comment about why these are commented out. could be something like `// TODO(issue/xxxxxx): Not thread-safe; re-enable as part of session aggregated logging.`

darinwf (approved):

PR #4286, opened 2022-08-24
Update lastExecutedTime setUrl

evancrow:
Fixes a bug where backgrounding the app overnight would move the `selectedTab` to yesterday even after interacting with the tab (tapping a link, navigating, reloading the tab).

### How was this tested?
- [x] I tested this manually


Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -130,6 +130,12 @@ class Tab: NSObject, ObservableObject, GenericTab {
         }
 
         updateCanGoBackForward()
+
+        if browserViewController?.tabManager.selectedTab == self {
```
EdwardJTL, line 134: does this fire if you bring the app into foreground after midnight? they might just scroll on the tab for a bit and not change the URL?

evancrow: No, I'm looking into that separately.

darinwf: I think we should avoid having `Tab` know about `BrowserViewController` and even `TabManager`. Instead we can have `TabManager` set a boolean `isSelected` on `Tab` so that `Tab` can check that here.

darinwf: It is technical debt to remove current references to `BrowserViewController` and `TabManager` from `Tab` IMO.

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -266,7 +266,11 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
 
         assert(tab === selectedTab, "Expected tab is selected")
 
+        // Reset the isSelected state for every tab.
+        tabs.forEach { $0.isSelected = false }
```
darinwf, line 270: nit: use `previous` instead. will be faster. just compare to `selectedTab` and update `isSelected` accordingly.

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -485,7 +492,7 @@ class Tab: NSObject, ObservableObject, GenericTab {
             return
         }
 
-        if let bvc = browserViewController, bvc.tabManager.selectedTab == self {
+        if let bvc = browserViewController, isSelected {
```
darinwf, line 495: nice to see this member var helping with other bits!

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -415,7 +421,10 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
         else { return false }
 
         let parentTabIsMostRecentUsed = mostRecentTab(inTabs: viableTabs) == parentTab
-        if parentTabIsMostRecentUsed {
+
+        // If the parentTab is pinned we should return to it regardless of it being
+        // the most recently used to pass back the WebView and navigation stack.
+        if parentTabIsMostRecentUsed || parentTab.isPinned {
```
evancrow, line 427: Without this change, the user could have multiple child tabs (one more recent than the parent), and the user wouldn't be able to use the back button to return to the parent/pinned tab.

darinwf: OK, got it. But that's a different bug being fixed right? I think it would be better to make it into a separate PR with a test :-)

evancrow: It's caused by this PR, and the `testMultipleChildTabs` caught this bug, so we already have a  good test for this :)

darinwf: ah!!

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -266,7 +266,13 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
 
         assert(tab === selectedTab, "Expected tab is selected")
 
+        // Reset the isSelected state for every tab.
+        if previous != tab {
+            previous?.isSelected = false
+        }
+
         if let selectedTab = selectedTab {
+            selectedTab.isSelected = true
```
darinwf, line 275: How about moving this into the above branch?

```
if previous != selectedTab {
    previous?.isSelected = false
    selectedTab?.isSelected = true
}
```

Actually, come to think of it, we could just eliminate the check and just write this:

```
previous?.isSelected = false
selectedTab?.isSelected = true
```

So long as previous is assigned first, it should be fine even if selectedTab == previous.

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -267,6 +268,7 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
         assert(tab === selectedTab, "Expected tab is selected")
 
         if let selectedTab = selectedTab {
+            selectedTab.isSelected = true
```
darinwf, line 271: Just flagging, no need to change now, but the code between lines 251 and 271 includes code that calls out to other systems that could break if they ever looked for a `Tab` that has `isSelected` set to true.

darinwf (approved):

Client/Frontend/Browser/Tabs/TabManager.swift
evancrow: I was just thinking the same thing, moved variables around to prevent that!

darinwf, line 271: cool, then i think you don't need to set `isSelected = true` again here.

darinwf (approved):

PR #4287, opened 2022-08-24
Update lastExecutedTime When Scene Becomes Active

evancrow:
Works similar to #4286, just every time the user returns to the app and a tab is visible.

### How was this tested?
- [x] I tested this manually


Client/Application/SceneDelegate.swift
```diff
@@ -93,6 +93,11 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         if let interstitialViewModel = bvc.interstitialViewModel {
             interstitialViewModel.player?.play()
         }
+
+        // Update the selectedTab.lastExecutedTime if the tab is visible.
+        if !bvc.browserModel.showGrid {
```
darinwf, line 98: Maybe check `showContent` instead?

darinwf (approved):

darinwf (approved):

Issue #4288, opened 2022-08-24
Sign in option is not visible


EdwardJTL:
https://user-images.githubusercontent.com/44557675/186535113-8897793c-3526-44d3-893d-e987c5748be1.MOV

might be related to #4246

PR #4289, opened 2022-08-24
Fix Hidden Sign-In Button

evancrow:
Needed to account for vertical padding in the `minHeight`.

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #4288


EdwardJTL (approved):
tested on iPhone 13 simulator and iPad Air in stage manager. works well

PR #4297, opened 2022-08-25
add translation for week of 8/25

### LINKED ISSUE #4262: Onboarding Flow Premium Missing Translation
Running from main on Simulator.
Simulator Language and Region settings. Region is set to france and language is set to prefer french french
Simulator Screen Shot - iPhone 13 - 2022-08-19 at 11 01 24

Welcome Screen:
simulator_screenshot_302AF2B4-209F-4A40-961F-ACF01C42E20F

Issues:
- Subscription prices are in mixed currency
- not all bullet points are localized
- final price is not localized
- button is not localized

chung-lai:
Add new translations, fix #4262 

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Client/en-GB.lproj/Localizable.strings
```diff
@@ -1145,7 +1154,7 @@
 "Pin" = "Pin";
 
 /* No comment provided by engineer. */
-"Pin Tab" = "Pin Tab";
+"Pin Tab" = "Pin tab";
```
EdwardJTL, line 1157: should this be lower cased?

Client/fr.lproj/Localizable.strings
```diff
@@ -295,6 +295,9 @@
 /* No comment provided by engineer. */
 "arrows" = "flèches";
 
+/* No comment provided by engineer. */
+"Auto Archive Tabs" = "Archivage automatique des onglets";
```
EdwardJTL, line 299: I think this one has wrong tense. To be honest, I am not 100% sure what the english version is supposed to convey. I am guessing we are saying "The system should automatically archive tabs after a time period which a user can select", since this string is shown in the archived settings view. If my understanding of the purpose of the string is right, the french version is slightly off. I'm not sure about german.

EdwardJTL (commented):
some doubts...

Client/en-GB.lproj/Localizable.strings
chung-lai: good catch. Hmm..i wonder why they would do that, I don't see this happen in other cases. Will fix manually on the tool

Client/fr.lproj/Localizable.strings
chung-lai: True, "Auto Archive Tabs" don't seem right to start with. we may want to name it "Archive Tabs" or "Archive Tabs Automatically"

@Ria-yu, what do you think?

Ria-yu: Please bring this question to Design sync. Folks will want to have a say in this.

chung-lai: Discussed offline, we will

* change this to "Archive Tabs"
* change settings to add the value for the Archive Tabs setting
* add description at the footer to explain what the options mean

chung-lai: Will do this work in a follow up PR

EdwardJTL (approved):

Issue #4296, opened 2022-08-25
Tab Switcher picker can get stuck in an offset position


EdwardJTL:
NOTE: If this is a security or privacy related issue, please email
incidents@neeva.co instead of submitting an issue on GitHub.

Thank you for submitting an issue. Provide as much information as you can.

## Expected behavior

This is what I expected to happen.

## Actual behavior

This went wrong.

## Steps to reproduce the problem

1. 
1. 
1. 

## Screenshots or recordings

This is a big help if possible and relevant.

## Hardware and software

- Your device: 
- OS version: 
- App version: Neeva Browser x.x.x (x)
    - Found in app `Settings > About`

PR #4299, opened 2022-08-26
Fix Stuck DragOffset

evancrow:
Use `.updating` which is more reliable than `.onChanged`.

### How was this tested?
- [x] I tested this manually

(Dragging, dragging on tab, scrolling tab group, dragging `GridPicker`)

https://user-images.githubusercontent.com/37356984/186798954-b797305d-666a-48f3-b7c4-81234ae7d079.mp4

## Issues addressed
Fixed: #4296


EdwardJTL (approved):

Client/Frontend/Browser/CardGrid/Grid/GridPicker.swift
```diff
@@ -94,35 +94,39 @@ struct GridPicker: View {
...
+                    content.simultaneousGesture(gesture)
+                } else {
+                    content.highPriorityGesture(gesture)
+                }
+            } else {
+                content
+            }
+        }.onChange(of: dragOffset) { newValue in
+            if newValue == 0 {
+                switcherToolbarModel.dragOffset = nil
```
EdwardJTL, line 126: nit: to make the code more readable, we could make this value `private(set)` and have functions like `switcherToolbarModel.resetDragState()` etc for the view to use. This way the reader can understand why we need to set it to `nil` instead of `0`

PR #4301, opened 2022-08-26
Sync Simulated Backswipe Preview to Selection Target

EdwardJTL:
Save a preferred navigation target on simulated swipe model. 
Then pass this value to `tab.goBack` to select the correct tab.

this resolves the problem where if the parent tab is not the most recently used, even though the preview is shown, it will not be selected when swiping back.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
See videos
also checked that:
- closing a tab in tab grid still selects most recently used
- back swipe on page opened from zero query still opens zero query
- back swipe is disabled if the parent tab is deleted

| Before | After |
| ------ | ------ |
|<video src="https://user-images.githubusercontent.com/44557675/186966396-bbbd951d-04b6-49c2-9157-04730b186123.MOV"> | <video src="https://user-images.githubusercontent.com/44557675/186966320-6a58a41c-d772-446a-9b71-5eab82f3d314.MOV"> |


EdwardJTL:
when @darinwf, ria, and I talked about this behaviour in design sync, our design was that "treat closing from tab switcher and back swiping as different. Simulated swipe controller should always use the parent."
However, since then, on #4017, the discussion was that `Tab` should own the `canGoBack` state and `goBack` logic. I'm open to changing how this PR is implemented.

EdwardJTL:
but i also have this suggestion on what we can do to make it less messy #4302

evancrow:
Interesting! I'll do some testing and try out this PR. We should also keep in mind how the new pinned tab logic plays into this (since that relies heavily on returning to the correct tab on back nav).

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -546,6 +551,13 @@ class Tab: NSObject, ObservableObject, GenericTab {
                     bvc.zeroQueryModel.targetTab = .currentTab
                 }
             }
+        } else if case .parent = preferredTarget,
+            let parent = self.parent
+        {
+            browserViewController?.tabManager.removeTab(
```
evancrow, line 557: We should try to make this a function in `TabManager`: `removeTabAndSelectParent`. Or modify `removeTab` to have a `selectParent` argument.

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -546,6 +551,13 @@ class Tab: NSObject, ObservableObject, GenericTab {
...
+                self, showToast: false, updateSelectedTab: false
+            )
+            browserViewController?.tabManager.selectTab(parent, notify: true)
```
evancrow, line 560: This `else if case` should be merged with the `else if` below since they intend to do the same thing.

Client/Frontend/Browser/Tabs/Tab.swift
EdwardJTL: the current removeTab has an `updateSelectedTab` argument, which could select the parent tab if it's the most recently used one. i'm not sure what the updated signature should look like if we want to make the code clear and readable

Client/Frontend/Browser/Tabs/Tab.swift
EdwardJTL: the one below will only select the parent tab if it's also the most recently used. You can see the behaviour in tab grid. I kept that behaviour since the original intent was to just match the preview image to the tab.

EdwardJTL: added some comments

Client/Frontend/Browser/Tabs/Tab.swift
evancrow: Why don't we do something similar to `backNavigationTarget`? Create a SelectTabTarget enum:

```
enum SelectTabTarget {
    case parent
    case mostRecent
}

func removeTab(_ tab: Tab?, showToast: Bool = false, selectTabTarget: SelectTabTarget?)
```

_also have to update this for the other removeTab functions_

EdwardJTL: done :) also reorganize the code but the logic should still check out & tests cases are passing.

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -958,7 +945,14 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
     }
 
     // MARK: - CloseTabs
-    func removeTab(_ tab: Tab?, showToast: Bool = false, updateSelectedTab: Bool = true) {
+    /// - Parameters:
+    ///     - selectTabWith preference: preference for how the new tab should be picked. pass `nil` to not select a new tab
+    func removeTab(
+        _ tab: Tab?,
+        showToast: Bool = false,
+        updateSelectedTab: Bool = true,
```
evancrow, line 953: We can remove `updateSelectedTab`. Just rely on `selectTabWith`, if it's `none` don't select any tab. Can have a case `noPreference` or something like that.

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -1047,44 +1051,79 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
-        // select tab in today section.
+        // Check that there are viable tab targets
+
+        // Based on if the closed tab is incognito, create an ordered set of viable tabs
+        // In time-based switcher, the we only selected tabs in the `Today` section
         let viableTabs: [Tab] =
             tab.isIncognito
             ? incognitoTabs
             : normalTabs.filter {
                 $0.isIncluded(in: .today)
```
evancrow, line 1067: This should be updated to `$0.isIncluded(in: [.pinned, .today])`

EdwardJTL: That would allow it to select a pinned tab when the last unpinned tab is closed. Do we want that?

evancrow: Yes, we also need that for the updated pinned tab behavior.

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -958,7 +953,13 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
     }
 
     // MARK: - CloseTabs
-    func removeTab(_ tab: Tab?, showToast: Bool = false, updateSelectedTab: Bool = true) {
+    /// - Parameters:
+    ///     - selectTabWith preference: preference for how the new tab should be picked. pass `nil` to not select a new tab
```
evancrow, line 957: Nit: Update this to `selectTabPrefering:`, and really minor but capitalize Preference (writing like a sentence).

evancrow (approved):
Alright, this LGTM, with a minor nit!

PR #4303, opened 2022-08-26
Improve Readability in Tab Manager Creation Methods

EdwardJTL:
Adding two structs as inputs to `addTab` and `configureTab`

Personally, I find it challenging to track down all the parameters used to create tabs. It's hard to tell which parameters are actually needed in a function, and which ones are passed down. By adding these structs, it's easier to see how the values are being passed and where they are read.
It also clears up the signature a bit

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -645,116 +654,88 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
             tab,
-            atIndex: atIndex,
-            parent: parent,
-            keepInParentTabGroup: keepInParentTabGroup,
+            at: configuration.insertLocation,
             notify: notify
         )
 
-        if let webView = webView {
+        if let webView = configuration.webView {
```
EdwardJTL, line 698: I didn't see any invocation that passes a non-nil `webView` value. Do we still need this?

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -934,12 +924,14 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
 
-    @discardableResult func restore(savedTab: SavedTab, resolveParentRef: Bool = true) -> Tab {
+    @discardableResult
+    func restore(savedTab: SavedTab, resolveParentRef: Bool = true) -> Tab {
         // Provide an empty request to prevent a new tab from loading the home screen.
         // NOTE: tabIndex is ignored here as tabs are assumed to be saved in the order
         // they should be re-inserted.
         let tab = addTab(
-            flushToDisk: false, zombie: true, isIncognito: savedTab.isIncognito,
+            tabConfig: TabConfig(flushToDisk: false, zombie: true),
```
EdwardJTL, line 933: I think this is a common config too. perhaps we can come up with a name for it, then we can create another static variable that can be reused. (think `.ephemeral` config for `URLSession`)

Client/Frontend/Browser/Tabs/TabManager.swift
evancrow: Might be important for ZombieTabs 🤷‍♂️, though I think those have a unique `addTab` argument, right?

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -1682,13 +1689,43 @@ extension TabManager {
...
+struct InsertTabLocation {
+    var index: Int? = nil
+    weak var parent: Tab? = nil
+    var keepInParentTabGroup: Bool = true
+
+    static let `default`: Self = InsertTabLocation()
+}
+
+/// Parameters with which TabManager configures a tab after its creation
+struct TabConfig {
```
evancrow, line 1715: For readability, can we move these two new Structs to a separate file?

evancrow (commented):
I think this is a great idea!

Client/Frontend/Browser/Tabs/TabManager.swift
evancrow: There is the `zombie`  argument, which just prevents a `WebView` from being created but doesn't allow for a `WebView` to be passed if one already exists. I think we probably ought to leave `WebView` as optional in case we need it in the future.

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -645,116 +654,88 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
 
     private func configureTab(
-        _ tab: Tab, request: URLRequest?, webView: WKWebView? = nil, atIndex: Int? = nil,
-        afterTab parent: Tab? = nil, keepInParentTabGroup: Bool = true,
-        flushToDisk: Bool, zombie: Bool, isPopup: Bool = false,
-        query: String? = nil, suggestedQuery: String? = nil,
-        queryLocation: QueryForNavigation.Query.Location = .suggestion,
-        visitType: VisitType? = nil, notify: Bool
+        _ tab: Tab,
+        configuration: TabConfig = .default,
```
darinwf, line 683: nit: might make sense to stick with the same `tabConfig` parameter naming here? That way it is consistent throughout?

Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -785,10 +785,12 @@ class BrowserViewController: UIViewController, ModalPresenter {
         DispatchQueue.main.async {
             self.tabManager.selectTab(
                 self.tabManager.addTab(
-                    request,
-                    isIncognito: isIncognito,
-                    query: query,
-                    visitType: visitType
+                    tabConfig: TabConfig(
```
darinwf, line 788: nit: could consider using `.init(` instead of spelling out `TabConfig` given the name of the parameter is `tabConfig`

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -1682,13 +1689,43 @@ extension TabManager {
...
+
+struct InsertTabLocation {
+    var index: Int? = nil
+    weak var parent: Tab? = nil
+    var keepInParentTabGroup: Bool = true
+
+    static let `default`: Self = InsertTabLocation()
+}
+
+/// Parameters with which TabManager configures a tab after its creation
```
darinwf, line 1714: Since this new type is also referenced by `Tab`, I'd either define it at the top of `Tab.swift` or create a new file named `TabConfig.swift` for it. I like defining `InsertTabLocation` alongside whereever `TabConfig` is defined.

darinwf:
Love this change!

Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
EdwardJTL: changed everywhere :)

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -711,13 +711,14 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
         }
 
-        if let request = tabConfig.request,
-            let nav = tab.loadRequest(request),
-            let visitType = tabConfig.visitType
-        {
-            tab.browserViewController?.recordNavigationInTab(
-                navigation: nav, visitType: visitType
-            )
+        if let request = tabConfig.request {
```
darinwf, line 714: This commit says it is about improving readability, but isn't this a logic change? Is this logic change intended?

EdwardJTL: I made a logic change by accident when formatting the code in a previous commit, which also caused all the CI to fail. This commit puts the logic back; I compared the final changes to main and there's no logic change :P

darinwf: Oh, I should have looked at the net diff.

darinwf (approved):

PR #4304, opened 2022-08-26
Remove SiteArchiver

evancrow:
- Removes transition code from over a year ago
- Moves the rest of the code to `TabManagerStore` (the only place it's used)

### How was this tested?


Client/Frontend/Browser/Tabs/TabManagerStore.swift
```diff
@@ -234,15 +233,36 @@ class TabManagerStore {
             return [SavedTab]()
         }
     }
+
+    fileprivate func retrieveSavedTabsFromArchive(tabsStateArchivePath: String?) -> [SavedTab]? {
```
darinwf, line 237: naming nit: can take advantage of swift param names to write as:

```
retrieveSavedTabs(fromArchivePath archivePath: String?) -> [SavedTab]?
```

Read it like a sentence. No need to repeat the "from archive" bit :)

Within the function body use the name "archivePath" instead of "fromArchivePath" so that the code reads better too.

darinwf (approved):

darinwf (approved):

PR #4305, opened 2022-08-26
Allow selecting incognito tabs without date checking

EdwardJTL:
Since incognito tab switcher does not use date based sections, we won't need to enforce that when selecting a tab after the removal of the currently selected tab.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually tested tab selection code flow by changing system date time


Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -1074,7 +1074,7 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
         }
 
         if closedLastNormalTab || closedLastIncognitoTab
-            || !viableTabs.contains(where: { $0.isIncluded(in: .today) })
```
EdwardJTL, line 1077: this filter is already applied to `normalTabs` when `viableTabs` gets created

evancrow (approved):

PR #4308, opened 2022-08-29
Fix Another Pinned Tab Edge Case

evancrow:
Fixes in this PR:
- Prevents a new tab from being created on back navigation
- Fixes a bug where a pinned tab wouldn't be selected when removing a child tab
- Fixes the hackernews bug:
Caused by the current page index getting out of index, failing the assert in `SessionData.init`: `assert(currentPage > -urls.count && currentPage <= 0, "Session index is valid")`. This sets a cap on both the `navigationIndex` and `currentPage`, so it can't go past the last page in the navigation index.

### How was this tested?
- [x] I tested this manually
- [x] I added automated tests

### Manual test cases

**(hackernews case)**

- Pin a hackernews page
- Tap a comments link
- Go back
- Tap the same comments link
- Go back
- Should be back on the original hackernews tab

**(twitter case)**

- Pin twitter.com/home
- Tap a profile
- Go back
- Should be back on the original twitter.com/home tab


Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -782,7 +782,14 @@ class Tab: NSObject, ObservableObject, GenericTab {
...
                 urls.removeLast()
                 queries.removeLast()
-                navigationStackIndex -= 1
+
+                if navigationStackIndex > 1 {
+                    navigationStackIndex -= 1
+                }
+
+                if currentPage < 0 {
+                    currentPage += 1
```
darinwf, line 791: Should we just do `currentPage = 0` here instead?

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -782,7 +782,14 @@ class Tab: NSObject, ObservableObject, GenericTab {
             if urls.last != url, isForPinnedTabPlaceholder {
                 urls.removeLast()
                 queries.removeLast()
-                navigationStackIndex -= 1
+
+                if navigationStackIndex > 1 {
+                    navigationStackIndex -= 1
```
darinwf, line 787: Should this just be `navigationStackIndex = 1` instead?

evancrow:
Hmm, if the user has an extensive back/forward list, the Tab would always load on the last item then. Don't think that would work.

darinwf:
> Hmm, if the user has an extensive back/forward list, the Tab would always load on the last item then. Don't think that would work.

OK, I get it now. Thanks

darinwf (approved):

PR #4311, opened 2022-08-29
Continue ArchivedTabs Re-Design

evancrow:
In this PR:
Adds UI for both history and archived tabs.

Still left to do:
- [ ] Make rows tappable
- [ ] Make menu options functional
- [ ] Need context menus

### How was this tested?
- [x] I tested this manually

History:

ArchivedTabs:


evancrow:
Pretty big PR; going to mark it as ready for review, and I'll polish everything in a follow-up.

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
```diff
@@ -4,25 +4,141 @@
...
+        tabManager.archivedTabs.forEach {
+            if !handledRootIDs.contains($0.rootUUID) {
+                let lastExecutedTimeSeconds = $0.lastExecutedTime / 1000
+                groupedData.add(
+                    .tab($0),
+                    date: Date(timeIntervalSince1970: TimeInterval(lastExecutedTimeSeconds)))
+            }
+        }
+
+        self.objectWillChange.send()
```
darinwf, line 50: since you are signaling here, maybe `groupedData` should not be `@Published`?

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
```diff
@@ -4,25 +4,141 @@
...
+
+    let tabCardModel: TabCardModel
     let tabManager: TabManager
-    @Published var groupedData = DateGroupedTableData<ArchivedTab>()
-    @Published var filteredGroupedData = DateGroupedTableData<ArchivedTab>()
+    var numOfArchivedTabs: Int {
+        return tabManager.archivedTabs.count
+    }
+
+    func loadData() -> Deferred<Maybe<Cursor<T?>>> {
```
darinwf, line 22: I guess by using `Deferred<...>` here, you are setting it up so in the future we can complete `loadData` asynchronously? Given the code the way it is now, it is not clear why returning `Deferred` is helpful.

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
```diff
@@ -4,25 +4,141 @@
...
+        var singleTabRows: [[ArchivedTab]] = [[]]
+        var tabGroupRows: [[ArchivedTab]] = [[]]
+        var tabGroups: [String: ArchivedTabGroup] = [:]
+
+        data.forEach {
+            switch $0 {
+            case .tab(let tab):
+                addItemToRowOrCreateNewRowIfNeeded(rows: &singleTabRows, item: tab)
+            case .tabGroup(let tabGroup):
+                tabGroup.children.forEach { tab in
```
darinwf, line 89: This is a nice approach!

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
```diff
@@ -4,25 +4,141 @@
...
+                    addItemToRowOrCreateNewRowIfNeeded(rows: &tabGroupRows, item: tab)
+                }
+
+                tabGroupRows.append([])
+                tabGroups[tabGroup.id] = tabGroup
+            }
+        }
+
+        let rows = singleTabRows + tabGroupRows
```
darinwf, line 98: doesn't this mess up the order? it looks like this puts all single tab rows ahead of all groups? but shouldn't they be intermixed?

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
```diff
@@ -4,25 +4,141 @@
...
+
+        return AnyView(
```
darinwf, line 100: I'm a little surprised to see buildRows returning the `LazyVStack`. It might be helpful to separate these out. The CardGrid does so in case `body` needs to get called a bunch of times. That way the bulk of the `buildRows` complexity does not need to get recomputed. In other words, you could make the output of `buildRows` be a vector of `rows` that is the state of the model. Then the view can just render the computed rows. If it needs to do so repeatedly, it is okay since it will be cheap.

maybe for the archived tabs view we don't need to worry about the cost of such recomputation.

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
```diff
@@ -4,25 +4,141 @@
...
         // TODO: (Evan) Load data from TabManager filtering by title/URL.
         // No current implementation but similar to FindInGrid.
+
+        return .init()
+    }
+
+    func buildRows(
+        with data: [ArchivedTabRow],
+        for section: DateGroupedTableDataSection
+    ) -> AnyView {
```
EdwardJTL, line 65: i was able to get this to build by
1. changing this to `some View`, and then
2. in `GroupedDataPanelModel` protocol, add an associated type `associatedtype RowView: View`, and have the function `buildRows` return a `RowView`.

hopefully that works for you

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
evancrow: I moved the `LazyVStack` out to the shared view (`GroupedDataPanelView`). The reason I have the view model draw the views instead of just the data layout is that it's very different for History than for ArchivedTabs, but I still wanted to have a generic view and model that we could reuse.

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
evancrow: I did it the other way around, removed the `objectWillChange` and kept `@Published`

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
evancrow: I made the generic return parameter of `loadData()` `Deferred`, so that History can continue returning async, and ArchivedTabs doesn't have to but can in the future. Just keeps the code simpler.

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
evancrow: Let me modify the approach to keep them intermixed

Client/Frontend/Browser/History+ArchivedTabs/GroupedDataPanelModel.swift
```diff
@@ -4,14 +4,18 @@
...
     associatedtype T: Equatable
+    associatedtype Rows
 
     var tabManager: TabManager { get }
     var groupedData: DateGroupedTableData<T> { get set }
     var filteredGroupedData: DateGroupedTableData<T> { get set }
 
-    func loadData()
-    func loadData(filter query: String)
+    func loadData() -> Deferred<Maybe<Cursor<T?>>>
```
EdwardJTL, line 18: could we have a comment on why we need these return types?

Client/Frontend/Browser/History+ArchivedTabs/GroupedDataPanelView.swift
```diff
@@ -2,45 +2,132 @@
...
+            }
+        }
+        .animation(.interactiveSpring())
+        .transition(.identity)
+    }
 
-            } label: {
-                HStack {
-                    Text("Clear Browsing Data")
+    private func buildDaySections(section: DateGroupedTableDataSection) -> some View {
```
EdwardJTL, line 35: add a viewbuilder decorator

Client/Frontend/Browser/History+ArchivedTabs/GroupedDataPanelView.swift
```diff
@@ -2,45 +2,132 @@
...
-
-            Color.groupedBackground.frame(height: 8)
+            }
         }
     }
 
-    var body: some View {
-        VStack {
-            optionSections
+    private func buildSection(
```
EdwardJTL, line 58: viewbuilder

Client/Frontend/Browser/History+ArchivedTabs/History/HistoryGroupedDataModel.swift
```diff
@@ -5,25 +5,137 @@
...
+
+    private func addSiteDataToGroupedSites(_ result: Maybe<Cursor<Site?>>) {
+        if let sites = result.successValue {
+            for site in sites {
+                guard let site = site as? Site, let latestVisit = site.latestVisit else {
+                    return
+                }
+
+                let date = Date(
+                    timeIntervalSince1970: TimeInterval.fromMicrosecondTimestamp(latestVisit.date))
```
EdwardJTL, line 48: could we add a millisecond util for the `/1000` calculations we have in other parts of the PR?

Client/Frontend/Browser/History+ArchivedTabs/History/HistoryGroupedDataModel.swift
```diff
@@ -5,25 +5,137 @@
...
+        return profile.history.getSitesByLastVisit(
+            limit: queryFetchLimit, offset: currentFetchOffset) >>== { result in
+                // Force 100ms delay between resolution of the last batch of results
+                // and the next time `fetchData()` can be called.
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                    self.currentFetchOffset += self.queryFetchLimit
+                    self.isFetchInProgress = false
+                }
+
+                return deferMaybe(result)
```
EdwardJTL, line 114: so this may be called before the async block gets executed?

Client/Frontend/Browser/History+ArchivedTabs/HistoryAndArchivedTabsPanelView.swift
```diff
@@ -39,48 +54,94 @@ struct HistoryAndArchivedTabsPanelView: View {
...
+                        Text("Recently Closed Tabs")
+                        Spacer()
+                        Symbol(decorative: .chevronRight)
+                    }
+                }
+                .foregroundColor(.label)
+                .padding(16)
+
+                (FeatureFlag[.archivedTabsRedesign]
+                    ? Color.DefaultBackground : Color.groupedBackground).frame(height: 1)
```
EdwardJTL, line 73: could we have the frame on a new line?

Client/Frontend/Browser/History+ArchivedTabs/HistoryAndArchivedTabsPanelView.swift
```diff
@@ -39,48 +54,94 @@ struct HistoryAndArchivedTabsPanelView: View {
...
+                            Text(archivedTabsAfterLabel)
+                                .foregroundColor(.secondaryLabel)
+                        }
+                    }
+                }
+                .foregroundColor(.label)
+                .padding(16)
+
+                (FeatureFlag[.archivedTabsRedesign]
+                    ? Color.DefaultBackground : Color.groupedBackground).frame(height: 1)
```
EdwardJTL, line 116: same here

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsGroupedDataModel.swift
```diff
@@ -4,25 +4,104 @@
...
+    }
+
+    func buildRows(
+        with data: [ArchivedTabRow],
+        for section: DateGroupedTableDataSection
+    ) -> Rows {
+        let tabsPerRow = tabCardModel.columnCount
+
+        // TODO: (Evan) Rename ArchivedTabRow since it's very confusing now.
+        func addItemToRowOrCreateNewRowIfNeeded(rows: inout [[ArchivedTab]], item: ArchivedTab) {
```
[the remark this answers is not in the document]
evancrow: I initially set this up with `stride` because that is what `TabDetailModel.buildRows` uses. But because each `ArchivedTabRow` has to be unwrapped to an `ArchivedTab,` and since TabGroups need to start on a new line, this logic I set up was simpler.

Client/Frontend/Browser/History+ArchivedTabs/History/HistoryGroupedDataModel.swift
evancrow: Yes, this code was originally inherited from Firefox, it doesn't mean it's necessarily correct, but it's worked for as long as it's been in our app. 

I think it's okay if it is executed before anyways; find to have a bit of a buffer, so we're not refreshing data right after we just refreshed it.

EdwardJTL (approved):

PR #4316, opened 2022-08-30
Sync Canonical V3 to Backend

EdwardJTL:
matching https://github.com/neevaco/neeva/commit/8d3bff203e3e441bacda6c9fc0b5724f417a9421

### How was this tested?
- [ ] I tested this manually
- [x] I added automated tests


darinwf (commented):
Do we need a new URL for the newly generated bloomfilter as well? I assume this represents a protocol change and that the existing bloomfilter would no longer be the correct one to use? Or am I misunderstanding the impact of this change?

EdwardJTL:
> Do we need a new URL for the newly generated bloomfilter as well? I assume this represents a protocol change and that the existing bloomfilter would no longer be the correct one to use? Or am I misunderstanding the impact of this change?

Checked with Yifan, the current binary we are serving is built with this changed canonical URL. So this PR actually brings the client up to using the same correct canonical function.

Yifan and I talked about better versioning, and he's investigating some ideas. I think this canonical change is small, and we can add the versioning later.

darinwf (approved):

PR #4317, opened 2022-08-30
Use Neeva Constants for Preview URL

EdwardJTL:
Forgot to use the app host url.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Manual test cases
Observed traffic


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
```diff
@@ -334,7 +334,11 @@ private class PreviewCookieFetcher {
             self.session.configuration.httpCookieStorage?.removeCookies(since: .distantPast)
 
             // Pin endpoint to get new cookie synchronously
-            let url = URL(string: "https://neeva.com/preview?pid=neeva_scope")!
+            // Rather this code crashes the app than make the app hang
+            precondition(!Thread.isMainThread)
+            let url = DispatchQueue.main.sync {
```
darinwf, line 339: Can this be done ahead of time, earlier on, when code is still on the main thread? For example, if `fetch` is only ever called on the main thread, then we could capture this value then and use it from the background thread. This way we avoid the need to sync back to the main thread like this.

It looks like `fetch` can only be called on the main thread. I see that `fetchRichResult` ensures that is the case at least. Why not simplify and just require that `fetch` is only called on the main thread and then compute `buildAppURL` upfront.

EdwardJTL: yup makes sense from the caller's perspective to get the url before we queue up the operation.

also i realized that the result handler still needs a check to requeue on main thread if there was an invalid preview cookie in the keychain

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
```diff
@@ -334,7 +351,7 @@ private class PreviewCookieFetcher {
             self.session.configuration.httpCookieStorage?.removeCookies(since: .distantPast)
 
             // Pin endpoint to get new cookie synchronously
-            let url = URL(string: "https://neeva.com/preview?pid=neeva_scope")!
+            // Rather this code crashes the app than make the app hang
```
darinwf, line 354: How might this next line of code result in the app crashing?

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
```diff
@@ -165,14 +165,27 @@ final class CheatsheetServiceProvider: CheatsheetDataService {
...
-                self.fetchPreviewResults(
-                    query: query,
-                    // Ping "preview" to get new cookie on retry
-                    forceFetchCookie: true,
-                    // Max 1 retry
-                    retryOnAuthError: false,
-                    completion: completion
-                )
+                if Thread.isMainThread {
+                    self.fetchPreviewResults(
```
darinwf, line 169: Avoid duplicating code, maybe like this?

```
let callFetchPreviewResults = {
    self.fetchPreviewResults(
        query: query,
        // Ping "preview" to get new cookie on retry
        forceFetchCookie: true,
        // Max 1 retry
        retryOnAuthError: false,
        completion: completion
    )
}

if Thread.isMainThread {
    callFetchPreviewResults()
} else {
    DispatchQueue.main.async {
        callFetchPreviewResults()
    }
}
```

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
EdwardJTL: we build release with `-O` config, so `precondition` will stop execution if it evaluates to false

darinwf (approved):

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
darinwf: Sorry, where is the `precondition`?

EdwardJTL: ah sorry i should've removed the comment when i moved building the URL out of this scope. the precondition is on line 340 now.

PR #4319, opened 2022-08-30
Allow Multiple Domains For International Sites Safari Extension

evancrow:
### How was this tested?
- [x] I tested this manually


NeevaForSafari/NeevaForSafariExtension/Resources/content.js
```diff
@@ -64,3 +68,17 @@ function getParameterByName(name, url = window.location.href) {
...
+function trimURLHost(host) {
+    if (host == null) {
+        return host;
+    }
+
+    // Remove www if it was added to the host.
+    host = host.replace("www.", "");
+
+    // Remove .com/.org/etc from the host.
+    host = host.substring(0, host.lastIndexOf("."));
```
chung-lai, line 81: This covers some cases but not all? (e.g., https://hk.search.yahoo.com/, then this will return `hk.search.yahoo` instead of `search.yahoo`, or https://search.com.de and it may return `search.com`). It may work for certain search engine like google, in that case, may want to call this function only for google domains?

evancrow: Ah true. I don't want to hardcode all these subdomains but I'll try and see if there's already a list online.

evancrow:
@chung-lai I couldn't find a list, but I've made it easy to support new subdomains. So I'll try and find some and let me know if you come across any more.

NeevaForSafari/NeevaForSafariExtension/Resources/content.js
```diff
@@ -1,47 +1,88 @@
...
```
chung-lai, line 35: This may not work with domain `www.google.com.uy`? I think this will return `google.com` in this case and it won't be included in the `SearchEngineHosts`

chung-lai: For now, may be we can use the list from `NavigationRouter+MaybeRewriteURL` for google domain?

evancrow: I can't use that function since this is a JS file and technically a separate app, but I should be able to copy the logic. Should also double check `NavigationRouter+MaybeRewriteURL` because that may have the same problem here.

chung-lai: Right, i think we may start just copying the google list from that file for now. So far, i found only Google and Yahoo seem to have different domain in different countries. Let me try to see if there is a yahoo list.

evancrow: K copied the domain list from that file. I forgot that was there! Think setting up that code was one of my first bug fixes :)

NeevaForSafari/NeevaForSafariExtension/Resources/content.js
```diff
@@ -1,47 +1,285 @@
...
+
+//
+// Navigiation handling methods.
 browser.runtime.sendMessage({ "getPreference": "neevaRedirect"}).then((response) => {
     const searchQuery = navigateIfNeeded(window.location);
-    const possibleReferalURLs = ["https://www.google.com/", "https://www.bing.com/", "https://www.ecosia.com/",
-                                  "https://search.yahoo.com/", "https://duckduckgo.com/", "https://yandex.com/",
-                                  "https://www.baidu.com/", "https://www.so.com/", "https://www.sogou.com/", "https://neeva.com/"];
     let referrerURL = window.document.referrer;
+    let referrerHost = trimURLHost(referrerURL.replace("https://", ""));
```
chung-lai, line 242: Not super familiar with javascript. Instead of `trimURLHost`, can we convert the referrer to [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) object and get the hostname that way? Feels like it should be more robust and future proof by relying on library to do the job

chung-lai (approved):
Left one comment; otherwise, LGTM! Thanks Evan

NeevaForSafari/NeevaForSafariExtension/Resources/content.js
evancrow: URLs do have the `.host` option (see `navigateIfNeeded`), but `referrerURL` is actually a `String` so we have to manually get it. All `trimURLHost` does is remove `www.`

PR #4320, opened 2022-08-31
auth; finish auth handshake without opening new tab

jedireza:
This PR enables us to realize a user is authenticated without needing to open a new tab.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested on my physical device.

N/A

## Issues addressed
N/A


Client/Frontend/Authentication/AuthStore.swift
```diff
@@ -425,6 +446,8 @@ extension AuthStore {
                     let signInURL = URL(
                         string: "https://\(NeevaConstants.appHost)/login/qr/finish?q=\(token)")!
 
+                    // TODO: finish handshake without opening new tab
```
jedireza, line 449: This will be part of this PR. Will commit before EOD.

chung-lai:
So instead of opening the neeva url in a webview, we request the auth URL (neeva url) using URLSession and extract the cookie from the response, is that the idea?

Client/Frontend/Authentication/AuthStore.swift
```diff
@@ -21,20 +21,17 @@ class AuthStore: NSObject, ObservableObject {
     }
 
     private func setLoginToken(token: String) {
-        NeevaUserInfo.shared.setLoginCookie(token)
+        DispatchQueue.main.async {
```
darinwf, line 24: Why is this needed?

A concern I have when seeing this is that it can result in a small delay in calling `NeevaUserInfo.shared.setLoginCookie`, which means that if someone tries to read the login token immediately after calling `setLoginToken` then they will get the old value. This could result in very subtle bugs.

jedireza:
> So instead of opening the neeva url in a webview, we request the auth URL (neeva url) using URLSession and extract the cookie from the response, is that the idea?

Yes

Client/Frontend/Authentication/AuthStore.swift
jedireza: I wish I had a better intuition for using this. I get code smell vibes from it. I only added this after the app kept crashing and the stack trace showed the crash was at a guard/precondition that require the call to happen on the main thread.

jedireza: I'll double check this one later today. I'll remove the call and see if I still have a crash.

darinwf: Cool, would be happy to review the crash stack. Would be interesting to see what is calling this function from a background thread. If anything, maybe the `DispatchQueue.main.async` should happen in that code. This function could then have a `precondition(Thread.isMainThread)`

jedireza: So `NeevaUserInfo.setLoginCookie` internally calls `NeevaConstants.removePreviewCookie` which itself has a precondition `precondition(Thread.isMainThread)`.

darinwf (approved):

PR #4329, opened 2022-09-01
Update selected tab when restoring tabs with tab content visible

EdwardJTL:
Main UI change:
- when a tab is selected and in focus (i.e. visible), restoring a tab -- via toast or via keyboard shortcut -- should select the restored tab(s).
- when a tab is selected from recently closed tabs in history, it should restore and mark the tab as selected so user can find it in the tab grid.

Other code clean ups:
- Recently closed tabs no longer filter by internal URL
- Recently closed tabs uses did set instead of computed filters.
- select tab logic clean ups. (fewer loops)

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


EdwardJTL:
awaiting design review

evancrow (approved):
Code changes LGTM

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -84,12 +84,12 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
-    var recentlyClosedTabs = [[SavedTab]]()
-    var recentlyClosedTabsFlattened: [SavedTab] {
-        Array(recentlyClosedTabs.joined()).filter {
-            !InternalURL.isValid(url: ($0.url ?? URL(string: "")))
+    var recentlyClosedTabs = [[SavedTab]]() {
+        didSet {
+            recentlyClosedTabsFlattened = recentlyClosedTabs.flatMap { $0 }
         }
     }
+    var recentlyClosedTabsFlattened: [SavedTab] = []
```
darinwf, line 92: nit: should this be marked `private(set)`?

darinwf (approved):

PR #4331, opened 2022-09-02
Refactor BuildRows

evancrow:
This is basically done, just need to wrap up some TabGroup stuff and make sure iPad works well.

### How was this tested?
- [x] I tested this manually


Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -35,17 +35,11 @@ struct TabGridContainer: View {
     @State var cardStackGeom: CGSize = CGSize.zero
 
     var selectedRowId: TabCardModel.Row.ID? {
-        let rows = tabModel.getRows(incognito: isIncognito)
+        let rows = tabModel.getRows(for: .all, incognito: isIncognito)
         if let row = rows.first(where: { row in
             row.cells.contains(where: \.isSelected)
         }) {
-            if row.index == 2 {
```
darinwf, line 42: This was to help ensure that the scroll offset will be zero when scrolling to cards in the first row. Are you sure this isn't breaking that?

Client/Frontend/Browser/CardGrid/Grid/TabGridCardView.swift
```diff
@@ -0,0 +1,26 @@
...
+    @EnvironmentObject var tabModel: TabCardModel
+
+    let containerGeometry: GeometryProxy
+
+    var body: some View {
+        ForEach(
+            tabModel.getRowSectionsNeeded(incognito: incognitoModel.isIncognito),
+            id: \.self
+        ) { section in
+            TabGridSectionView(
```
darinwf, line 18: I don't know enough about how `LazyVStack` works. Does this mean that whole tab grid sections are lazily loaded or does it still incrementally bring in individual rows? If the former, then this will be bad for performance and doesn't work.

If you instead let the section headers be rows (as in the current implementation), then you can avoid so much nesting, which should ensure that SwiftUI is able to do what we want.

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
evancrow: I'm going to improve this implementation if I can. Currently testing out my new design to see if it's necessary.

Client/Frontend/Browser/CardGrid/Grid/TabGridCardView.swift
evancrow: I had the same thoughts too. Still testing but so far it's still lazy loading individual rows. I took this implementation from the new `ArchivedTabsSectionView` and that lazy loads as well.

darinwf: I see. You can probably eliminate any uncertainty here with some well placed `print` statements :)

evancrow:
This is good to go now, screenshots of it working:

iPhone:

https://user-images.githubusercontent.com/37356984/188290667-1606facf-715c-463c-99ed-482617f281cc.mp4



iPad:
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-09-03 at 14 02 41
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-09-03 at 14 05 32
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-09-03 at 14 02 31
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-09-03 at 13 59 14

Client/Frontend/Browser/CardGrid/Grid/TabGroupHeaderView.swift
```diff
@@ -0,0 +1,109 @@
...
+        }
+        .actionSheet(isPresented: $groupDetails.deleting) {
+            let buttons: [ActionSheet.Button] = [
+                .destructive(Text("Close All")) {
+                    groupDetails.onClose(showToast: false)
+                },
+                .cancel(),
+            ]
+
+            if let title = groupDetails.customTitle {
```
EdwardJTL, line 98: nit: could we refactor this into just a compute for the title string, and then have a single return statement?

Client/Frontend/Browser/CardGrid/Grid/TabGroupHeaderView.swift
```diff
@@ -0,0 +1,109 @@
...
+                .withFont(.labelLarge)
+                .foregroundColor(.label)
+                .accessibility(identifier: "TabGroupTitle")
+                .accessibility(value: Text(groupDetails.title))
+
+            Spacer()
+
+            if groupDetails.allDetails.count > columns.count {
+                Button {
+                    if groupDetails.isExpanded {
```
EdwardJTL, line 33: nit: could we refactor this into a helper function so that the view is easier to read?

Client/Frontend/Browser/CardGrid/Grid/TabGridSectionView.swift
```diff
@@ -13,42 +13,70 @@ enum SingleLevelTabCardsViewUX {
...
+
+struct TabGridSectionView: View {
     @EnvironmentObject var tabModel: TabCardModel
     @EnvironmentObject private var browserModel: BrowserModel
 
     let containerGeometry: GeometryProxy
+    let section: TabSection
     let incognito: Bool
 
+    var rows: [Row] {
```
EdwardJTL, line 46: could this be a let, and replace the `section` param as the initializer param?

EdwardJTL: if you do that, this view also doesn't need to observe to `tabModel: TabCardModel` updates which could save us view graph computes

Client/Frontend/Browser/CardGrid/Grid/TabGridSectionView.swift
```diff
@@ -36,16 +36,12 @@ struct TabGridSectionHeaderView: View {
 }
 
 struct TabGridSectionView: View {
-    @EnvironmentObject var tabModel: TabCardModel
     @EnvironmentObject private var browserModel: BrowserModel
 
+    let tabModel: TabCardModel
```
EdwardJTL, line 41: do we need to keep this reference?

evancrow: Nope!

EdwardJTL (approved):

PR #4334, opened 2022-09-02
adding easy list in other languages

chung-lai:
* Add german and french easy list for ad blocker
* Load language specific ad block rules based on app store country

Will add unit test for verifying the list

Tested with some french/german sites and it blocks some (but not all) ads.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


content-blocker-lib-ios/src/ContentBlocker.swift
```diff
@@ -24,12 +25,30 @@ enum BlocklistCategory: CaseIterable {
...
+        case .easyPrivacy:
+            return "easy_privacy"
+        case .easyPrivacyStrict:
+            return "easy_privacy_strict"
+        case .easyListAdBlock:
+            if let countryCode = SKPaymentQueue.default().storefront?.countryCode,
+                countryCode == "DEU" || countryCode == "FRA"
+            {
+                if countryCode == "DEU" {
+                    return "easy_list_german"
```
darinwf, line 44: I see, so these are concatenations, e.g. easylist + easylist-germany? Perhaps we should have a document explaining how to re-generate these files?

chung-lai: The naming of the file is up to us, sounds good with concatenation :)

Right, will write up a quick doc on this and will start thinking how to keep these up to date for each app release as they update very frequently. (https://github.com/easylist/easylist) Also at some point in the future as we support more and more rules for other countries, we should host these files and have the app download on launch.

darinwf: Oh because the rules files change that quickly?

chung-lai: yes. When you open the easylist.txt, you will see a date on top and it seems to update daily

chung-lai:
Haven't been able to test the `storefront` part with simulator/debug device. Will test on device with testflight build.

darinwf (approved):

PR #4337, opened 2022-09-07
CMD+L Edit Current Page

evancrow:
In line with other browsers, `CMD+L` should edit the current page instead of using `createOrSwitchToTab`.

### How was this tested?
- [x] I tested this manually


EdwardJTL (approved):
tested and works :) ty!

darinwf:
Good change!

Issue #4332, opened 2022-09-02
Conflict in keyboard command when restore tab toast is showing


EdwardJTL:
I am not sure if this problem is limited to iPadOS 16.

Steps:
- On iPad with physical keyboard, close two tabs from the tab strip. Then, wait for the toast to show. Press cmd+shift+T. 

Expect:
- As shown in the toast label, restore 2 tabs.

Actual behaviour:
- Restore only the last closed tab. I think the always active BVC keyboard commands are prioritized?

evancrow:
I had fixed this before, must have been broken again. Will take a look.

PR #4339, opened 2022-09-08
Group Tabs With AddToRecentlyClosed

evancrow:
Uses the same method for `closedTabsToShowToastFor` and applies it to `recentlyClosedTabs`.

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #4332


EdwardJTL:
hmm i'm not sure this is the right fix. the keyboard interaction priority seems inconsistent. occasionally on `main`, i can press cmd+shift+T when the toast shows up, and it will restore multiple tabs. So this fix matches the two shortcuts' behaviours, but it doesn't resolve the cause of the problem

evancrow:
Probably should just disable the BVC command while the restore tab toast is visible.

EdwardJTL (approved):
looks good. tested restoring both with and without toast. thanks!

Issue #4342, opened 2022-09-08
Excessive calls to FaviconView.body


darinwf:
I'm observing what feels like an excessive number of calls to `FaviconView.body`, which is not exactly cheap since it calls `WebImage` load an image. That image should be fetched from a local cache, but it is still not free.

Just with a single tab open, I see about 20 calls to FaviconView.body when showing or hiding the card grid. That should be far less.

darinwf:
It looks like `Card.body` is getting invoked even more times during a card transition animation. Hmm... that could point to other performance opportunities.

darinwf:
Here's a trace from showing the card grid w/ a single tab:

```
>>> Card.body: 108
>>> Card.body: 109
>>> CardTransitionModifier.body(): 61
>>> FaviconView.body: 78
>>> Card.body: 110
>>> Card.body: 111
>>> CardTransitionModifier.body(): 62
>>> FaviconView.body: 79
>>> Card.body: 112
>>> Card.body: 113
>>> CardTransitionModifier.body(): 63
>>> Card.body: 114
>>> FaviconView.body: 80
>>> FaviconView.body: 81
>>> Card.body: 115
>>> Card.body: 116
>>> CardTransitionModifier.body(): 64
>>> Card.body: 117
>>> FaviconView.body: 82
>>> FaviconView.body: 83
>>> Card.body: 118
>>> Card.body: 119
>>> CardTransitionModifier.body(): 65
>>> Card.body: 120
>>> FaviconView.body: 84
>>> FaviconView.body: 85
>>> Card.body: 121
>>> CardTransitionModifier.body(): 66
>>> Card.body: 122
>>> FaviconView.body: 86
>>> FaviconView.body: 87
>>> Card.body: 123
>>> Card.body: 124
>>> CardTransitionModifier.body(): 67
>>> Card.body: 125
>>> FaviconView.body: 88
>>> FaviconView.body: 89
>>> Card.body: 126
>>> Card.body: 127
>>> CardTransitionModifier.body(): 68
>>> Card.body: 128
>>> FaviconView.body: 90
>>> FaviconView.body: 91
>>> Card.body: 129
>>> Card.body: 130
>>> CardTransitionModifier.body(): 69
>>> Card.body: 131
>>> FaviconView.body: 92
```

darinwf:
Adding more to the trace:

```
>>> Card.body: 61
>>> SingleLevelTabCardsView.body: 83
>>> Card.body: 62
[...]

PR #4347, opened 2022-09-09
Reduce various expensive updates that happen while showing the card grid

darinwf:
Summary of changes:

- Avoid calling `TabCardModel.updateRows()` when taking a new screenshot of the tab, which happens as we go to show the card grid. We avoid this by having `TabCardDetails` monitor the published `Tab.screenshotUUID` instead. Because `Card` observes the `CardDetails`, this results in the `Card` picking up the new screenshot.

- Avoid calling `updateAllTabDataAndSendNotifications` when just selecting the tab that is already selected. Showing the card grid and then showing the same tab again should not need to re-generate all of the tabs data. This is done by adding a `needsHeavyUpdatesPostAnimation` flag to `TabManager` that gets set when `selectTab` would otherwise do this heavy work.

- Avoid watching all of `GridModel` in various layers of the code. A number of properties of `GridModel` get poked during the transition to the card grid, and by breaking these out into their own model classes, we can avoid updating so much UI when making changes to a subset of those properties. This translated to creating `GridResizeModel`, `GridScrollModel` and `GridSwitcherModel`.

- Switch to just observing some particular published vars on some models (e.g., `GridScrollModel`) instead of the entire model to avoid spurious UI updates as well. Add a `useEffect` variant to make this easier.

- Move observation of `TabChromeModel` out of `CardScrollContainer` and into `EmptyCardGrid` where it is needed, which helps cut down on invalidations of `CardScrollContainer`.

The net result of this is that for a card grid with a single card, we reduce the number of `FaviconView` updates by about 30%.
### How was this tested?
- [x] I tested this manually

## Issues addressed
* Issue #4342


darinwf:
@evancrow @EdwardJTL - would be especially great to get your eyes on this change, thanks!

Client/Frontend/Browser/BrowserView.swift
```diff
@@ -154,6 +154,9 @@ struct BrowserView: View {
         .environmentObject(browserModel.toastViewManager)
         .environmentObject(bvc.gridModel)
         .environmentObject(bvc.gridModel.spaceCardModel)
+        .environmentObject(bvc.gridModel.resizeModel)
```
evancrow, line 157: nit: sort this alphabetically (move `spaceCardModel`)

Client/Frontend/Browser/CardGrid/Grid/CardGrid.swift
```diff
@@ -14,6 +14,36 @@ enum CardGridUX {
...
+                .environment(\.shareURL, shareURL)
+                .environment(\.columns, columns)
+        }
+    }
+
+    var body: some View {
+        NavigationLink(
+            destination: detailedSpaceView,
+            isActive: $gridModel.showingDetailView
+        ) {}.useEffect(deps: spaceModel.detailedSpace) { detailedSpace in
```
evancrow, line 41: Usually, when we have an empty `NavigationLink`, we've been inserting `EmptyView()` into the content closure.

Client/Frontend/Browser/CardGrid/Grid/CardGrid.swift
```diff
@@ -14,6 +14,36 @@ enum CardGridUX {
...
+    @Environment(\.onOpenURLForSpace) var onOpenURLForSpace
+    @Environment(\.shareURL) var shareURL
+
+    let columns: [GridItem]
+
+    @ViewBuilder
+    var detailedSpaceView: some View {
+        if let detailedSpace = spaceModel.detailedSpace {
+            SpaceContainerView(primitive: detailedSpace)
+                .environment(\.onOpenURLForSpace, onOpenURLForSpace)
```
evancrow, line 31: Think we can remove these; since we have a full SwiftUI stack, should be passed down automatically.

Client/Frontend/Browser/CardGrid/Grid/CardGrid.swift
```diff
@@ -14,6 +14,36 @@ enum CardGridUX {
     static let GridSpacing: CGFloat = 20
 }
 
+// Isolating dependency on GridModel and SpaceCardModel.
+struct SpaceNavigationLink: View {
+    @EnvironmentObject var gridModel: GridModel
+    @EnvironmentObject var spaceModel: SpaceCardModel
+
+    @Environment(\.onOpenURLForSpace) var onOpenURLForSpace
```
evancrow, line 22: And then these

Client/Frontend/Browser/CardGrid/Grid/GridModel.swift
```diff
@@ -5,26 +5,79 @@
...
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+            self.gridCanAnimate = false
         }
     }
-    @Published var gridCanAnimate = false
-    @Published var switchModeWithoutAnimation = false
+
+    @Published private(set) var switchModeWithoutAnimation = false
+    func update(switchModeWithoutAnimation: Bool) {
```
evancrow, line 62: Since this is usually always called before `update(state: SwitcherView)`, does it make sense to combine it into a single function?

Client/Frontend/Browser/CardGrid/Grid/TabGridCardView.swift
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import Shared
```
evancrow, line 5: nit: unnecessary?

Client/Frontend/Favicons/FaviconView.swift
```diff
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import SDWebImageSwiftUI
+import Shared
```
evancrow, line 6: nit: unnecessary?

Client/Frontend/Browser/CardGrid/Grid/CardGrid.swift
darinwf: Somehow this is necessary. Without it, URLs from the space list open in the default browser (Safari). I don't understand why yet.

Client/Frontend/Browser/CardGrid/Grid/GridModel.swift
darinwf: I don't think so. There are cases where the two functions are called independently.

Client/Frontend/Browser/CardGrid/Grid/CardGrid.swift
darinwf: Maybe the environment is not inherited through `NavigationLink` somehow.

Client/Frontend/Browser/CardGrid/Grid/GridModel.swift
```diff
@@ -5,26 +5,79 @@
...
-            if case .spaces = switcherState {
+            if case .spaces = state {
                 ClientLogger.shared.logCounter(
                     .SpacesUIVisited,
                     attributes: EnvironmentHelper.shared.getAttributes())
             }
+            SceneDelegate.getCurrentSceneDelegate(with: tabManager.scene)?
+                .setSceneUIState(to: .cardGrid(state, tabManager.isIncognito))
+        }
+    }
```
evancrow, line 46: nit: should be on a new line

Client/Frontend/Browser/CardGrid/Grid/GridModel.swift
evancrow: Could it be an optional convenience parameter?

```
func update(state: SwitcherView, setSwitchModeWithoutAnimationTo: Bool?) {
    if let setSwitchModeWithoutAnimationTo = setSwitchModeWithoutAnimationTo {
        update(switchModeWithoutAnimation: setSwitchModeWithoutAnimationTo)
    }
}
```

darinwf: But in most cases we only set one of these variables, not both. There's really only one spot where it would help to combine them. It then seems better to have them be independent. Am I missing something?

evancrow: Hm alright, seemed more common to me, but it's not an issue. Thanks!

evancrow (approved):
LGTM!

Client/Extensions/ViewExtensions.swift
```diff
@@ -64,6 +65,22 @@ extension View {
         }
         .onAppear { updater(zero, one, two) }
     }
+
+    // Publisher variants. Useful when you just want to observe a particular published
+    // var of a model and not the entire model. Runs updater task asynchronously from
+    // onReceive to simulate the behavior of onChange and to ensure that the published
+    // var being updated has been updated by the time updater runs.
+    func useEffect<P>(_ p1: P, perform updater: @escaping () -> Void) -> some View
```
EdwardJTL, line 73: is this different from SwiftUI's built-in `onReceive`?

Client/Frontend/Browser/CardGrid/Card/CardDetails.swift
```diff
@@ -199,6 +199,10 @@ public class TabCardDetails: CardDropDelegate, CardDetails, AccessingManagerProv
             self?.objectWillChange.send()
         }.store(in: &subscriptions)
 
+        tab.$screenshotUUID.sink { [weak self] _ in
```
EdwardJTL, line 202: if this list gets too long, we can made a separate publisher like `tabCardRefreshSignal` that receives and erases the values from these three, and then put a subscriber on that publisher to send to object will change

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
```diff
@@ -2,38 +2,56 @@
...
 struct CardTransitionUX {
     static let animation = Animation.interpolatingSpring(stiffness: 425, damping: 30)
 }
 
+struct ScrollToCompletionModifier<Details: CardDetails>: ViewModifier {
+    let details: Details
+    let gridScrollModel: GridScrollModel
+
+    func runCompletion() {
+        if details.isSelected, let completion = gridScrollModel.scrollToCompletion {
```
EdwardJTL, line 17: is it possible for scrollToCompletion to have a race?
for example, caller 1 sets this call back, then model starts scrolling, then caller 2 sets the call back, then model did scroll gets fired.

Client/Frontend/Browser/CardGrid/Grid/EmptyCardGrid.swift
```diff
@@ -7,11 +7,15 @@ import SwiftUI
 
 struct EmptyCardGrid: View {
     let isIncognito: Bool
-    let isTopBar: Bool
     let showArchivedTabsView: Bool
 
+    @EnvironmentObject var chromeModel: TabChromeModel
```
EdwardJTL, line 12: doesnt this increase the number of times this view's graph gets evaluated?

Client/Frontend/Browser/CardGrid/Grid/GridModel.swift
```diff
@@ -5,26 +5,79 @@
...
+    init(tabManager: TabManager) {
+        self.tabManager = tabManager
+    }
+
+    @Published private(set) var state: SwitcherView = .tabs
+
+    func update(state: SwitcherView) {
+        if self.state == state {
+            return
+        }
```
EdwardJTL, line 41: nit: guard else (personal preference really)

Client/Extensions/ViewExtensions.swift
darinwf: Yes, notice how this also runs the callback from `onAppear`. This is modeled after the React `useEffect` API, I believe.

Client/Frontend/Browser/CardGrid/Card/CardDetails.swift
darinwf: Yeah, that's probably the better way to do it.

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
darinwf: Yes, there is a comment explaining the race that can happen. But the race you are describing is not a worry in practice given how `BrowserModel.show/hideGridWithAnimation` works. It has protections again the overlapping case.

Note, this PR isn't changing anything substantive with this system. It does have fragility though, and I'd love for a better solution, but definitely out of scope for this PR.

Client/Frontend/Browser/CardGrid/Grid/EmptyCardGrid.swift
darinwf: No, not really. This is just moving the dependency down to a leaf node. The parent used to have this dependency.

Client/Frontend/Browser/CardGrid/Card/CardDetails.swift
darinwf: I'm going to keep this as is for now just to minimize changes in this already complex PR.

Client/Extensions/ViewExtensions.swift
EdwardJTL: i see. It's also different in that it calls`DispatchQueue.main.async` here to reschedule the work. Wondering if that should be left as a choice on the caller to reschedule the event.

darinwf: Yeah, I thought about it, but I decided it was more valuable to keep both variants of `useEffect` consistent. The `onChange`-based variant also runs asynchronously via the `DispatchQueue` (as a side-effect of SwiftUI update processing).

EdwardJTL (approved):

PR #4349, opened 2022-09-09
premium; prime the apple uuid from the server

jedireza:
This PR pings the server to receive a UUID associated with the user's account before we call the purchase function on the product.

I'm not 100% sure what problem this solves for the backend, but it might help with some edge cases they've identified.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested this with the simulator and observed the calls via the Proxyman app.

N/A

## Issues addressed
N/A


Client/Frontend/Premium/PremiumStore.swift
```diff
@@ -59,13 +61,41 @@ class PremiumStore: ObservableObject {
...
             }
-
-            return false
         }
 
         return false
     }
 
+    func primeUUID() async -> String? {
+        await withCheckedContinuation { continuation in
```
chung-lai, line 70: Not super familiar with this API but looks good and we are calling `resume` in all cases.

chung-lai (approved):
LGTM

Issue #4346, opened 2022-09-09
Regression: Crash using HTTP Auth


darinwf:
There's a user report of a crash happening when the browser hits a page that requires HTTP auth (uncommon):

More details here:
https://github.com/neevaco/neeva/issues/74839

evancrow:
Good thing we disabled all the tests for this 🥲

Issue #1958, opened 2021-10-13
AuthenticationTests.testIncorrectCredentials() is flaky


chung-lai:
AuthenticationTests.testIncorrectCredentials() is flaky lately. It passes locally but fails often in CI.

Issue #1755, opened 2021-09-20
AuthenticationTests.testCorrectCredentials() is flaky


darinwf:
AuthenticationTests.testCorrectCredentials() is flaky

See for example:
https://app.circleci.com/pipelines/github/neevaco/neeva-ios-phoenix/4469/workflows/c045e50e-65e3-4cec-b1d8-f95ba0e8134c/jobs/13312

PR #4351, opened 2022-09-10
Fix HTTP Auth

evancrow:
Caused by: #4195, I guess we still need the LoginHelper code.

### How was this tested?

## Issues addressed
Fixed: #1755
Fixed: #1958
Fixed: #4346


Client/Frontend/Browser/ContentScripts/LoginsHelper.swift
```diff
@@ -0,0 +1,221 @@
...
+import Foundation
+import Shared
+import Storage
+import SwiftyJSON
+import WebKit
+import XCGLogger
+
+private let log = Logger.browser
+
+class LoginsHelper: TabContentScript {
```
darinwf, line 15: I'm surprised we still need this system for HTTP auth.

darinwf (commented):
I don't think we need the LoginsHelper part. The HTTP Auth system is just driven from a webview delegate method, right?

evancrow:
That's what I originally thought, but `Authenticator.handleAuthRequest` isn't ever called by the `WebView`, but by `LoginHelper.requestLogin`.

evancrow:
Wait, it doesn't look like it's called from there... I think I can remove it again.

darinwf (approved):

darinwf:
Hopefully the tests will be stable / not flaky given the tweaks you made.

evancrow:
Yeah I'll run them a few times to be sure.

evancrow:
Marking as cherry-pick-needed since this fixes a crash.

PR #4352, opened 2022-09-12
Only Close Today Tabs When CardStrip Open

evancrow:
When the CardStrip is visible, it's confusing when the user tries to close all the tabs from the URL bar it asks if they want to close more tabs than are visible. Should only close the visible tabs in the CardStrip (today tabs).

### How was this tested?
- [x] I tested this manually.


EdwardJTL:
could you add a screen recording? I can't figure out how to "close all the tabs from the URL bar"

evancrow:
https://user-images.githubusercontent.com/37356984/189793515-632810b4-d5d0-42f3-901c-e0d302e7b122.mp4

EdwardJTL (approved):
ran the code it works. LGTM.
should probably loop @Ria-yu in for design sync

Ria-yu:
LGTM, thanks for improving the UX!

PR #4354, opened 2022-09-12
allow empty url path in cheatsheet web result

EdwardJTL:
We don't need to filter root domains. some results will comeback with empty display paths

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


darinwf (approved):

darinwf:
It would be lovely to have some way to write a test for this :-)

I realize we would have some infrastructure to build.

macy-n (approved):

Issue #4082, opened 2022-07-25
Cookie cutter script not triggered on iOS 14


chung-lai:
When I was trying to test the cookie cutter onboarding screen on iOS 14, I was not able to trigger the onboarding flow. Looks like the script callback is not being invoked. We should either disable this feature or look into why it does not get triggered on iOS 14.

cc @evancrow

chung-lai:
Also, the cookie cutter setting layout is broken with the picker view.
<img width="564" alt="Screen Shot 2022-07-25 at 1 14 58 PM" src="https://user-images.githubusercontent.com/87332578/180866450-b2c8c707-1ae1-43f7-8d0e-a80df30ee916.png">

evancrow:
Think we're dropping support for iOS 14 soon, might make sense to just disable this for now.

chung-lai:
Think we should at least keep the tracking protection rules but disable cookie popup and ad blocking for iOS 14. @darinwf thoughts?

evancrow:
> Think we should at least keep the tracking protection rules but disable cookie popup and ad blocking for iOS 14. @darinwf thoughts?

Yup, that was my thought, just disable Cookie Cutter.

evancrow:
cc @jlynch630

darinwf:
From our discussion today, it seems there are two issues:

1- iOS 14 WebKit cannot handle as many rules as iOS 15 and later, and with easyprivacy + easylist, we exceed that limit.
2- cookie cutter engine script is not working properly on iOS 14.

One concern with just disabling these features is that we then have to support special casing on a variety of UI surfaces, both in settings but also in various promo and enablement flows. That might add another form of complexity.

For 1, we might be able to get away with just specifying easylist and excluding easyprivacy on iOS 14. Or maybe we just disable ad blocking altogether.

For 2, @jlynch630 is going to take a look and see if anything obvious jumps out. Maybe it is a trivial issue to fix.

darinwf:
Looks like easyprivacy is a supplemental addition to easylist, and if we have to choose between tracking prevention or ad blocking, we should probably stick with tracking prevention since that's more core to the product.

PR #4356, opened 2022-09-13
Set Minimum Version to iOS 15

evancrow:
- Sets minimum to iOS 15 for all target
- Sets CircleCI to use 14.0
- Removes all the `if #available(iOS 15.0, *)`

Bunch of a warnings I gotta fix real quick. With Swift 5.7 you have to set a value for `value` with `.animation`: `.animation(_, value: V)`.

### How was this tested?

## Issues addressed
Fixed: #4082


EdwardJTL:
i dont know if we are committing to dropping iOS 14 soon, but i think we can set CircleCI to Xcode 14 soon. I'd advocate for isolating the changes

evancrow:
Yeah, that's why I'm keeping this as a draft for now. I agree that it's wise to move out the CircleCI change though.

darinwf:
+1 to isolating changes

We have already unfortunately forced the issue with iOS 14. I think we will be clear to land this when the dust settles.

evancrow:
@EdwardJTL @darinwf Now that the iOS 14 build has been approved, can we consider merging this now?

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
```diff
@@ -64,7 +64,7 @@ struct CardTransitionModifier<Details: CardDetails>: ViewModifier {
                 overlayCard
                     .offset(x: rect.minX, y: rect.minY)
                     .frame(width: rect.width, height: rect.height)
-                    .animation(CardTransitionUX.animation)
+                    .animation(CardTransitionUX.animation, value: rect.minX)
```
darinwf, line 67: Why is it necessary to specify the `value` parameter now? Would it work to pass `rect` here if a value is really needed? The whole `rect` is changing during the transition. `minX` might not even change in some odd ball cases??

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
```diff
@@ -52,7 +52,7 @@ struct CardTransitionModifier<Details: CardDetails>: ViewModifier {
         content
             .zIndex(details.isSelected ? 1 : 0)
             .opacity(details.isSelected && cardTransitionModel.state != .hidden ? 0 : 1)
-            .animation(nil)
+            .animation(nil, value: 0)
```
darinwf, line 55: The SwiftUI documentation says that `value` is "A value to monitor for changes."
That tells me that this should be some state variable or expression derived from some observable property. Seems like `0` is not that.

evancrow: `cardTransitionModel.state` would work then right?

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
evancrow: I'll see if another variable is updating (to cause the animation) or just use `rect.width`.

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
evancrow: Using `cardTransitionModel.state` breaks the animation but rect.width works well

darinwf: Why is it necessary to specify a parameter here?

evancrow: Adding an `animation` without a `value` has been deprecated in Swift 5.7. It will be required to have some sort of value in the future.

darinwf: I see. But if not required at this time, then I think we should de-couple these changes from this PR. I worry about risk of regression and I'd like a bit more time to study what this parameter means and understand what values we should be providing.

darinwf (approved):

PR #4357, opened 2022-09-13
Update CircleCI Version to 14.0.0

evancrow:
Updates CircleCI to use Xcode 14

From: #4356


EdwardJTL:
maybe we should test on both iOS 15 and iOS 16 for now. @darinwf thoughts?

darinwf:
I agree, ideally we should run tests on both iOS 15 and iOS 16 simulators. I believe the status quo is just us using the latest iOS version supported by the version of Xcode we are using on CircleCI. Setting up older simulators requires installing those. That means a bit more setup for CircleCI, right?

evancrow:
Agreed. Let's at least get Xcode 14 working. Right now, the tests are timing out :(. I will have it fixed soon though.

EdwardJTL:
checked test-client, looks like the script is not handling flaky test correctly. It reports each time a test fails even though a later attempt passes

EdwardJTL:
also this
```
testToggleInUI(): XCTAssertFalse failed: threw error "Toggle's tap() and isOn() are currently unavailable for inspection on iOS 16. Situation may change with a minor OS version update. In the meanwhile, please add XCTSkip for iOS 16 and use an earlier OS version for testing."
```

evancrow:
Passed, finally!

EdwardJTL (approved):
good with me. we should probably get another approval since I pushed to this PR too :/

darinwf (approved):

PR #4361, opened 2022-09-13
Remove some unused frameworks

darinwf:
I think these became unnecessary when we removed the web3 code.


yozuysal:
Kanna is used in ShareTo for parsing html head for og tags.

darinwf:
> Kanna is used in ShareTo for parsing html head for og tags.

Thanks for catching that! I forgot to grep through that directory, and sure enough CI caught it as well.

Client.xcodeproj/project.pbxproj
```diff
@@ -5200,7 +5196,6 @@
 			);
 			name = SharedTests;
 			packageProductDependencies = (
-				E96D956E2832FDF100DD5BB3 /* Yams */,
```
EdwardJTL, line 5203: yams is used in testing

Client.xcodeproj/project.pbxproj
```diff
@@ -9332,14 +9321,6 @@
 /* End XCConfigurationList section */
 
 /* Begin XCRemoteSwiftPackageReference section */
-		1C45B366283C4F36008A1401 /* XCRemoteSwiftPackageReference "CodeScanner" */ = {
```
EdwardJTL, line 9335: i thought code scanner was used for the QR code sign in flow. @yasminyin96 is that correct?

darinwf: It is, but it is included twice. This is just removing the second instance.

jedireza (approved):
Build is passing. That should be the bar I think. 😉

EdwardJTL (approved):

PR #4363, opened 2022-09-14
premium; try restoring missing subscription entitlement

jedireza:
This PR attempts to restore a member's subscription if they're showing up as basic but have an app store entitlement to a product.

### How was this tested?
N/A

### Manual test cases
I didn't test this code path yet. It's a tricky path to get to from the welcome flow since I'd need to change my endpoint to `reza.local.neeva.dev`, which is only available under the settings menu. I might hack it and give it a try.

N/A

## Issues addressed
N/A


chung-lai (commented):
Change looks good, how frequent do we try to restore the subscription? Also, will the restore logic triggered when the user login with another basic account?

Client/Frontend/Premium/PremiumStore.swift
```diff
@@ -66,52 +98,93 @@ class PremiumStore: ObservableObject {
...
+        await withCheckedContinuation { continuation in
+            guard let uuid = self.appleUUID else {
+                continuation.resume(returning: false)
+                return
+            }
 
-    func priceText(_ plan: PremiumPlan?) -> String {
-        guard let product = self.getProductForPlan(plan) else {
-            return "FREE"
+            DispatchQueue.main.async {
```
darinwf, line 163: Why is this `DispatchQueue.main.async` needed? Seems like this code is already on the main thread and the `GraphQLAPI.shared.perform` call is already asynchronous.

jedireza: I moved this code from another place that had this wrapped. IIRC there was a crash since this could've happened off of the main thread. Also this is an async function which is executed in a `Task` and this ensures the GQL call is executed on the main thread.

Client/Frontend/Premium/PremiumStore.swift
```diff
@@ -66,52 +98,93 @@ class PremiumStore: ObservableObject {
         return false
     }
 
-    func primeUUID() async -> String? {
+    func getProductForPlan(_ plan: PremiumPlan?) -> Product? {
+        if plan == nil {
```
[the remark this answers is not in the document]
jedireza: Good call. Thanks.

jedireza:
> Change looks good, how frequent do we try to restore the subscription?
Only during a premium store initialization. I envision this to executed when a user goes into app settings.

> Also, will the restore logic triggered when the user login with another basic account? Want to make sure user won't be able to keep creating new premium account :)

Yes. I think this is something we should try to guarantee on the backend. Like if there is already an account associated with that purchase, not to grant premium to a different account.

Client/Frontend/Premium/PremiumStore.swift
darinwf: I see. I think you be sure that this code is already running on the main thread even if running via an async function. One way to check is to add an `assert(Thread.isMainThread)`.

That said, I'm okay if you want to just be conservative about it and preserve the existing code.

jedireza:
I'm hoping we can think through any potential changes we should consider on the client side. Even after approved, I'm going to wait to merge until the backend folks can chime in on if they want a new param or a separate mutation for this purpose.

jedireza:
After speaking with the folks working on the backend, we're going to create a new mutation for this case.

jedireza:
The latest changes instrument the new `RestoreAppleSubscriptionMutation` and moves the checking+fixing logic we had in the initializer into a separate function. This is more ergonomic and allows us to use it on demand.

📟 @chung-lai @darinwf

chung-lai (approved):
LGTM

Issue #4364, opened 2022-09-14
Regression: poor frame rate on card transition animation


darinwf:
Observed with build 1.58.2 (270) but not with the previous TestFlight build at 1.58.1 (269).

Here's what went into 270:
https://github.com/neevaco/neeva-ios/commits/Build-270

Here's what went into 269:
https://github.com/neevaco/neeva-ios/commits/Build-269

The only changes that differ in that interval are:
https://github.com/neevaco/neeva-ios/commit/0127399127bfecc04b1bd8863ff6f5b9ff93a174
https://github.com/neevaco/neeva-ios/commit/7c121827d43393f9747dbd4d907a81cece86feb0

Neither of those look like it could explain this performance issue. One possible difference could be toolchain though since 269 was built by @chung-lai and 270 was built by @EdwardJTL. Could there have been some differences in how the builds were produced that could explain performance differences?

The repro case is just showing and hiding the card grid. In the latest TestFlight, the animation appears choppy.

We also have https://github.com/neevaco/neeva-ios/commit/7d2d9e54e337d08f954e290d06a24c9328b571bf which landed recently but should have been included in Build 269.

darinwf:
Actually, I don't think toolchain differences can explain this issue. I went back and built the older version, and I see the difference before and after. I'll spend some more time bisecting to see if I can determine which PR caused this regression.

darinwf:
OK, this appears to be a regression from 7d2d9e54e337d08f954e290d06a24c9328b571bf.

darinwf:
One thing I'm noticing is a lot of calls to `getRows`, which is cheap, but it seems like it shouldn't have to be called while the card transition animation is happening.

```
>>> selectedRowId: 15
>>> getRows: 111
>>> TabGridSectionView.init: 97
>>> getRows: 112
>>> TabGridSectionView.body: 89
>>> TabGridSectionView.init: 98
>>> getRows: 113
>>> TabGridSectionView.body: 90
>>> TabGridSectionView.init: 99
>>> getRows: 114
>>> TabGridSectionView.body: 91
>>> TabGridSectionView.init: 100
>>> getRows: 115
>>> TabGridSectionView.body: 92
>>> TabGridSectionView.init: 101
>>> getRows: 116
>>> TabGridSectionView.body: 93
>>> TabGridSectionView.init: 102
>>> getRows: 117
>>> TabGridSectionView.body: 94
>>> TabGridSectionView.init: 103
>>> getRows: 118
[...]

PR #4367, opened 2022-09-14
Reduce Calls to getRows

evancrow:
Reduces calls to `getRows` by not calling `getRows` from the view unless the model has updated the data.

Similar to how `updateRows` is called.

### How was this tested?
- [x] I tested this manually

Much snappier now:

https://user-images.githubusercontent.com/37356984/190069642-2ba5518c-6020-4921-b387-eb344ccc45e3.mp4

Sections still good:

## Issues addressed
Fixed: #4364


Client/Frontend/Browser/CardGrid/Card/CardModels.swift
```diff
@@ -222,7 +223,8 @@ class TabCardModel: CardDropDelegate, CardModel {
             sections.append(.overAMonth)
         }
 
-        return sections
+        // Prevents sections with no cards (only headers) from being shown.
+        return sections.filter { (timeBasedNormalRows[$0]?.count ?? 0) > 1 }
```
darinwf, line 227: Do you see this having a significant impact on performance?

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -54,20 +56,22 @@ struct TabGridContainer: View {
...
+                containerGeometry: geom,
+                rows: rows
             )
         }
+        .padding(.horizontal, CardGridUX.GridSpacing)
+        .background(
+            GeometryReader { proxy in
+                Color.background
+                    .useEffect(deps: proxy.size) { newValue in
+                        cardStackGeom = newValue
```
darinwf, line 70: Is this `cardStackGeom` used for anything? I think maybe it can just be deleted.

Client/Frontend/Browser/CardGrid/Card/CardModels.swift
```diff
@@ -195,7 +196,7 @@ class TabCardModel: CardDropDelegate, CardModel {
 
         // Defer signaling until after we have finished updating. This way our state is
         // completely consistent with TabManager prior to accessing allDetails, etc.
-        self.objectWillChange.send()
+        self.rowsUpdated += 1
```
darinwf, line 199: Nice change. I think it makes the code easier to understand.

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -80,6 +84,13 @@ struct TabGridContainer: View {
         .useEffect(gridScrollModel.$needsScrollToSelectedTab) {
             scrollToSelectedRowId()
         }
+        .useEffect(deps: tabModel.rowsUpdated) { _ in
+            self.rows = tabModel.getRowSectionsNeeded(incognito: isIncognito).flatMap {
+                tabModel.getRows(for: $0, incognito: isIncognito)
+            }
+
+            self.allRows = tabModel.getRows(for: .all, incognito: isIncognito)
```
darinwf, line 92: I don't quite understand why we need to compute `allRows` and `rows` here.

Also, why not just maintain `rows` on `TabCardModel`? Wouldn't that ultimately be a bit simpler than needing to cascade down the data like this?

Client/Frontend/Browser/CardGrid/Card/CardModels.swift
evancrow: It does help reduce the number of calls made to getRows. Also means we don't need to check the if the rows are empty in the view, can just handle it in the model.

darinwf: Yeah, moving more work to the model is the way.

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
evancrow: Prefer not to keep `rows` on `TabCardModel`, since we would need to make it `Published`, and different models use rows differently. Easier for a view to only get the rows it needs.

darinwf: It can be kept there without being published. Agree it shouldn't be `@Published`. You can still use the same technique you are using in this PR to signal when it should be read again.

darinwf: But I'm okay with the approach in the PR. Okay for `rows` to be kept on the view as you have it here.

darinwf (approved):
LGTM!

PR #4369, opened 2022-09-14
Add GridVisibilityModel and make BrowserModel have no published vars

darinwf:


Client/Frontend/Browser/BrowserModel.swift
```diff
@@ -193,5 +185,13 @@ class BrowserModel: ObservableObject {
 
         self.overlayManager = overlayManager
         self.toastViewManager = toastViewManager
+
+        subscription = self.gridModel.visibilityModel.$showGrid.sink { [weak self] showGrid in
```
evancrow, line 189: Instead of creating a subscription, should  `scrollingControlModel` be passed to `visibilityModel`? `ShowToolbars` from there instead?

darinwf: I did it this way so that `GridVisibilityModel` would not need to know about `ScrollingControlModel`. I think that's valuable and worth the tiny bit of overhead. WDYT?

Client/Frontend/Browser/BrowserBottomBarView.swift
```diff
@@ -7,16 +7,17 @@ import Shared
...
 
 struct BrowserBottomBarView: View {
-    @EnvironmentObject var browserModel: BrowserModel
     @EnvironmentObject var chromeModel: TabChromeModel
     @EnvironmentObject var overlayManager: OverlayManager
+    @EnvironmentObject var gridVisibilityModel: GridVisibilityModel
 
     @ViewBuilder var toolbar: some View {
-        if !browserModel.showGrid && !chromeModel.inlineToolbar && !chromeModel.isEditingLocation
+        if !gridVisibilityModel.showGrid && !chromeModel.inlineToolbar
```
EdwardJTL, line 15: nit: do we want to put all of them on separate lines?

Client/Frontend/Browser/BrowserModel.swift
```diff
@@ -6,32 +6,28 @@ import Combine
 import Shared
 import SwiftUI
 
+// This is the root model for the browser and is available via @EnvironmentObject to
+// any View. It explicitly does not have any @Published vars. This way it is safe to
+// access via @EnvironmentObject without worry about that causing unwanted updates to
+// the View graph. However as this provides a means to access just about any model
+// from anywhere, it can be all too easy for this to open the door to spaghetti code.
```
EdwardJTL, line 13: is it possible to have some access control? 
1. we can make computed variables to read child model properties?
2. we can have a separate ticket to track the effort of making all the models `private`. Also might make a good method for examining if views actually need access to the brower model.

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
```diff
@@ -11,7 +11,9 @@ struct CardTransitionUX {
 
 struct ScrollToCompletionModifier<Details: CardDetails>: ViewModifier {
     let details: Details
-    let gridScrollModel: GridScrollModel
+
+    @EnvironmentObject var browserModel: BrowserModel
+    var gridScrollModel: GridScrollModel { browserModel.gridModel.scrollModel }
```
EdwardJTL, line 16: could we have a comment explaining why this needs to read the model from the environment vs just accepting the model as a param?

darinwf: This is just moving the dependency to a leaf node. Just did it to tidy up the parent View.

There is a comment on line 30 about why this View does not directly observe `GridScrollModel`.

Client/Frontend/Browser/BrowserModel.swift
darinwf: What kind of access control did you have in mind?

By computed properties, are you talking about shortcuts to things like `gridModel.visibilityModel` so that instead of writing `browserModel.gridModel.visibilityModel` we'd have `browserModel.gridVisibilityModel` or something like that?

A lot of the optimizations I've been putting in place to minimize View graph updating have come down to eliminating usage of `@EnvironmentObject`. But being able to access a class like `BrowserModel` that is not itself getting any updates, enables me to do things like you see in `ScrollToCompletionModifier` to selectively watch a subset of `@Published` vars on a model.

Maybe this points to the models not being setup optimally, but this pathway of selectively listening to `@Published` vars and not the entire model is one that seems valuable for improving performance. It means we need a way to get access to the model with those `@Published` vars without listening for the model itself through the environment.

One alternative is to pass models via constructor parameters. That can help us resist the spaghetti code, but as we have seen with BVC, it is still super easy to plumb god objects where they don't belong. So I'm not yet convinced that is the answer.

Frankly, I think it is going to take some iteration to land on a better approach. For now, just opening the door to some more of these optimizations feels appropriate. I think we can work to design our models to better match up with our views. They should be more ViewModel like than not. That's probably the right direction.

EdwardJTL: by computed properties, i mean writing like `browserModel.isGridVisible` that returns `browserModel.gridModel.visibilityModel.isVisible`, so that the model itself is not available, but the view can grab the right data when it updates.

I agree lets go with optimization first and see the number of view property and view updates go down. It won't be very maintainable since we dont have guards against spaghetti code; but we aren't actively working on any features so it's probably okay for now

Client/Frontend/Browser/CardGrid/Grid/CardTransitionModifier.swift
EdwardJTL: I see. it's a little bit surprising to see a view modifier that uses an environment object, and the reason for getting gridScrollModel this way was not immediately obvious to me.

darinwf: Got it. I'll add some more comments about this.

Client/Frontend/Browser/BrowserModel.swift
darinwf: I see. I had avoided writing those shortcuts because it might lead to people writing bad SwiftUI Views that depend on those computed properties without observing the correct underlying model (`GridVisibilityModel` in this case). Having to instead deal with the long paths helps make this more clear and avoid such mistakes. (I've seen us make those kinds of mistakes, which is why this is on my mind.)

Client/Frontend/Browser/BrowserModel.swift
evancrow: I agree this is lightweight, so in this case, I don't feel too strongly about removing it. It does make more sense to have the model that handles the view changing, call all the methods surrounding the change if that makes sense.

darinwf: Yeah... I think this code is still too complicated and twisted. It would be nice to isolate more of the card transition animation logic. I'll be giving that more thought... but for now, I think this is a reasonable simple step in the right direction.

EdwardJTL (approved):
Since this commit changes when views are updated, I'd advocate for keeping this change on testflight for at least a week before submitting it to app store once we get unblocked there.

darinwf:
> Since this commit changes when views are updated, I'd advocate for keeping this change on testflight for at least a week before submitting it to app store once we get unblocked there.

I like the cautious spirit! I think it is impossible to know when we will get unblocked by Apple, but perhaps at this rate, it will not be until next week's branch cut that we will have an opportunity for a new build. So now is probably a good time to land this and let it bake on TestFlight.

PR #4373, opened 2022-09-15
Upgrade SDWebImage and SDWebImageSwiftUI

darinwf:
Upgrade to latest versions of these libraries. This should fix the runtime error we are seeing with SDWebImageSwiftUI.


darinwf:
In my testing, this seems solid.

evancrow (approved):

Issue #4360, opened 2022-09-13
Regression: top bar height is slightly shorter when showing the card grid


darinwf:
Observed using 1.57.1 (266):

https://user-images.githubusercontent.com/79867437/189990930-049708f6-5b6d-47a9-aa51-2fa23f5f5737.mp4

Adding a print statement in `BrowserView` here:
<img width="719" alt="Screen Shot 2022-09-13 at 12 21 35 PM" src="https://user-images.githubusercontent.com/79867437/189991032-d257b5e7-7c67-4376-a298-c2716be52eb6.png">

On an iPhone 12 sim, I see the height switching between 54.5 and 50.

darinwf:
`GridPicker` has a fixed height of 50, given by `GridModel.pickerHeight`. That is given by `TopToolbarHeightWithToolbarButtonsShowing`, which is only meant to be used for landscape or tablet mode.

PR #4375, opened 2022-09-16
Fix TopBar Padding 

evancrow:
Ported from #3819
Fixed: #4360


evancrow:
Should be screenshots/recordings attached in the PR this was ported from.

jedireza (approved):

PR #4380, opened 2022-09-16
Reduce updates to non-animating cards

darinwf:
What this change does is change `Card` so that it does not directly monitor state variables like `showGrid`. This way all of the `FittedCard` instances that are part of the card grid will not need to be updated when `showGrid` changes, and only the `Card` created by the `CardTransitionModifier`(i.e., the overlay card) will animate in response to `showGrid` changing. This nicely cuts down on the number of updates happening to non-animating cards when performing a card transition animation.

With this change I see fewer `FaviconView.body` calls. There are still too many as other state changes are invalidating the entire `CardGrid` still, but that's a separate problem and will be a separate PR.

In addition, this PR eliminates the redundant `showsSelection` parameter to `Card` in favor of just using the newly added parameter.

Issue #4364


darinwf:
In case not obvious, with this change the `Card` will get re-constructed through the `CardTransitionModifier`, which is observing `GridVisibilityModel`. But `Card` will not get re-constructed through `FittedCard`, which is what the rest of the cards in the grid use since `FittedCard` does not observe `GridVisibilityModel`.

evancrow (approved):

PR #4382, opened 2022-09-16
Eliminate one update of CardsContainer when transitioning to / from card grid

darinwf:
Move `CardsContainer`'s dependency on `GridVisibilityModifier` into a modifier to isolate the impact of updates to that model. This dependency is just used to ensure that the spaces view is showing the first card. This change cuts down the number of card grid updates from 3 to 2 when transitioning to / from the card grid.

Other small improvements in this PR include ensuring that we only update `cardSize` when it changes. That way we avoid more unnecessary updates to SwiftUI. This PR also eliminates an unused `containerGeom` state variable.

Also fixes it so that scrolling to the top-most space works when the `enableSpaceDigestCard` experiment is disabled by using `SpaceCardViewModel.dataSource` instead of `SpaceCardModel.allDetails`.

Issue #4364


Client/Frontend/Browser/CardGrid/Grid/CardGrid.swift
```diff
@@ -133,13 +133,19 @@ struct CardGrid: View {
             columnCount = topToolbar ? 3 : 2
         }
 
+        // NOTE: Avoid updating state if the values didn't change. That way unnecessary
+        // and potentially expensive SwiftUI updates are avoided.
+
```
evancrow, line 138: nit: remove extra line

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -23,6 +23,22 @@ extension EnvironmentValues {
...
+    @EnvironmentObject var browserModel: BrowserModel
+    @EnvironmentObject var gridVisibilityModel: GridVisibilityModel
+
+    func body(content: Content) -> some View {
+        content.useEffect(deps: gridVisibilityModel.showGrid) { _ in
+            scrollProxy.scrollTo(
+                browserModel.gridModel.spaceCardModel.allDetails.first?.id ?? ""
+            )
+        }
+
```
evancrow, line 38: nit: remove extra line

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -23,6 +23,22 @@ extension EnvironmentValues {
 
 }
 
+struct ScrollToFirstSpaceModifier: ViewModifier {
+    let scrollProxy: ScrollViewProxy
+
+    @EnvironmentObject var browserModel: BrowserModel
```
evancrow, line 29: Should you just call `@EnvironmentObject var gridModel: GridModel`? Or is this to avoid unnecessary updates?

darinwf: Yup, this is to avoid unnecessary updates.

Client/Frontend/Browser/CardGrid/Grid/CardGrid.swift
darinwf: I had the new line here since the comment applies to both updating `columnCount` as well as `cardSize` and there is a newline between those two code blocks. This way it doesn't look like the comment is only applying to `columnCount`.

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -23,6 +23,22 @@ extension EnvironmentValues {
...
+    @EnvironmentObject var gridVisibilityModel: GridVisibilityModel
+
+    func body(content: Content) -> some View {
+        content.useEffect(deps: gridVisibilityModel.showGrid) { _ in
+            scrollProxy.scrollTo(
+                browserModel.gridModel.spaceCardModel.allDetails.first?.id ?? ""
```
EdwardJTL, line 35: what happens if scrollTo gets called with and id it cannot find? if it breaks we should only call scroll to if the id is not nil.

EdwardJTL: i see the code was like this before. it's okay if we don't change it

darinwf: `scrollTo` does nothing if called with an ID it can't find, so this should be safe.

I just moved this code in this PR, so this is the original logic. I could also make the change to only call `scrollTo` if there is a space card to scroll to.

darinwf:
OK, actually.... I just tried testing this out, and well, it doesn't work and it didn't work before. The first space in the list has an ID of "spaceDailyDigest", but that space is not rendered (for me), so as a result, it does not scroll to the top of the spaces list.

@yozuysal @buracules - Should we just delete this code or should we still try to scroll to the top of the spaces grid? Perhaps this code should skip over the first space if its ID is "spaceDailyDigest" and try to scroll to the second space in the list instead? Or should we remove this space from `allDetails` when it should not be shown?

darinwf:
> @yozuysal @buracules - Should we just delete this code or should we still try to scroll to the top of the spaces grid? Perhaps this code should skip over the first space if its ID is "spaceDailyDigest" and try to scroll to the second space in the list instead? Or should we remove this space from `allDetails` when it should not be shown?

I think this code should be using the `viewModel.dataSource` instead since that corresponds to what is actually rendered.

yozuysal:
> > @yozuysal @buracules - Should we just delete this code or should we still try to scroll to the top of the spaces grid? Perhaps this code should skip over the first space if its ID is "spaceDailyDigest" and try to scroll to the second space in the list instead? Or should we remove this space from `allDetails` when it should not be shown?
> 
> I think this code should be using the `viewModel.dataSource` instead since that corresponds to what is actually rendered.

Code archeology: So the initial intention of this scrollTo was to scroll to the top of the list and it preceded the addition of daily digest. But also back then, our tabs and spaces were inside the same ScrollView, meaning when you selected new tabs etc, what spaces were showing up in focus were changing as well. So this was to avoid scrolling down spaces randomly as we jump around inside the tab container. None of this might be needed today, but again the intention was to always always reset Spaces UI to the top of the list since CurrentSpace ~= LastEditedSpace ~= First Space in the list.

From your description, dailyDigest + introduction of dataSource might have caused a regression here already. Looks like the current version addresses this?

darinwf:
Thanks for taking a look @yozuysal. The PR does attempt to ensure that the spaces view is always scrolled to the top. It does it probably not at the most optimal time (when opening / closing the card grid). Would probably be way better to defer it until showing spaces. But the point of this PR is just to reduce other expensive updates caused by `GridVisibilityModel` changing, and it is enough to just move the code for that to a modifier as this PR does.

yozuysal (approved):

PR #4387, opened 2022-09-19
Fix TabGridContainer.selectRowId implementation

### LINKED ISSUE #4386: Regression: show tabs button does not work after doing a search (which results in a new tab)
Observed using TestFlight 1.59.4 (276):

If you do a search, which results in a new tab, then the "show tabs" will animate when tapped but not result in showing you the card grid.

Most likely this means we are missing a call to `updateAllTabDataAndSendNotifications` -- i.e., the newly created tab does not have a corresponding `TabCardDetails`.

darinwf:
When deciding to select the heading instead of the row, be sure to select the right heading (not just the pinned section). This requires searching for the heading within the set of rows.

Also be sure to actually set the Row.id on each generate row HStack. This way we can scroll to the specified row.

This fixes a bug where users would sometimes not be able to see the card grid when tapping on the "show tabs" button.

Also fixes the case of a param of getLogCounterAttributesForTabGroups.

Fixes #4386


darinwf:
I'm thinking about how to write a test for this so we don't regress again.....

Client/Frontend/Browser/CardGrid/Card/CardModels.swift
```diff
@@ -156,7 +156,7 @@ class TabCardModel: CardDropDelegate, CardModel {
 
         var id: Set<String> { Set(cells.map(\.id)) }
         var cells: [Cell]
-        var index: Int
+        var index: Int  // Index within TabSection
```
evancrow, line 159: Nit: put the comment above the variable

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -39,16 +39,20 @@ struct TabGridContainer: View {
...
-        }) {
-            if row.index == 1 {
-                return rows[0].id
+        var previous: Row!
+        for row in rows {
+            if row.cells.contains(where: \.isSelected) {
+                // If we are selecting the first row of cards in the tab section, then
+                // select the header instead (i.e., the previous row). This way the header
+                // will be visible too.
+                if row.index == 1 {
```
evancrow, line 48: Idea (not necessarily for this PR): Always scroll to the previous row. This way, `selectedTab` is always centered in the grid, and the user sees there are move tabs above/below.

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
```diff
@@ -39,16 +39,20 @@ struct TabGridContainer: View {
     }
 
     var selectedRowId: TabCardModel.Row.ID? {
-        if let row = rows.first(where: { row in
-            row.cells.contains(where: \.isSelected)
-        }) {
-            if row.index == 1 {
-                return rows[0].id
+        var previous: Row!
```
evancrow, line 42: Instead of creating a new variable, you could use the array extension `previousItem(before item: Element) -> Element?`. So: `return rows.previousItem(before: row).id`.

Client/Frontend/Browser/CardGrid/Grid/TabGridRowsView.swift
```diff
@@ -84,10 +84,11 @@ struct TabGridRowsView: View {
                             }
                     case .sectionHeader(let section):
                         TabGridSectionHeaderView(section: section)
-                            .id(row.id)
                     }
                 }
-            }.zIndex(row.cells.contains(where: \.isSelected) ? 1 : 0)
+            }
+            .id(row.id)
+            .zIndex(row.cells.contains(where: \.isSelected) ? 1 : 0)
```
evancrow, line 91: Do you know what this (`zIndex`) is used for? Do we still need it after all of the recent refactoring?

evancrow (approved):
LGTM! Few questions/nits/ideas :)

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
darinwf: That variant of `previousItem` is a bit expensive as it has to search through the entire array again. I wrote it this way to avoid the need to do that. Since the number of rows can vary according to how many tabs one creates, and we've had performance issues with too many tabs, I'd prefer to avoid using that `previousItem` function.

Client/Frontend/Browser/CardGrid/Grid/CardsContainer.swift
darinwf: I think we might need to play around with the second parameter to `scrollTo` to get that kind of effect. It's a good idea though. Will keep in mind for future revisions.

Client/Frontend/Browser/CardGrid/Card/CardModels.swift
evancrow: Yeah, it's no big deal. Just following the pattern for the rest of the app. Usually, we've been comments for a variable right above it. Up to you, though!

PR #4388, opened 2022-09-19
Reenable UGC BloomFilter Health Logs

EdwardJTL:
Objective:
- Re-enable UGC filter stats logging
- Not trigger the bug in `Defaults`

Changes:
- Add new logger that uses a queue to write to a in memory copy
- Write to `UserDefaults` directly when crucial information is changed, and when app is entering background

Other Considerations:
- Single serial queue to manage all in-memory mutations is probably better for not overloading GCD.
- Writing to `UserDefaults` directly seems to be safe to do on all threads without triggering `Defaults` library bug
- Writing to `UserDefaults` is not synced since `UserDefaults` is thread safe.

Future Work:
- If this util proves to not cause crashes, we can consider generalizing it to use it for other aggregated logging
- We could consider increasing the complexity to handle multiple scenes. Does not seem to be a priority to me.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually tested with thread sanitizer and observed logs


darinwf (approved):

darinwf:
I agree it is worth seeing if this doesn't crash.

PR #4393, opened 2022-09-20
Further isolate model updates to improve card transition performance

darinwf:
Pulls out switchModeWithoutAnimation from GridSwitcherModel into a new GridSwitcherAnimationModel to help isolate the impact of updates to this published variable during card grid transition animation.

Also inverts the sense of this state variable to avoid double negatives in the code. It becomes more simply switchWithAnimation, which is true by default.

This cuts down `CardsContainer.body` updates from 2 to 1 during the card grid transition animation.

Issue #4364


Client/Frontend/Browser/BrowserView.swift
```diff
@@ -157,6 +157,7 @@ struct BrowserView: View {
         .environmentObject(bvc.gridModel.scrollModel)
         .environmentObject(bvc.gridModel.spaceCardModel)
         .environmentObject(bvc.gridModel.switcherModel)
+        .environmentObject(bvc.gridModel.switcherAnimationModel)
```
evancrow, line 160: nit: If we're sorting alphabetically, this should be above `switcherModel`

evancrow: Same in other locations, but I also understand this _falls_ under `switcherModel` so no big deal.

evancrow (approved):
LGTM! Thanks for making the logic easier to read. One possible nit.

Client/Frontend/Browser/BrowserView.swift
darinwf: Oh good catch

PR #4395, opened 2022-09-20
Stop using WebImage for FaviconView to improve performance on iOS 16

darinwf:
There was some change in the iOS 16 SwiftUI implementation that has led to issues rendering favicons in the card grid. They always start out by showing fallback content before loading the actual image.

To workaround this issue we can just implement our own image fetcher and cache. The cache enables us to synchronously provide a UIImage to SwiftUI, which eliminates any flicker.

This change also helps avoid the issue of the favicon moving out of sync with the card during a card grid transition. It doesn't seem to completely eliminate that issue however.

This change should also help make the card grid transition smoother overall. However, I have not measured the impact.

The cache is written to in two places: 1) when loading a web page and fetching the favicon the first time, we store the image in this new cache, and 2) when there is a cache miss and we need to fetch the image again.

Issue #4364


Client/Frontend/Favicons/FaviconView.swift
```diff
@@ -23,19 +73,22 @@ struct FaviconView: View {
...
-            .transition(.opacity)
-            .scaledToFit()
+        if let image = fetcher.image {
+            Image(uiImage: image)
+                .resizable()
+                .background(Color(fetcher.backgroundColor))
+                .transition(.opacity)
+                .scaledToFit()
+        } else {
+            Color.blue
```
darinwf, line 91: Whoops, this is an error. Let me fix this.

EdwardJTL:
ran some profiling. I tested the case where we have a single tab, and we enter tab switcher.

I'm noticing that 5 FaviconFetcher's get created, but only one persists. and FaviconView gets updated five times.

There is probably something else that is "unequating" the FaviconViews. I am wondering if we could have the FaviconFetcher as a state object. Instead of creating new ones, each time we just update the subscription to listen to a different resolver and drop duplicates where we can.

darinwf:
> There is probably something else that is "unequating" the FaviconViews. I am wondering if we could have the FaviconFetcher as a state object. Instead of creating new ones, each time we just update the subscription to listen to a different resolver and drop duplicates where we can.

Yeah, I had started wondering about using a `@StateObject` too but didn't go further into playing around with it. I'll give it a try! Thanks for the suggestion.

I don't know that SwiftUI is determining that a FaviconView is unequal to an existing one. Rather, the whole card grid is getting updated and `CardsContainer.body` is being executed again. That causes new `FaviconView` instances to get created and/or causes `FaviconView.body` to get executed again.

This is why I have been going around trying to reduce overall `CardsContainer.body` updates.

darinwf:
Thinking about it some more. I think using `@StateObject` would not work in this case since the `fetcher` needs to get constructed from the parameters passed to the `View`. One way to make the `fetcher` not have to get created again would be to store it on `TabCardDetails` and pass that down to the `View`, but then we'd end up holding onto a `UIImage` per tab, which is what we are currently trying to avoid doing. The cache here seems to provide the desired behavior and I'm not sure that moving it into `TabCardModel` / `TabCardDetails` would really be better. I like the current PR in that case.

Client/Frontend/Favicons/FaviconHandler.swift
```diff
@@ -55,6 +56,10 @@ class FaviconHandler {
                 deferred.fill(Maybe(success: (favicon, data)))
             }
             FaviconResolver.updateCache(for: site, with: favicon)
+
+            if let img = img {
+                FaviconFetcher.updateCache(for: favicon.url, with: img)
+            }
```
EdwardJTL, line 62: do we need to write to our cache here? i assume SDWebImage will write to its own cache,

darinwf: This is just an optimization so that the favicon is already there in the cache when you open the card grid. So the journey of click a link and open the card grid will avoid a flicker of the favicon as the card grid animation runs.

darinwf: I can add that as a comment.

EdwardJTL (approved):

PR #4396, opened 2022-09-20
Enable ArchivedTabResign

evancrow:
### How was this tested?


evancrow:
Empty data text:

darinwf:
One behavioral change I noticed:

When clicking on a tab in the archived tabs view, it no longer loads that tab. Instead, it just shows the card grid with that tab selected.

I think this is not very ideal for two reasons: 1) I think the user more likely wants to see that page if they click on it in the archived tabs view and 2) we don't have a thumbnail saved for archived tabs so all you see in the card grid is a tab with a blank thumbnail--which doesn't look so great.

darinwf (commented):
Otherwise, this change LGTM.

evancrow:
Fixed:

https://user-images.githubusercontent.com/37356984/193710135-14723357-b38f-4487-bbb7-da170c5e5bf4.mp4

darinwf (approved):

PR #4397, opened 2022-09-21
Fix Client's code signing

chung-lai:
- Revert the code signing not to be managed automatically

That breaks the local build script.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


chung-lai:
@EdwardJTL, It was changed to become automatically managed in #4388, did we do it on purpose?

darinwf (approved):
Seems correct to put this back

PR #4401, opened 2022-09-22
Set a fixed width on the Card's favicon+label so that it does not re-layout during card transition animations

darinwf:
This way they stay together as the card transition animation runs, and it also avoids the issue of the ellipsis being stuck at the right most end of the card during transition animation, which was really odd.

Before:

https://user-images.githubusercontent.com/79867437/191634771-6ddf6c45-eb97-4ce4-9db1-ccebd632a515.mp4

After:

https://user-images.githubusercontent.com/79867437/191635176-69ec7af1-f0fb-451e-a196-97bf16788729.mp4


evancrow (approved):
Looks so much better!

Issue #4403, opened 2022-09-22
[UI] Odd double bar under tab switcher


jonathangarelick:
See immediately above the Yesterday label

<img src="https://user-images.githubusercontent.com/41495718/191849222-c72379f3-60a2-4814-a322-63b028305247.jpg" height="400">

iOS 16.0, TestFlight 1.60.1 (279)

evancrow:
Think this is caused by the toolbar padding adjustment... will fix.

PR #4404, opened 2022-09-22
Change Where TopBar Vertical Padding is Added

evancrow:
Caused by #4375 adding the padding in the wrong location.

Simulator Screen Shot - iPhone 14 Pro - 2022-09-22 at 14 17 38

https://user-images.githubusercontent.com/37356984/191853099-d5bed685-bd8c-45a2-b07b-2da583fdaeff.mp4

## Issues addressed
Fixed: #4403

### DIFF
--- a/Client/Frontend/Browser/BrowserTopBarView.swift
+++ b/Client/Frontend/Browser/BrowserTopBarView.swift
@@ -5,10 +5,10 @@
 import Defaults
 import SwiftUI
 
-private enum BrowserTopBarViewUX {
+enum BrowserTopBarViewUX {
     static let GridPickerBottomPadding: CGFloat = 4.5
     static let ShowHeaderTapAreaHeight = 32.0
-    static let SwitcherToolbarViewBottomPadding: CGFloat = 7.5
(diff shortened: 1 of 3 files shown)


darinwf:
Are you sure this is the intended UI design? I think it would be good to confirm with @Ria-yu. That extra bar above the section title is, I think, intended as a divider between the sections. Could also be that we simply don't need to have that bar at the top of the first section?

darinwf:
Or hmm... I think this is the design here:
https://www.figma.com/file/t1jW7cahkClVHAIlnHJuv8/Tab-Manager-(Spaces-%26-Tab-Groups)?node-id=3264%3A12992

Maybe the colors are off too for dark mode?

evancrow:
The extra padding between the top bar and the grid was unintentionally introduced. This reverts to the design we've used for a while, where the first section header top bar blends in with the browser top bar.

 I do agree that removing the first section's top bar makes sense.

evancrow:
Yeah, all the colors look wrong.

evancrow:
Well the padding is still wrong, even if the colors are wrong too. Should still merge this.

PR #4406, opened 2022-09-23
Create build github action

chung-lai:
* Add github action with build.yml
* Add a new `build.rb` script to do daily build and branch cut as github action (written in Ruby since fastlane is in Ruby and we may reuse it in the future)
* Trigger build and deploy by running the action

Tested daily build by creating another private repo with the same directory structure and test drove all those scripts

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


.github/workflows/branch_cut.yml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/checkout@v3
+
+      - name: Initialize mandatory git config
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email noreply@github.com
+
+      - name: Run branch cut and daily build script
+        run: |
+          export branch_build_number=`fgrep "BROWSER_PROJECT_VERSION = " Client/AppConfig.xcconfig | uniq | cut -d' ' -f3 | cut -d';' -f1`
```
darinwf, line 21: What about moving more of this into a helper script? That way the circle ci configuration stays a bit simpler.

Scripts/build.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  # Branch cut
+  branch_cut_name = "Build-#{get_build_number}-release-branch"
+  puts "Branch cut: Creating branch #{branch_cut_name}"
+  execute_shell_command("git checkout -b #{branch_cut_name}")
+  bump_version("patch")
+  commit_and_push_version_bump(false)
+
+  # Switch back to main and bump minor version
+  execute_shell_command("git checkout main")
```
darinwf, line 123: I don't have much experience with Ruby. What happens if `execute_shell_command` fails? Does the program exit at that point?

chung-lai: good question, just tried it out. It does exit the script when there is an error from the shell command

chung-lai: Spoke too soon, it doesn't bail when there is non-zero exit code. Will add the check for that. Good catch!

buracules (approved):
lgtm

Issue #4409, opened 2022-09-23
Crash writing to UserDefaults on a background thread


darinwf:
From 1.60.1 via TestFlight:

```
Incident Identifier: 8A68934A-E3AB-4354-ABF4-1277DD63B8BE
Hardware Model:      iPhone14,7
Process:             Client [3753]
Path:                /private/var/containers/Bundle/Application/78691A46-F211-40FD-8EAA-08EF16D6324B/Client.app/Client
Identifier:          co.neeva.app.ios.browser
Version:             1.60.1 (279)
AppStoreTools:       14A305
AppVariant:          1:iPhone14,7:16
Beta:                YES
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           co.neeva.app.ios.browser [743]

Date/Time:           2022-09-23 08:50:03.7951 -0700
Launch Time:         2022-09-23 08:43:16.1937 -0700
OS Version:          iPhone OS 16.0.1 (20A371)
Release Type:        User
Baseband Version:    1.00.05
Report Version:      104

Exception Type:  EXC_BREAKPOINT (SIGTRAP)
Exception Codes: 0x0000000000000001, 0x000000018a8c10fc
Termination Reason: SIGNAL 5 Trace/BPT trap: 5
Terminating Process: exc handler [3753]

Triggered by Thread:  10


Thread 0 name:
Thread 0:
0   libobjc.A.dylib               	0x000000018a8bdc20 objc_msgSend + 32
1   CoreFoundation                	0x00000001915263b0 cow_cleanup + 140 (NSCollectionAux.h:48)
2   CoreFoundation                	0x000000019156044c -[__NSDictionaryM dealloc] + 148 (NSDictionaryM.m:440)
3   CoreFoundation                	0x000000019154106c -[CFPrefsSearchListSource alreadylocked_getDictionary:] + 900 (CFPrefsSearchListSource.m:1265)
4   CoreFoundation                	0x00000001915187a4 -[CFPrefsSearchListSource alreadylocked_copyValueForKey:] + 172 (CFPrefsSearchListSource.m:639)
5   CoreFoundation                	0x00000001915186d8 -[CFPrefsSource copyValueForKey:] + 52 (CFPrefsSource.m:894)
[...]

PR #4410, opened 2022-09-23
Write bloom filter logs to UserDefaults on main

EdwardJTL:
Write to user defaults on main to not trigger observer crash

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

## Issues addressed
#4409


Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetLogger.swift
```diff
@@ -183,6 +183,12 @@ class CheatsheetLogger {
 
     /// Write in memory values to UserDefaults
     private func writeToPersistentStorage() {
+        if !Thread.isMainThread {
+            DispatchQueue.main.sync {
```
darinwf, line 187: I'm a little nervous about the `sync` calls back to the main thread. It means we have to be sure that the main thread is never blocked waiting on this thread. I think that shouldn't be a problem today, but it does become a thing we need to keep in mind going forward.

I was wondering though... what about using https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver instead? This form of `addObserver` lets you specify a queue where the notification should run. Maybe that would help us feel confident that we would not miss this notification when the app is being backgrounded?

Maybe we could then just use `DispatchQueue.main.async` here in case we find ourselves not on the main thread? (Thinking about the other call to `writeToPersistentStorage`.)

WDYT?

EdwardJTL: `addObserver` seems like a right move. Will try that.

I'm not sure I understand "we have to be sure that the main thread is never blocked waiting on this thread". I think it will be this thread that is waiting for the main thread.

I chose not to use `.async` because I found in testing that when i used a `.async` with a lower priority queue, the write operation may not finish executing since the function returns right away. Using `.sync` avoided that.

darinwf: > I'm not sure I understand "we have to be sure that the main thread is never
> blocked waiting on this thread". I think it will be this thread that is waiting for
> the main thread.

Right, this PR makes it so that this background thread sync waits on the main
thread. If in the future someone makes the main thread sync wait on this thread,
then there could be a dead-lock. That's the point I was making. It is something
we can decide to be okay with. Just pointing out the risk and wondering if there
is a way to avoid using `DispatchQueue.main.sync` to prevent the risk.

I get that we need to worry about the task getting actually scheduled on the main
thread, so simply using `DispatchQueue.main.async` here would not work reliably.

What I was wondering is if the other variant of `addObserver` that allows you to
specify a destination queue would be a reliable way of delivering that notification
to the main thread and therefore avoid the need to use `DispatchQueue.main.sync`.

Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetLogger.swift
```diff
@@ -117,7 +124,9 @@ class CheatsheetLogger {
 
             // Write crucial logs straight away
             if attributes.contains(.numOfUGCClears) {
-                self.writeToPersistentStorage()
+                DispatchQueue.main.sync { [self] in
```
darinwf, line 127: Do these ones need to be `sync` though? If we aren't worried about shutdown because that is already handled, then why can't these just run whenever via `async`?

EdwardJTL: if we use `async` here, it unblocks this serial queue and allows it to process the next update to the in memory copy. Then when `main` read from in memory copy, it could be mutated at the same time by another worker on the background thread. so we can get things like "1 UGC hit and 0 UGC test" written to user defaults because reading from all the in-memory values is not atomic.

EdwardJTL: that mostly happens during the start up sequence because the bloom filter is getting initialized, and those logs are emitted in close proximity to the first few url tests.

once that is done, the url tests and clears should be well distributed

darinwf: I see. So if such a data race is possible, then we should guard against it with a mutex, no? We have to worry about it in the app being backgrounded case too, right?

By the way, this file needs a meaty comment explaining these very subtle details involving multiple threads and synchronization. It was not obvious that you were using the `sync` calls to both get on the main thread and to guard against data races.

EdwardJTL: will add comments once we converge on a design 👍

I am worried about using a mutex/lock/semaphore because it could cause a priority inversion when the lower priority memory writer has the lock, and the higher priority user defaults writer is waiting on the lock to read the in memory copy. Currently, with the on app background case, I'm being optimistic in assuming that it will likely happen during a time when there are not many writes to the in memory copy.

open to other ideas. Went through a couple ideas myself. my priority is to avoid thread explosion or causing blocks/priority inversion on main; in doing so I am being optimistic and accepting slight errors in some cases.

darinwf: The lock-free option is to pass-by-value to the main thread. Can we do that instead? Just define a `struct` for the data and pass that along to the main thread where it gets written to `UserDefaults`.

darinwf: Although I guess that is not a complete solution since the `appMovedToBackground` call comes in on the main thread.

darinwf: If in the lock all you are doing is copying some data (and that data is very small), then contention on the lock will be exceptionally rare. So I think you shouldn't need to worry too much about it causing thread priority inversions.

darinwf: My recommendation is to protect access to the data objects that are used on multiple threads and stick with the `addObserver` method that schedules `appMovedToBackground` via the `main` queue.

darinwf: Btw, a way to minimize lock contention is to avoid making the calls to `UserDefaults` while holding the lock. Just copy data with the lock held, and then write the copied data to `UserDefaults`.

EdwardJTL: Added comments.
Added a call to synchronously get a copy on the `updateQueue` and pass it to the writer. 

Synchronizing on dispatch queue is one way to get a lock. I'm not sure if it fits to use a regular OSLock or DispatchSemaphore, since the lock will need to be grabbed when mutating the in memory counters as well. "app enter background" could be called while we are mutating the in memory copy. If we decide to wait, then we achieve almost the same effect as waiting on `updateQueue`; the difference is that waiting on queue also waits for all the current blocks on the queue, while it is unclear which thread will get priority when multiple threads are waiting on a lock/semaphore. Another consideration is that calling `queue.sync` might be an easier semantic for swift to understand, and it carries the execution context. I am hoping that this gives GCD more context to better schedule the jobs.

Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetLogger.swift
```diff
@@ -170,35 +203,55 @@ class CheatsheetLogger {
...
         }
     }
 
     /// Clear values stored in UserDefaults
+    ///
+    /// Writing to UserDefaults must be done on main to avoid crash triggered by KVO in Defaults library
     private func resetPersistentStorage() {
+        assert(Thread.isMainThread)
+
         self.incrementCounters.keys.forEach { attribute in
```
darinwf, line 255: This is also reading from `incrementCounters` on the main thread. Can we count on the set of keys being stable?

Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetLogger.swift
```diff
@@ -134,18 +153,32 @@ class CheatsheetLogger {
...
         }
     }
 
     // MARK: - Private Methods
-    @objc
+    /// Write current in memory values to user defaults when app is backgrounded
+    ///
+    /// This function acquires a copy of the counters by waiting on the `updateQueue`.
+    /// This ensures that we get a safe copy of the values that are currently in memory
+    /// As a result, this function must not be called from the `updateQueue` to avoid deadlocks.
```
darinwf, line 169: I think you should add a comment about why this function needs to be called on the main thread. <- So that we intercept the app backgrounded notification on a thread where we can safely write out the values.

Note, probably overkill at this point, but it could also be sensible to hide the data that can only be accessed on the `updateQueue` behind an interface that can assert that that rule is always followed. (i.e., this could look like moving `self.incrementCounters and self.stringCounters` into a class that only lives on the `updateQueue` and guards all of its accesses with checks that ensure it is only be accessed on that queue.)

Often it can be a very good design to have classes restricted to a single thread. And then use messaging (dispatch queue) to access them as you are doing here. In this case we have a single class being used on multiple threads.

Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetLogger.swift
```diff
@@ -134,18 +153,32 @@ class CheatsheetLogger {
...
     private func appMovedToBackground() {
-        writeToPersistentStorage()
+        dispatchPrecondition(condition: .notOnQueue(self.updateQueue))
+        let copy = self.updateQueue.sync {
```
darinwf, line 172: I think this makes sense in that it is making it so that the only synchronization cost happens at shutdown. Clever! I think it is probably a bit fancy / fragile when a simple mutex would work well here, be very robust to future editors of this code and not likely to be a performance problem :-)

Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetLogger.swift
EdwardJTL: in the current implementation, keys are initialized on `init`, and we never remove entries from the dict, so it should be.

Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetLogger.swift
EdwardJTL: added comment

also did refactor. It does help separate concern and make code more readable. :)

Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetLogger.swift
EdwardJTL: the check is now moved. added an additional comment to give reason to why we must wait for the value on a lower priority thread from main

Client/Frontend/Overlay/Content/Cheatsheet/Logger/CheatsheetSessionUsageLogger.swift
```diff
@@ -0,0 +1,351 @@
...
+    private var stringCounters: [CheatsheetStringAttribute: String]
+    // Queue used to synchronize in memory counter updates
+    private let updateQueue = DispatchQueue(
+        label: bundleID + ".CheatsheetAttributeStorage.updateQueue",
+        qos: .utility
+    )
+
+    // MARK: - Static Properties
+    static var bundleID: String {
+        Bundle.main.bundleIdentifier ?? "co.neeva.app.ios.browser"
```
darinwf, line 57: nit: Just like `AppInfo.baseBundleIdentifier` does, I think you can use `Bundle.main.bundleIdentifier!` here instead of hardcoding a fallback value.

darinwf (approved):
OK, thanks for taking so much care on this! Looks good...

The only other observation I have (almost afraid to mention it!), is why not just proxy over to the main thread whenever `increment` or `setStrings` are called and have `inMemoryStorage` be an object that is accessed only on the main thread? Wouldn't everything just get loads simpler then? Is there a compelling reason to mutate this synchronously on the thread that calls `increment` / `setStrings`?

EdwardJTL:
> OK, thanks for taking so much care on this! Looks good...
> 
> The only other observation I have (almost afraid to mention it!), is why not just proxy over to the main thread whenever `increment` or `setStrings` are called and have `inMemoryStorage` be an object that is accessed only on the main thread? Wouldn't everything just get loads simpler then? Is there a compelling reason to mutate this synchronously on the thread that calls `increment` / `setStrings`?

if we are hopping to main every time to write to an in memory copy, it's not too different from writing to UserDefaults directly each time. I am worried that we will be crowding the main thread with low priority logging tasks, especially since we are likely to use this type of logger more widely as we are moving to more aggregated anonymized logging.

chung-lai:
@EdwardJTL do we want to merge this sooner to make it for cherry-picking into tonight's build so we have more time to test out the change?

darinwf:
> if we are hopping to main every time to write to an in memory copy, it's not too different from writing to UserDefaults directly each time. I am worried that we will be crowding the main thread with low priority logging tasks, especially since we are likely to use this type of logger more widely as we are moving to more aggregated anonymized logging.

Good point that you are not sending a message to the main thread with each call to `increment` or `setStrings`.

(One other thing to keep in mind: `DispatchQueue.main.async` is also like acquiring a mutex that the main thread will also try to acquire, so I'm not sure all of the effort to avoid an explicit mutex is worth it.)

PR #4413, opened 2022-09-25
Correct Colors to Match Mocks

evancrow:
As discussed in #4404, all the `CardGrid` colors are mismatched and incorrect. This corrects the colors to what is used in Figma.

### How was this tested?
- [x] I tested this manually

**Light Mode:**

**Dark Mode:**


evancrow:
> I think something is still missing. Take a look at this side-by-side comparison:
> 
> Screen Shot 2022-09-25 at 4 26 36 PM
> 
> In the mock, the Today section has a dark grey background (#1C1C1E if I am reading it correctly).

The colors look different in different mocks. @Ria-yu Can you confirm the right bg color for card grid sections and the separator?

evancrow:
Merged in #4404

evancrow:
Everything is now inline with the mocks:

evancrow:
The new icon is currently behind a flag and is waiting to be enabled by #4396

Ria-yu:
> Looks pixel perfect! I did notice a discrepancy between the Archived Tabs icon in the design vs. actual (box vs. clock). Please confirm the correct icon with @Ria-yu then we should be good to go.

For context, we started with clock icon when Archived tabs first came out, and introduced this new icon when we decided to combine History and Archived Tabs into the same page with a picker where History keeps the clock icon and Archived Tabs gets this archive-ish looking icon to distinguish the two.

PR #4418, opened 2022-09-26
Sanitize SessionUUIDs from Cheatsheet Logs

EdwardJTL:
Remove sessionUUID and sessionUUIDv2 from cheatsheet logs, since most cheatsheet logs contain some sensitive data.

Only place where we keep sessionUUIDV2 is `CheatsheetUGCStatsForSession` and the two popover impression logs

More Details:
- `sessionUUID` is added automatically when calling `EnvironmentHelper.getAttributes()`. Remove all calls to those in cheatsheet.
- `sessionUUIDv2` is added automatically when logging if the log passes `LogConfig.shouldAddSessionID(for path:)`. For cheatsheet logs, changed the logic for this function to only allow the aforementioned three paths.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually observed logs


Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -291,16 +271,39 @@ class CheatsheetMenuViewModel: ObservableObject {
...
+                ),
+                ClientLogCounterAttribute(
+                    key: LogConfig.CheatsheetAttribute.currentCheatsheetQuery,
+                    value: query ?? "nil"
+                ),
+            ]
+            + attributes
 
-        // Evaluate computed `loggerAttributes` synchronously
-        let allAttributes = envAttributes + loggerAttributes + attributes
```
EdwardJTL, line 299: this is why we had duplicate logging attributes. the three sources could overlap and produce attributes with the same key multiple times

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
```diff
@@ -103,8 +103,7 @@ final class CheatsheetServiceProvider: CheatsheetDataService {
                     DispatchQueue.main.async {
                         ClientLogger.shared.logCounter(
                             .CheatsheetBadURLString,
-                            attributes: EnvironmentHelper.shared.getAttributes()
```
chung-lai, line 106: Once we remove `SessionUUID`, we should be able to continue sending these `EnvironmentHelper.shared.getAttributes()` attributes since they shouldn't have any other joining ID

chung-lai (approved):
LGTM!

macy-n (approved):

PR #4420, opened 2022-09-26
premium; log when storefront is nil

jedireza:
Since we've observed `SKPaymentQueue.default().storefront` coming up `nil` in otherwise valid contexts, I'd like to see how often this happens and what the `Locale.current.identifier` value is. We might need to switch to something more reliable.

### How was this tested?
N/A

### Manual test cases
N/A

N/A

## Issues addressed
N/A


chung-lai (approved):
LGTM

chung-lai:
Removing cherry pick label as it made it to release branch

PR #4428, opened 2022-09-27
request the ios as reviewer

chung-lai:
Somehow it doesn't recognize neevaco/ios as reviewer. Trying to change to `ios` as reviewer but not 100% sure this will work. If this doesn't work, will think about alternative on who to assign as reviewer :(

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


darinwf (approved):

chung-lai:
Added a github action playground so we can do testing when we iterate on the action. Merging now, the change is very safe as it doesn't touch any core code.

PR #4431, opened 2022-09-27
welcome flow; new copy

jedireza:
This PR updates the copy for parts of the welcome flow. I'm still waiting for confirmation on the climate pledge copy, but besides that this PR is good for a review.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

image

## Issues addressed
N/A

### DIFF
--- a/Client/Frontend/WelcomeFlow/WelcomeFlowPlansView.swift
+++ b/Client/Frontend/WelcomeFlow/WelcomeFlowPlansView.swift
@@ -9,15 +9,14 @@ struct WelcomeFlowPlansView: View {
     @ObservedObject var model: WelcomeFlowModel
 
     var premiumBullets = [
-        ("Browser + ad blocker", ""),
-        ("Tracking prevention", ""),
+        ("Browser + ad blocker + tracking prevention", ""),
         ("Unlimited ad-free, private search", ""),
-        ("Premium password manager + VPN", ""),
+        ("Unlimited devices", ""),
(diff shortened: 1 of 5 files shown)


Client/Frontend/WelcomeFlow/WelcomeFlowPlansView.swift
```diff
@@ -9,15 +9,14 @@ struct WelcomeFlowPlansView: View {
     @ObservedObject var model: WelcomeFlowModel
 
     var premiumBullets = [
-        ("Browser + ad blocker", ""),
-        ("Tracking prevention", ""),
+        ("Browser + ad blocker + tracking prevention", ""),
         ("Unlimited ad-free, private search", ""),
-        ("Premium password manager + VPN", ""),
+        ("Unlimited devices", ""),
```
chung-lai, line 14: Since we cast this into `LocalizedStringKey` when we display on the screen. We have to add this string to `Localizable.strings` file (similar to what have been done [here](https://github.com/neevaco/neeva-ios/pull/4264/files) ). Otherwise, it doesn't get exported for translation.

chung-lai (approved):
LGTM. Couple things:

* I assume we will address the truncated climate pledge will be addressed
* translation won't happen in time for this release given it takes a couple days for translator to get back; however, given international users won't see the premium plan screen so we should be good, just want to flag it

PR #4442, opened 2022-09-30
session id allow list

chung-lai:
* Limit `sessionUUIDv2` on an audited set of events so no user information can be joined
* Remove the old `SessionUUID`

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


Client/Logging/LogConfig.swift
```diff
@@ -416,6 +417,56 @@ public enum LogConfig {
         case StorefrontWasNil
     }
 
+    public static let sessionIDEventWhiteList: Set =
```
clischeske, line 420: Might be worth adding a note here that any changes / additions to this list also need to be updated in cmd/httpd/logging/counter_filter_allow_list.go

macy-n (approved):

PR #4443, opened 2022-09-30
Merge ViewExtension Files

evancrow:
Built from #4433.

Merges two files under `Shared`, both named `ViewExtension`, into one.


evancrow:
Yeah that makes sense, I'll follow your suggestion above thanks.

PR #4447, opened 2022-10-03
Fix bottom tool bar item sizing

EdwardJTL:
When system font changes, the bookmark button reacts to it and scales, while the others dont. 

This change makes all of them fixed sized by adding support for using `fixedSize` fonts when initializing `Symbol` with `custom` font

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
see screenshots. Taken at a higher than regular text size. Look at the bookmark icon

| Before | After |
| ---- | ----- |
| IMG_0210 | IMG_0211 |


Client/Frontend/TabChrome/Toolbar/TabToolbarButtons.swift
```diff
@@ -228,7 +229,7 @@ enum TabToolbarButtons {
     }
 
     struct AddToSpace: View {
-        let weight: NiconFont
+        let weight: Font.Weight
```
darinwf, line 232: Actually, it was intentional to use the Nicon font for this. Notice how the aspect ratio of the bookmark icon is slightly different in the Nicon font than in SFSymbol font. That's intentional. Can we fix this issue another way?

darinwf (approved):
Makes sense! (The other toolbar buttons do not get bigger, so this one should stay fixed too.)

Issue #4210, opened 2022-08-12
Donation to Siri Suggestion appears to be broken


EdwardJTL:
search and visit url should be donated. not seeing them

EdwardJTL:
seems like search donations are still being donated, but url visits are not

EdwardJTL:
fixed in #4456

PR #4456, opened 2022-10-04
Fix missing intent donations

EdwardJTL:
1. unify new search & edit search code paths so that both donate search intent
2. add `isEligibleForPrediction` property to non-incognito activities in order to donate them to siri

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually observed donations by enabling it in developer mode

## Issues addressed
#4210


darinwf (approved):
Is there any way to write a test to help provide coverage for this? Could it make sense to mock out the calls to the OS and then write a test to confirm they get called as expected? That way this is unlikely to break again?

EdwardJTL:
> 

we can do a wrapper to test that we are making the mocked calls, but we can't test if we are really making the OS calls correctly

darinwf:
> > 
> 
> we can do a wrapper to test that we are making the mocked calls, but we can't test if we are really making the OS calls correctly

Right, that's what I was imagining we'd do. It would help ensure that we always make the OS calls we expect to make. We can trust that the OS probably does the right thing with those calls (at least for the sake of such tests).

EdwardJTL:
> > we can do a wrapper to test that we are making the mocked calls, but we can't test if we are really making the OS calls correctly
> 
> Right, that's what I was imagining we'd do. It would help ensure that we always make the OS calls we expect to make. We can trust that the OS probably does the right thing with those calls (at least for the sake of such tests).

I just remembered that we decided to not show website visits in Siri suggestion. The reason was that those suggestions show up without being initiated by the user, so it can appear as if Neeva is tracking/keeping a record what they are viewing

darinwf (approved):
LGTM!

PR #4459, opened 2022-10-04
Eliminate unnecessary spring animation added to scrollview fling

### LINKED ISSUE #2696: Still seeing issues w/ scroll rebound animation being jumpy
Still seeing issues w/ scroll rebound animation being jumpy

Perhaps we need to adjust some of the parameters on the `interactiveSpring` in `ScrollingControlModel`.

See video here (iPad Pro 12.9 inch Sim):
https://user-images.githubusercontent.com/79867437/151040786-016c5b14-6490-42a7-aea6-4c5cb8463e85.mp4

I've also seen this repro on my old iPhone 6s.

@evancrow FYI

darinwf:
When scrolling the page up or down and hitting the extents of the scrollview, you will often see the scrollview jitter a bit as it hits those bounds. It does not smoothly rubberband like normal scrollviews. This is due to the `withAnimation` around the assignment of `headerTopOffset` and `footerBottomOffset`.

Fixes #2696


evancrow (approved):
LGTM! I tried it out, and it works well.

PR #4464, opened 2022-10-05
Enable logging statistics on app start up

EdwardJTL:
- Add logging method that bypasses incognito check
- Move cheatsheet aggregated logging to app delegate to prevent re-submitting logs

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
Manually tested that logs are registered
Manually tested that logs do not get re-submitted on subsequent app entering foreground


EdwardJTL:
future work in #4465

chung-lai (approved):

Client/Logging/ClientLogger.swift
```diff
@@ -63,22 +63,53 @@ class ClientLogger {
...
+            performLogMutation(clientLog)
+        }
+        loggingQueue.removeAll()
+    }
+
+    // MARK: - Private Methods
+    private func logCounter(
+        _ path: LogConfig.Interaction,
+        attributes: [ClientLogCounterAttribute],
+        enforceIncognito: Bool
```
chung-lai, line 97: Think it may be cleaner to have one method:

```
    func logCounter(
        _ path: LogConfig.Interaction,
        attributes: [ClientLogCounterAttribute] = [],
        byPassIncognito: Bool = false
    )
```

Thoughts?

Client/Logging/ClientLogger.swift
```diff
@@ -63,22 +63,53 @@ class ClientLogger {
...
+    }
+
+    // TODO: - Document usage
+    /// Register a logging event without performing incognito checks
+    ///
+    /// This method is only intended to be used with payloads that had already been scanned for incognito
+    /// For example, on such payload is aggregated statistics from an app session, where the statistics collected also
+    /// do not collect incognito usage.
+    /// - Warning: This method does not verify if BVC is in incognito beforing transmitting the logs. Please read documentation on its intended usage
+    func logCounterBypassIncognito(
```
chung-lai, line 80: Is the purpose having `BypassIncognito` as part of the function making it more explicit? Somehow it feels more natural to me to make BypassIncognito as an additional parameter in the function. And that makes me think more about this library method. See comment below for thought.

chung-lai (approved):
LGTM, just a thought about the logCounter function

Client/Logging/ClientLogger.swift
EdwardJTL: yea i did it this way so that
1) its easier to track down where this function is used within Xcode. it's hard to search for a function and also filter by a specific value passed to a param
2) by having it as a separate function, we can write a longer warning description lol....

PR #4467, opened 2022-10-05
add client counters

macy-n:
added following client counters that doesn't include journey id

        case PreviewNavigationOutbound
        case PreviewNavigationInbound
        case PreviewSearch

        case SkOpenCheatsheet
        case SkCloseCheatsheet
        case SkCheatsheetEducationImpressionOnSRP
        case SkCheatsheetEducationImpressionOnPage
        case SkAckCheatsheetEducationOnSRP
        case SkAckCheatsheetEducationOnPage
        case SkShowCheatsheetEducationOnSRP
        case SkShowCheatsheetEducationOnPage
        case SkShowCheatsheetContent
        case SkLoadCheatsheet
        case SkCheatsheetUGCStatsForPage
        case SkCheatsheetEmpty
        case SkOpenLinkFromCheatsheet
        case SkCheatsheetQueryFallback
        case SkOpenCheatsheetSupport
        case SkCheatsheetFetchError


paktree (approved):

chung-lai (approved):
LGTM! Will let Edward also review the cheatsheet part

Also, as a follow up after the release, if we are planning to keep both shrink and full version of the log for a while, we may want to do some refactoring as we may want to have a common helper that we can specify what extra attributes we should send in the full version and log it with the right counter path.

macy-n:
> LGTM! Will let Edward also review the cheatsheet part
> 
> Also, as a follow up after the release, if we are planning to keep both shrink and full version of the log for a while, we may want to do some refactoring as we may want to have a common helper that we can specify what extra attributes we should send in the full version and log it with the right counter path.

Thanks! I think as we are prioritizing structure logging, that should solve the issue.

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetMenuViewModel.swift
```diff
@@ -141,6 +141,17 @@ class CheatsheetMenuViewModel: ObservableObject {
             ]
         )
 
+        self.log(
+            .SkLoadCheatsheet,
+            attributes: [
+                ClientLogCounterAttribute(
+                    key: LogConfig.CheatsheetAttribute.currentPageURL,
```
EdwardJTL, line 148: are these ones going to be added to the sensitive path allow list? if so they shouldn't have urls

macy-n: good catch, updated.

EdwardJTL (approved):

Issue #4327, opened 2022-09-01
Zero query broken in first run


EdwardJTL:
## Expected behavior

Zero query model should be repopulated upon returning from spaces.
Suggestion and text in url bar should be synced.

## Actual behavior

Returning to zero query from space, or pressing clear in after typing, will clear suggestions.

If you press cancel when you type, the suggestions stay but the text is deleted. When you tab the bar again, the text is gone.

## Steps to reproduce the problem

1. Fresh install the app
2. Get to zero query screen
3. go to a space, then return to zero query. its empty

1. Fresh install the app
2. get to zero query screen
3. Type something, then press cancel

## Hardware and software
building from main @ 63a4caa9ddd1ede15290d4d193033dfee0fc3034

PR #4469, opened 2022-10-05
Reattempt to fix first run navigation blank screen

chung-lai:
Reattempt fix for: #4327

* There is different behavior on `onDisappear` being called on iOS 15.5 which causes the first run flow dismissing itself.

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


evancrow (approved):

Client/Frontend/ZeroQuery/RecommendedSpacesView.swift
```diff
@@ -29,6 +29,13 @@ struct RecommendedSpacesView: View {
                         manager: SpaceStore.suggested),
                     state: expandSuggestedSpace
                 )
+                .onDisappear {
```
evancrow, line 32: It would be good to add a comment explaining this

Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -1055,6 +1055,10 @@ extension BrowserViewController {
 
     func urlBarDidLeaveOverlayMode() {
         updateInZeroQuery(tabManager.selectedTab?.url as URL?)
+        if !Defaults[.didFirstNavigation] {
```
evancrow, line 1058: Ditto

Issue #4189, opened 2022-08-10
"Settings" option in overflow menu appears selected when tapping on "Support"


darinwf:
## Expected behavior

Only "Support" should highlight.

## Actual behavior

Tapping on "Support" results in both "Settings" and "Support" being highlighted. "Settings" becomes unhighlighted as you scroll the overflow menu up.

## Screenshots or recordings

https://user-images.githubusercontent.com/79867437/184020302-c63a9d9a-b5e4-4a8b-b2f8-560ad14f52de.mp4

## Hardware and software

- Your device: iPhone 13 simulator
- OS version: 15.5
- App version: Neeva Browser 1.54.0-dev (256)

darinwf:
Still seeing this issue on iOS 16. Maybe we are doing something wrong that is causing this glitch.

PR #4470, opened 2022-10-06
Fix Multiple GroupedButton Show As Selected

evancrow:
Cause by the first button being directly on the border of the `SafeArea`, which caused the `tableCell` effect to transfer to the button below the `SafeArea`. The effect works fine without the `ignoreSafeArea` modifier.

### How was this tested?
- [x] I tested this manually

Fixed:

https://user-images.githubusercontent.com/37356984/194424473-b55e5648-940a-49cc-9bb4-5f2d9697a343.mp4

## Issues addressed
Fixed: #4189


darinwf (approved):
woohoo, thanks for fixing this!

EdwardJTL (approved):

PR #4474, opened 2022-10-07
Fix account info crash

EdwardJTL:
we submit a logger event on background inside concurrency context

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
manually tested and stepped through to make sure there's no crash


jedireza (changes requested):
This main thread stuff is a foot gun.

EdwardJTL:
> This is main thread stuff is a foot gun.

We need to get the constants on main since we don't have a better design atm. Is there a suggestion here?

jedireza (approved):
I thought I clicked approve before.

PR #4480, opened 2022-10-10
premium; better region validation

jedireza:
This PR extends the `isOfferedInCountry` function we use limit who premium is offered to. We experienced false negatives where the default `Storefront` instance is `nil`. To help work around this we're extending the function to check the locale region as a fallback.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator

N/A

## Issues addressed
N/A


Client/Frontend/Premium/PremiumStore.swift
```diff
@@ -58,19 +62,19 @@ class PremiumStore: ObservableObject {
...
             }
-        } else {
-            ClientLogger.shared.logCounter(
-                .StorefrontWasNil,
-                attributes: [
-                    ClientLogCounterAttribute(
-                        key: LogConfig.Attribute.locale, value: Locale.current.identifier
-                    )
-                ]
-            )
```
jedireza, line 73: I'm cleaning up the logger to help reduce future clutter. It was a short lived log and I'm not sure we're going to get much more value out of it.

I do wonder if we should log anything else if both of the checks fail. But wasn't sure what else to capture. Open to ideas on this.

chung-lai: Did we see this log in the wild and wonder what's the locale in that case?

chung-lai (approved):

Client/Frontend/Premium/PremiumStore.swift
jedireza: Yep. I pinged you on Slack. But there were a lot of cases where the locale was `en_US`.

PR #4483, opened 2022-10-10
premium; en, fr, de

jedireza:
This PR adds the UK, DE, FR to the countries/regions we offer premium in.

### How was this tested?
N/A

### Manual test cases
N/A

N/A

## Issues addressed
🔒 https://app.clickup.com/t/36034046/GR-2450


chung-lai (approved):

jedireza:
FYI: this PR switched gears and is based on language and not locality.

chung-lai (approved):
Change looks good.

I remember reading somewhere that the languageCode can be `en-GB` for UK (something like what's mentioned here: https://www.ibabbleon.com/iOS-Language-Codes-ISO-639.html) May need to do prefix. I could be wrong, worth checking UK app store to make sure.

jedireza:
Thanks for calling that out. I went in an changed my Language to `English (UK)` and Region to `United Kingdom` in the simulator's settings and when printing out the value I got:

```
print("***LANG:\(lang) (\(Locale.current.identifier))")

***LANG:en (en_GB)
```

So I think we're going to be okay. Do you think there are other cases this might not be the case?

Issue #4482, opened 2022-10-10
Cookie Popups # does not reset after navigating to new page


chung-lai:
1. Perform a neeva search (cookie pop up # stays 0)
2. Install new app
3. Navigate to a site with cookie popup
4. Perform a neeva search (cookie pop up # now reports 1)
5. Open Neeva shield

https://user-images.githubusercontent.com/87332578/194924991-8387eb54-6d75-4f5a-8a0e-2d26b2a8deb8.mov

Expected: cookie pop up # reset to 0

Actual: cookie pop up # still report 1

cc @evancrow

evancrow:
This is probably because we show 1 in the menu if there was a cookie blocked on the domain before. So we probably blocked a popup on the marketing page, and now for every Neeva domain, we show 1 in the menu. 

Should restrict that domain filter more.

PR #4485, opened 2022-10-10
Update the Cookie Cutter Engine

evancrow:
Updated the Cookie Engine to 0.0.3, which fixes a bug where the Cookie Popup and Tracker count wouldn't get reset when navigating to a different page.

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #4482


chung-lai (approved):
LGTM, thanks!

PR #4486, opened 2022-10-10
premium; update benefit copy

jedireza:
This PR is updating the premium bullet point copy.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator

image

image

## Issues addressed
N/A


Client/Frontend/Settings/Neeva/NeevaPremiumView.swift
```diff
@@ -28,15 +28,15 @@ struct NeevaPremiumView: View {
     }
 
     var premiumBullets = [
-        ("Browser + ad blocker", ""),
-        ("Tracking prevention", ""),
+        ("Browser + ad blocker + tracking prevention", ""),
```
chung-lai, line 31: Since these are strings, we will need to add these bullet strings into `Localizable.strings`. Compiler is not smart enough even they are casted into LocalizedStringKey later on :D

chung-lai: Something like this: https://github.com/neevaco/neeva-ios/blob/main/Client/en.lproj/Localizable.strings#L149-L171

chung-lai (approved):
LGTM!

Client/en.lproj/Localizable.strings
```diff
@@ -147,25 +147,30 @@
...
 
 /* Bullet point text to describe Neeva benefit */
 "Ad-free, private search" = "Ad-free, private search";
 
 /* Bullet point text to describe Neeva free account benefit */
-"50 searches/week" = "50 searches/week";
+"50 searches/month" = "50 searches/month";
+
+/* Bullet point text to describe Neeva device limits */
+"Unlimited devices" = "Unlimited devices"
```
chung-lai, line 169: it's missing semi-colon so the build failed

Issue #4481, opened 2022-10-10
Have to scroll up by a lot to get toolbars to show


darinwf:
I think this has been an issue for a while. I got some user feedback that this makes it difficult to use the app. I may have just gotten used to the behavior and learned to deal with it. Now that I see it though, it does seem like it requires too much work to get the toolbars to show. See video:

https://user-images.githubusercontent.com/79867437/194906811-e31f987d-38db-40db-a84c-4260e0deb5c5.mp4

Other browsers like Chrome and Safari immediately show toolbars when you make even the smallest upward scroll gesture. We should too.

PR #4487, opened 2022-10-11
Make Show Toolbars More Sensitive

evancrow:
A couple of changes here:

- Removes the check to `decelerate`, which would return false at slow scroll speeds.
- Removes the animation for `scrollView.contentOffset` and replaces it with the header/footer offset. This makes the transition at slower drags smoother and has the same effect as before at faster scrolls. It also does not cause the scroll bounce bug like the previous animation.

### How was this tested?
- [x] I tested this manually

Also made sure #2696 did not return.

https://user-images.githubusercontent.com/37356984/194977733-48472111-a8ee-4a1e-b87a-7adaf719d6c1.mp4

## Issues addressed
Fixed: #4481


Client/Frontend/Browser/ScrollingControlModel.swift
```diff
@@ -324,9 +325,7 @@ extension ScrollingControlModel: UIScrollViewDelegate {
             return
         }
 
-        if (decelerate || (toolbarState == .animating && !decelerate))
-            && checkScrollHeightIsLargeEnoughForScrolling()
-        {
+        if checkScrollHeightIsLargeEnoughForScrolling() {
```
evancrow, line 328: Removed the check to `decelerate` because it prevented the toolbars from showing when scrolling up at slow speeds or short distances. Originally it was meant to prevent this exact case, but we've decided users should be able to quickly show the toolbars rather than guard against tiny movements.

Client/Frontend/Browser/ScrollingControlModel.swift
```diff
@@ -267,22 +267,23 @@ extension ScrollingControlModel {
...
                     withDuration: duration, delay: 0, options: .allowUserInteraction,
-                    animations: animation, completion: { _ in })
+                    animations: setOffset, completion: { _ in })
             } else {
-                animation()
+                setOffset()
+            }
+
+            if isShownFromHidden, let scrollView = self.scrollView {
+                scrollView.contentOffset = CGPoint(
```
evancrow, line 285: This was moved from `animation`, now `setOffset`, because putting it in the animation block caused jank when the toolbars became visible at slow speeds. Moving the offsets into the `UIAnimation` block kept the transition _smoother_ at all speeds. From @jonathangarelick comments above ^, some more work not related to this PR needs to be done.

PR #4490, opened 2022-10-11
update Neeva Shield setting menu

chung-lai:
Update Neeva shield setting menu:

* Show tracker & ad block section before cookie pop up setting
* Make tracker protection mode as inline picker instead of segmented picker
* Make "Accept non-essential cookie" as inline options instead of navigating to a new page
* Add subtitle to "Ad Block" option

| Decline | Accept |
| --- | --- |
| Screen Shot 2022-10-11 at 1 37 23 PM | Screen Shot 2022-10-11 at 1 37 21 PM |

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Ria-yu (commented):
Looks good to me! Nit: we wanted to have a custom spacing: 16px for blocks within a section (as shown in image), in order to have clearer visual hierarchy. Could you please double check if this is doable? Thanks!
Edit: fixed wrong image uploaded
Settings

chung-lai:
> Looks good to me! Nit: we wanted to have a custom spacing: 16px for blocks within a section (as shown in image), in order to have clearer visual hierarchy. Could you please double check if this is doable? Thanks! Settings

Turns out that's the default spacing between sections and so far haven't found ways to override it. @evancrow, do you know if we can control the spacing between sections in list?

evancrow:
You should be able to add padding to either the section header or the top/bottom of one of the section elements.

Client/Frontend/Overlay/Content/Menu/TrackingProtection/TrackingMenuSettingsView.swift
```diff
@@ -27,20 +27,14 @@ struct TrackingSettingsSectionBlock: View {
...
         Section(
-            header: Text("TRACKERS"),
-            footer: contentBlockingStrength == BlockingStrength.easyPrivacy.rawValue
-                ? Text(
-                    "Blocks most trackers. Minimizes disruption to ads and other functionality."
-                )
-                : Text(
-                    "Blocks more trackers. May disrupt ads and other functionality on some sites."
-                )
+            header: Text("TRACKER AND AD BLOCKING")
```
evancrow, line 30: `Padding` can be added to the `Text` I think.

Client/Frontend/Settings/CookieCutter/CookieCutterSettings.swift
```diff
@@ -9,9 +9,25 @@ struct CookieCutterSettings: View {
     @Environment(\.onOpenURL) var openURL
     @EnvironmentObject var cookieCutterModel: CookieCutterModel
 
-    @State var showNonEssentialCookieSettings = false
     @State var cookieCutterEnabled: Bool
 
+    @ViewBuilder
+    var footer: some View {
```
evancrow, line 15: Would name this `cookieCutterFooter`, since it's the footer for only the Cookie Cutter section.

Client/Frontend/Settings/CookieCutter/CookieCutterSettings.swift
```diff
@@ -29,45 +45,25 @@ struct CookieCutterSettings: View {
...
-                                    showNonEssentialCookieSettings = true
-                                }
-
-                                cookieCutterModel.cookieNotices = .userSelected
-                            }
-                        )
+                        Text("Accept Non-essential Cookies")
+                            .tag(CookieNotices.userSelected)
                     }.labelsHidden().pickerStyle(.inline).accessibilityLabel("Cookie Popups")
+
```
evancrow, line 60: nit remove extra line

Client/Frontend/Settings/CookieCutter/NonEssentialCookieSettings.swift
```diff
@@ -11,42 +11,38 @@ struct NonEssentialCookieSettings: View {
...
-                    }
-            ) {
-                Toggle(isOn: $cookieCutterModel.marketingCookiesAllowed) {
-                    DetailedSettingsLabel(
-                        title: "Marketing Cookies",
-                        description: "Tracks your online activity for advertising purposes"
-                    ).padding(.vertical, 8)
+        Section(
+            footer:
+                VStack(alignment: .leading) {
```
evancrow, line 16: Is this `VStack` still necessary? (I see you only removed the indent, but it doesn't look necessary)

chung-lai:
> You should be able to add padding to either the section header or the top/bottom of one of the section elements.

Right, I tried to play around with the padding on those but seem to have side effect for the cells :(

Client/Frontend/Settings/CookieCutter/NonEssentialCookieSettings.swift
chung-lai: interesting, don't think it's necessary. I was removing the `List` and look like the VStack was there before.

Client/Frontend/Overlay/Content/Menu/TrackingProtection/TrackingMenuSettingsView.swift
chung-lai: So we want to reduce the spacing between the tracker mode and ad block section, i tried to change the padding on footer, and play around with padding on the section but don't seem to help with reducing the spacing for that

evancrow: Could you embed the tracker and ad block sections inside a single `VStack` (inside the section and header) and control the spacing there? Or does that break everything else?

chung-lai: when i change it to:

```
VStack() {
  Section() {
  }

  Section() {
  }
}
```

It looks like this, guess it groups everything together 😓 :

Screen Shot 2022-10-11 at 2 38 04 PM

evancrow: Yikes! I don't have any more ideas off the top of my head, would need to look into it more.

[This thread](https://stackoverflow.com/questions/62509987/reducing-space-between-sections-for-grouped-list-in-swiftui) discusses reducing the footer height via UIKit.

Ria-yu: Thanks both! If it's unreasonably hard to change the spacing, I'll compromise and go with the system standard (original) one. Lmk

chung-lai: Got it. Just tried to table view option but doesn't seem to work either. I'll leave this for now and go with the system standard. fyi @Ria-yu

evancrow (approved):

Issue #4379, opened 2022-09-16
Bottom bar bug when pasting URL for the first time


jedireza:
I noticed when pasting in the URL bar the first time that the bottom bar gets really big. It only seems to happen after I see the paste permission dialog.

https://user-images.githubusercontent.com/979929/190710095-7eb44c0e-159a-4d4c-8ac5-9f26a2f5e90a.MP4

darinwf:
@evancrow - any ideas what could be causing this bug?

darinwf:
@jedireza - which version of the app are you seeing this in?

We could probably try doing a bisect across existing TestFlight builds and then custom builds from different commits to see which commit caused this regression (assuming it is a regression).

evancrow:
No idea. I'll take a look though.

evancrow:
It also looks like the GridSwitcher bg colors are wrong.

evancrow:
After more investigation, here's what I know:

The copy/paste popover breaks the SafeArea, pushing up the `BottomBar`. That popover has been confirmed by Apple (in most cases) as a bug. I wonder if it's best to leave this for now and let Apple fix it on their end. Once the popover is dismissed, the SafeArea still thinks the keyboard is there and doesn't update to the correct value.

evancrow:
I'm reading that Apple fixed this in 16.1 B2, but it still occurs on my device :(

evancrow:
This has been fixed in 16.1 Beta 3

evancrow:
My guess: next month. I've been trying to find a fix by `ignoringSafeArea` but that modifier has no effect on the bottom bar for no reason I could find.

PR #4491, opened 2022-10-12
Actually Fix the Bottom Bar Bug

evancrow:
Half-height `BottomBar` bug was caused by the safe-area value getting set high by the "Allow Paste" iOS 16 feature and not being reset on its dismissal. This bug is fixed by ignoring the safe area. 

I came across [this thread](https://stackoverflow.com/questions/63968241/swiftui-in-ios14-keyboard-avoidance-issues-and-ignoressafearea-modifier-issues) that had a method that worked for ignoring the keyboard `SafeArea`. (I had some prior issues where `ignoreSafeArea` wouldn't work on the `BottomBar`).

Also did some refactoring.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/195257811-33eca8b8-70c3-455c-826b-53d6797f8c0d.mp4

## Issues addressed
Fixed: #4379


darinwf:
@evancrow - could you expand on how this fix works? (i know you explained it to me over Zoom, but would be good to have it written down for others to see too, thanks!)

Client/Frontend/Browser/BrowserBottomBarView.swift
```diff
@@ -8,31 +8,25 @@ import SwiftUI
 
 struct BrowserBottomBarView: View {
     @EnvironmentObject var chromeModel: TabChromeModel
-    @EnvironmentObject var overlayManager: OverlayManager
```
darinwf, line 11: nice to eliminate a dependency

darinwf (approved):

evancrow:
Updated my description.

evancrow:
I win tests!!

PR #4498, opened 2022-10-13
language based filtering for subscription

chung-lai:
* Filter subscription based on preferredLanguage instead of Locale.current.languageCode. (languageCode only return app's localized language so in our case, it will always be `en`, `de` and `fr`)
* Add PremiumEligible logging to identify user eligibility

**Test:**

Change device language and not seeing the plan screen when it's not in our whitelist

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


jedireza (approved):

Client/Frontend/Premium/PremiumStore.swift
```diff
@@ -65,10 +65,19 @@ class PremiumStore: ObservableObject {
...
+            // preferredLanguage returns in format en, de-US...
+            let components = preferredLanguages[0].components(separatedBy: "-")
+            if components.count > 0 && PremiumStore.languages.contains(components[0]) {
                 return true
             }
+        } else {
+            // will offer premium in the case there is no preferredLanguage
+            // we want to track how frequent this case happens
+            ClientLogger.shared.logCounter(.NoPreferredLanguage)
+            return true
```
jedireza, line 80: I’m not sure we should return true here. With this, the function always returns true.

chung-lai: right, in the case when we don't see any preferred language, we consider user as eligible for now. Again, this case should be rare and we will be looking to see how often it happens.

chung-lai: but i do see there can be confusion. I just reversed the logic so it's more obvious :)

jedireza (approved):
Thanks for working through this!

PR #4505, opened 2022-10-15
Prevent Row Animations When Scrolling History

evancrow:
This disables the animation of the objects inside the History/Archived Tabs `ScrollView`, which would look bad when scrolling through a long list.

### How was this tested?
- [x] I tested this manually

Before:

https://user-images.githubusercontent.com/37356984/196004060-37062de9-9c49-4cd0-b426-031a3f39f9b8.MP4

After:

https://user-images.githubusercontent.com/37356984/196004043-0303f2b0-535b-4d45-88b3-d058bde4c354.MP4


evancrow:
Interesting! Will need to apply that everywhere else where `animation(nil)` is used as well.

evancrow:
`transaction.disablesAnimations = true` did not work, so stayed with `transaction.animation = false`

Issue #4506, opened 2022-10-16
Search bar layout issue on iOS 15


chung-lai:
Looks like it's happening on the latest internal build and only happen on iOS 15 simulator/device:

<img width="564" alt="Screen Shot 2022-10-15 at 5 23 37 PM" src="https://user-images.githubusercontent.com/87332578/196012245-71d71474-89ee-4142-a31c-1fc8d42e6f50.png">

cc @evancrow

PR #4507, opened 2022-10-17
Fix Large Search Bar [iOS 15]

evancrow:
Caused by #4491 when `.navigationBarHidden(true)` was moved outside of the `NavigationView` stack. This caused the `NavigationBar` to show on iOS 15 devices, causing the extra top padding. Fixed the bug by moving `.navigationBarHidden(true)` inside the `NavigationView`.

### How was this tested?
- [x] I tested this manually

## Issues addressed
Fixed: #4506


chung-lai (approved):
LGTM

PR #4510, opened 2022-10-18
Fix Cookie Education Popover Height

evancrow:
Made sure the Cookie Education UI does not fill the entire iPad screen, just takes up the space it needs.

### How was this tested?
- [x] I tested this manually

Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-10-17 at 16 29 08


chung-lai:
This release we will be launching the new ad block onboarding pop up so user won't see this one anymore :(

evancrow:
Oh, can just close this then :)

PR #4512, opened 2022-10-18
ScrollTo Selected Cell in CardStrip

evancrow:
Adds `scrollTo` support in the `CardStripView` similar to how it works in `TabGridContainer`.

### How was this tested?
- [x] I tested this manually

https://user-images.githubusercontent.com/37356984/196314278-40582a91-9e7a-46ec-a980-24d692e6de0c.mp4


evancrow:
It's called whenever the CardGrid scrollTo is called, which is when it's opened _and_ closed. Would be nice to only update on close though, will make that change!

Client/Frontend/Browser/CardGrid/Strip/CardStripView.swift
```diff
@@ -44,10 +44,13 @@ struct CardStripView: View {
...
+                        .id(cell.id)
                 case .tabGroupGridRow(let groupDetails, _):
                     CardStripTabGroupCardView(groupDetails: groupDetails)
+                        .id(cell.id)
                 case .tab(let tabDetails):
                     CardStripCard(details: tabDetails)
+                        .id(cell.id)
                 default:
                     EmptyView()
                 }
```
EdwardJTL, line 56: can the `.id` modifier be applied to the entire block instead of inside each `switch case`? It could cause problems with this case
```
case .all:
  // No id for all.
  return ""
```

EdwardJTL (approved):

PR #4516, opened 2022-10-18
Use NormalDetails in DetailCount [CardStrip]

evancrow:
This fixes a bug where the card strip would show even if there only was one tab because there were incognito tabs included in `allDetails`. `detailCount` should only look at `normalDetails` in normal mode.

### How was this tested?
- [x] I tested this manually

Marking this as cherry-pick-needed since it would be nice to have this fixed this week.


chung-lai (approved):
LGTM

PR #4514, opened 2022-10-18
premium; avg monthly price i18n

jedireza:
This PR attempts to address a bug when we're displaying the average price per month for the annual plan.

In all other places we display prices, we're using `product.displayPrice` which includes formatting appropriate for the user's App Store local.

When we display the average price per month for the annual plan, we use a `NumberFormatter`. The display styles available from a number formatter can be found [here](https://developer.apple.com/documentation/foundation/numberformatter/style). Note: none of the formatting options will match the way `product.displayPrice` is formatted exactly 😕

For example if the App Store country is USA and the device region is set to UK, we divide the USD price by 12 and the `NumberFormatter` displays that value in Euros. This isn't just bad formatting, it's incorrect.

All is good when the app store country and region match.

As-is this PR isn't complete as it only handles the case where the App Store country is USA but the region isn't. I was really hoping we didn't need to maintain a bunch of custom formatting rules.

Maybe I missed another API available that would resolve this automatically. I also don't know how common it is for a users's App Store country to be different than their device region.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator.

### Screenshots and screen recordings
App Store United States with language and region settings changed.
image

## Issues addressed
N/A

PR #4519, opened 2022-10-18
premium; better avg monthly price i18n

jedireza:
In #4514 we used a semi-documented API (`product.priceFormatStyle.locale`) which ended up causing a crash on iOS 15 devices.

I had another idea which I think results in even more consistent results. This PR updates the `PremiumHelpers.annualAvgPricePerMonth` with a string replacement approach that results in a format the same as `product.displayPrice`.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator; English + United States, German + Germany, French + France

image

## Issues addressed
N/A


Client/Frontend/Premium/PremiumStore.swift
```diff
@@ -365,9 +365,23 @@ class PremiumHelpers {
     }
 
     static func annualAvgPricePerMonth(_ product: Product) -> String {
+        let pricePerMonth = product.price / 12
+        var output = ""
```
chung-lai, line 369: It's very unlikely and just nice to have. Should we have a fall back value for the `output` in the case formatter returns nothing?

chung-lai (approved):
that's awesome solution :) Ship it!

Client/Frontend/Premium/PremiumStore.swift
jedireza: I was thinking about that too. But we need to handle better at the call sites actually cause they just use this value in view code. Open to suggestions.

PR #4522, opened 2022-10-19
premium; graceful display when avg monthly price output is empty

jedireza:
Follow up from #4519. In this PR we handle empty output from `PremiumHelpers.annualAvgPricePerMonth` at the call sites.

In #4519, when testing with the French language selected, I saw that `NumberFormatter` was appending extra whitespace that didn't exist in the`product.displayPrice` formatting. The visual clue that something was wrong was that the average monthly price was the yearly price since the string replacements didn't change anything. In this PR I also change the output to be empty if the output we produce is the same as the original display price. If that happens, something went wrong with the formatting.

I also changed the `PremiumHelpers.annualAvgPricePerMonth` function to return a `String?`. This reduced an extra condition check at the call sites.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator. However I don't know of a case where this would actually be incorrect.

N/A

## Issues addressed
N/A


chung-lai (approved):
LGTM

PR #4527, opened 2022-10-20
welcome flow; allow content to scroll

jedireza:
This PR updates the welcome flow screens to be scrollable. Some screens like the plans page is very text heavy and when translated ends up clipping too much content.

### How was this tested?
- [X] I tested this manually

### Manual test cases
Simulator.

image

## Issues addressed
N/A


chung-lai (approved):
Changes look good. Only thing we may want to follow up or consider is things squeeze together in the welcome screen after the change:

| Before | After |
| --- | --- |
| Screen Shot 2022-10-19 at 9 01 54 PM | Screen Shot 2022-10-19 at 9 01 22 PM |

jedireza:
> Changes look good. Only thing we may want to follow up or consider is things squeeze together in the welcome screen after the change:

Yeah. I sunk a good amount of time into using a `GeometryReader ` in order to allow the main screen content to fill it's available height. My attempts were inspired by the answers here: https://stackoverflow.com/q/62877863 -- I thought I was making some headway and then noticed with that approach, the content started clipping again, which was the reason to make this change.

chung-lai:
> > Changes look good. Only thing we may want to follow up or consider is things squeeze together in the welcome screen after the change:
> 
> Yeah. I sunk a good amount of time into using a `GeometryReader` in order to allow the main screen content to fill it's available height, while having the content be scrollable. My attempts were inspired by the answers here: https://stackoverflow.com/q/62877863 -- I thought I was making some headway and then noticed with that approach, the content started clipping again, which was the reason to make this change.

Right, guess will need GeometryReader. Scroll view with spacer is tricky. What we have here is better than clipping. We can just follow up on this.

PR #4530, opened 2022-10-20
settings; premium availability check update

jedireza:
We had a report from a user where they saw the premium screen and then upgraded to iOS 16 and when they went back the menu item for premium was gone. I'm not sure if this will help resolve that, but it may be worth a try.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator.

N/A

## Issues addressed
N/A


chung-lai (approved):

jedireza:
Thanks for reviewing. I'm going to hold off on this for now. I'll probably create a new PR to log when we don't show premium.

PR #4531, opened 2022-10-21
premium; new sign in or up flow

jedireza:
This PR replaces the old `IntroView` overlay for two specific call sites; under settings and on the zero query screen. If these go well, we can remove any other remaining call sites and also retire that old code.

While updating this flow I also found some improvements relevant to the welcome flow as well.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator; welcome flow, settings, zero query

https://user-images.githubusercontent.com/979929/197091957-ff545564-6a6f-4545-8da7-60a2077260ca.mp4

## Issues addressed
N/A


jedireza:
I'm working on the failing tests today.

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
```diff
@@ -7,17 +7,19 @@ import SwiftUI
...
-        ("Premium password manager + VPN", ""),
+        ("Unlimited devices", ""),
+        ("Password manager + VPN", ""),
     ]
     var freeBullets = [
-        ("Browser + ad blocker", ""),
-        ("Tracking prevention", ""),
-        ("Ad-free, private search", "50 searches/week"),
+        ("Browser + ad blocker + tracking prevention", ""),
+        ("Limited Ad-free, private search", "50 searches/month"),
```
chung-lai, line 21: Assume these are strings from the welcome flow so we don't need to add them for translation, right?

chung-lai (commented):
LGTM. Approve after test are fixed. 🙏

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
jedireza: That’s right.

chung-lai (approved):
Thanks for fixing the test!

PR #4538, opened 2022-10-26
premium; fix pending state handling

jedireza:
This PR updates the handling of "pending" purchase in our premium checkout flows. The idea is that we at least change the screen if this happens instead of doing nothing.

We had a report where a user said they purchased premium but the screen didn't do anything and I wonder if this pending state is what caused that.

In any case, this is a good change since before it was a noop code path. And we have had more of these states arise after enabling the free trial period.

image

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator. Although it's unknown to me how to trigger this state.

N/A

## Issues addressed
N/A


chung-lai (approved):
LGTM.

PR #4543, opened 2022-10-28
Encode suggested URL for i18n

chung-lai:
Fast tap URL may not be encoded to be url query friendly. We should encode that.

https://user-images.githubusercontent.com/87332578/198692215-e456d4dd-975e-4cab-8501-3f33b33cf71a.mov

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


jedireza (approved):
Confirmed this fixes the bug we identified.

PR #4549, opened 2022-10-31
Fix SafariExtension Width

evancrow:
Needed to set a min-width for iPad devices where the HTML isn't automatically set to fill the space.

### How was this tested?
- [x] I tested this manually

Before:
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-10-31 at 09 27 55

After:
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-10-31 at 09 32 26


NeevaForSafari/NeevaForSafari/ViewController.swift
```diff
@@ -36,7 +36,7 @@ class ViewController: UIViewController, WKNavigationDelegate, WKScriptMessageHan
         // Override point for customization.
     }
 
-    func userContentController(didReceive message: WKScriptMessage) {
+    func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
```
jedireza, line 39: Is this change necessary? We have a different function signature, but still a noop.

evancrow, line 39: BTW, this change was caused by the protocol being updated with the new version of Swift. I didn't catch the build failure until I ran NeevaForSafari, so all should be good now.

evancrow: Just left a comment as you asked :) (see below)

PR #4552, opened 2022-11-01
Fix URL Check Cheatsheet View Evaluation

EdwardJTL:
in #4169, we delayed when fetch happens. This caused the view to first evaluate to show intro view on SRP pages, which sets `seenCheatsheetIntro` to `true` immediately in its `onDisappear`. This breaks the happy path in the onboarding flow.

this fix allows the view to read the tab's url before fetch.
also added test for the happy path

### How was this tested?
- [ ] I tested this manually
- [x] I added automated tests


chung-lai (approved):
thank you! LGTM

PR #4574, opened 2022-11-14
Fix space thumbnail loading error in thumbnail group

EdwardJTL:
The same code path for getting a thumbnail view works in space details list but not in the card grid.
This behaviours appears to be in line with the documentation of the expected behaviour for `drawingGroup`

we should try to keep using the latest SDWebImageSwiftUI since it has fixes for 2.1.0

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


jedireza (approved):
Thanks!

PR #4576, opened 2022-11-15
Silence access environment value layout direction warning

EdwardJTL:
this warning is very spammy and slows down my xcode with runtime warnings.
It appears that we are allowed to read this value inside a `ViewModifier` without triggering a warning

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests


evancrow (approved):

evancrow (approved):
Still LGTM!

PR #4577, opened 2022-11-16
Update Scene Lifecycle Methods

EdwardJTL:
For referencer, documentation on what scene lifecycle delegates should do:
https://developer.apple.com/documentation/uikit/app_and_environment/scenes
https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_foreground
https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background

This PR implements `sceneWillResignActive`
This PR moves calls to lifecycle methods to appropriate delegates following the recommendation of the docs linked above.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
on ipad, with stage manager, with browser and storage logging enabled
create multiple stages with the help of another app, say, photos, create at least once instance of neeva in each stage. 
**tab persistence test**
1) Open tabs in one instance. verify log that tab is saved. switch to an instance in another stage, open a tab, and verify log that tab is saved correctly. 
3) exit to desktop. come back, repeat
4) force close the app from xcode (don't swipe away from the switcher). relaunch the app, verify that each instance still has its tabs.
5) verify that the number of active scenes logged doesnt not exceed the count you expect.

**reader mode test**
 enter reader mode in on instance. switch to another stage, enter reader mode in an instance on that stage. Go back to the first instance, verify the page still works

** Open in background extension test **
open a url in safari. Go to share>Neeva>open in background
switch to neeva via dock, verify that an instance loads the tab


Client/AppClipHelper.swift
```diff
@@ -7,7 +7,7 @@ import Foundation
 class AppClipHelper {
     static let appClipGroupId = "group.co.neeva.app.ios.browser.app-clip.login"
 
-    static func retreiveAppClipData() -> String? {
+    static func retrieveAppClipData() -> String? {
```
chung-lai, line 10: i'm very curious on what's changed on this line lol

EdwardJTL: typo haha

Client/Application/SceneDelegate.swift
```diff
@@ -108,10 +88,36 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             callSite: LocalNotifications.ScheduleCallSite.enterForeground
         )
 
+        // If server is already running this won't do anything.
+        // This will restart the server if it was stopped in `sceneDidEnterBackground`.
+        if !WebServer.sharedInstance.isRunning {
```
chung-lai, line 93: Curious, does this change fix a particular bug that we know of?

Client/AppClipHelper.swift
evancrow: lol that was me :|

chung-lai (commented):
Change looks good to me. Curious what's the motivation behind making this change now. Does this fix a particular bug or it's prep work for some upcoming changes?

@evancrow also heavily involve in the SceneDelegate, would love his review :)

chung-lai:
May also worth checking if we set the app as default browser and test the deep link flow to make sure the enter foreground flow is still working. Pretty sure it follows the same lifecycle as user bringing an app in foreground.

evancrow (approved):
Yeah, this is just moving stuff around to the proper method. I'm cool with this. Agree with @chung-lai that deep-link, default browser, and app clip methods should all be tested.

Client/Application/SceneDelegate.swift
EdwardJTL: this particular line? no

EdwardJTL:
> Change looks good to me. Curious what's the motivation behind making this change now. Does this fix a particular bug or it's prep work for some upcoming changes?
> 
> @evancrow also heavily involve in the SceneDelegate, would love his review :)

thanks for tagging evan :D

yes, there are two motivations
1) there is a bug in the code where we keep track of active scene sessions, and this PR fixes that counter. I am not sure if we necessarily need to keep track of this ourselves, because there are system APIs for querying that count.
2) I need this as a dependency for my tab storage changes since that one also needs to interact with scene lifecycles.

chung-lai:
Looks like test build failed, we should fix that before merging.

PR #4581, opened 2022-11-18
settings; premium screen; add trial copy under button

jedireza:
This PR adds the "7 Day Free Trial" copy under the CTA button in the settings screen.

I also added the intro offer config to align the local testing config with the app store config.

### How was this tested?
- [x] I tested this manually

### Manual test cases
Simulator.

image

## Issues addressed
🔒 https://app.clickup.com/t/3vqrzc6


Client/Frontend/Settings/Neeva/NeevaPremiumView.swift
```diff
@@ -217,6 +217,13 @@ struct NeevaPremiumView: View {
                     }
                 )
                 .buttonStyle(.neeva(.primary))
+
+                if let subText = PremiumHelpers.primaryActionSubText(currentPremiumPlan),
+                    subText != ""
```
evancrow, line 222: I would use `!subText.isEmpty` instead

evancrow (approved):
One thing to fix before merging, but rest LGTM

PR #4588, opened 2022-12-01
Retire application cleanly backgrounded

EdwardJTL:
deprecate the defaults flag
clean up the logging method

Can't test this since PLCrashReporter doesn't work with debugger attached. https://bytemeta.vip/repo/microsoft/plcrashreporter/issues/220


chung-lai (approved):
Thanks for cleaning this up! @macy-n, i don't think there is any current dashboard relying on `AppCrashWithPageLoad`, right?

EdwardJTL:
> Thanks for cleaning this up! @macy-n, i don't think there is any current dashboard relying on `AppCrashWithPageLoad`, right?

https://app.mode.com/neeva/reports/38c0fb873cdc/runs/04d4a5d1055a
this still uses that counter path

chung-lai:
> > Thanks for cleaning this up! @macy-n, i don't think there is any current dashboard relying on `AppCrashWithPageLoad`, right?
> 
> https://app.mode.com/neeva/reports/38c0fb873cdc/runs/04d4a5d1055a this still uses that counter path

Oh interesting, think we can simply update that to use the CrashReporter counter instead given it also have the pageLoaded attribute. (Having said that, don't think we have that attribute in our allow list anymore :( )

chung-lai:
> > > Thanks for cleaning this up! @macy-n, i don't think there is any current dashboard relying on `AppCrashWithPageLoad`, right?
> > 
> > 
> > https://app.mode.com/neeva/reports/38c0fb873cdc/runs/04d4a5d1055a this still uses that counter path
> 
> Oh interesting, think we can simply update that to use the CrashReporter counter instead given it also have the pageLoaded attribute. (Having said that, don't think we have that attribute anymore with MMO)

Lol, turns out we reset `pageLoaded` after sending `AppCrashWithPageLoad` meaning `AppCrashWithCrashReporter` won't have the `pageLoaded` properly. HOWEVER, with this PR, we removed `AppCrashWithPageLoad` so it also fix that, we should be good :)

PR #4589, opened 2022-12-02
sign in/up; fix overlay auto closing

jedireza:
This PR attempts to fix the sign in/up overlay automatically closing when it's opened.

### How was this tested?
- [x] I tested this manually
- [x] I added automated tests
    - I added a simple a test unrelated to this fix. I want to add more tests around different states of `NeevaUserInfo`, but in my attempts to call `NeevaUserInfo.shared.setLoginCookie()`, an error is thrown related the bundle id not having access to the Neeva API. I also tried to set the cookie value via `NeevaConstants.keychain.set(...)` but that also throws. There are other API calls that would get fired when interacting with these methods, so we may need to abstract states into the view model or do some other mocking of `NeevaUserInfo`.

### Manual test cases
Tested in the simulator. Signing up and signing in.

N/A

## Issues addressed
🔒 https://app.clickup.com/t/32ceu9a


Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
```diff
@@ -308,7 +312,14 @@ struct SignInOrUpFlowPlansView: View {
                         model.complete()
                     }
                 } else {
-                    model.complete()
```
jedireza, line 311: What I think may be happening is the user might still have a cookie value (valid or not) and we were simply calling `model.complete()` in this arm where the overlay gets closed.

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
```diff
@@ -213,8 +213,12 @@ struct SignInOrUpFlowPlansView: View {
                         )
 
                         if model.currentPremiumPlan == nil {
-                            model.prevScreens.append(.plans)
-                            model.changeScreenTo(.signUp)
+                            if model.justSignedIn {
+                                model.complete()
```
jedireza, line 217: This makes for a better experience where the user might have considered premium, signed in/up then when returning to the plans page they change their mind. If they click the `Get FREE` button the overlay will close and they'll be on their way. Before they'd get sent back to sign up, creating a loop and requiring them to hit the `X` button.

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
```diff
@@ -308,7 +312,14 @@ struct SignInOrUpFlowPlansView: View {
...
-                    model.complete()
+                    if model.justSignedIn {
+                        model.complete()
+                    }
+                }
+            } else {
+                if PremiumStore.shared.products.count == 0 {
+                    // No product found, ensure the free plan is selected. By default annual is assumed.
+                    model.currentPremiumPlan = nil
                 }
```
jedireza, line 323: Since we prime the products on the previous screen, they _should_ be populated by now. If they're not and the only plan that shows is free, we'd have an incorrect current product (defaults to annual). This cleans that up.

EdwardJTL:
we should ship this and verify with users (and me) that this fixes their sign in auto-dismiss error.
but we should have more robust tests and code structure to make sure this doesnt happen again. This bug and PR make feel that there are some fragile pieces of interaction patterns, code designs, and concurrency management.

EdwardJTL (approved):

jedireza:
> but we should have more robust tests and code structure to make sure this doesnt happen again
pretty much my aim 

Did you read the part of the PR description about tests? We have active work effort to add tests around these code paths. Many code branches depend on `NeevaUserInfo` state and other GQL API calls also get fired. These either need to be abstracted into the view model or mocked at a higher level. Are there best practices for testing around this type of app state already?

>  This bug and PR make feel that there are some fragile pieces of interaction patterns, code designs, and concurrency management.

Code reviews and suggestions are welcome. 🙏

EdwardJTL:
we need to have tests around NeevaUserInfo to make sure that it behaves properly when the cookie is invalid, and we may not be able to do that until we can mock the networking interface.
> These either need to be abstracted into the view model or mocked at a higher level
I'm not sure what you mean here. Do you mean our SwiftUI Views should not be interacting with UserInfo directly? 
>  Are there best practices for testing around this type of app state already?
one way to do this is do create mocked user info objects and inject those in the testing flows. Like how we use BrowserProfile most of the time but MockProfile in some tests, while both conform to Profile protocol.

I will leave some comments on the code, but they won't be on the changes in this PR.

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowModel.swift
```diff
@@ -22,6 +22,7 @@ class SignInOrUpFlowModel: ObservableObject {
     @Published var showCloseButton = true
     @Published var showAllSignUpOptions = false
     @Published var currentPremiumPlan: PremiumPlan? = .annual  // nil is the free plan
+    var justSignedIn = false  // set back to false inside `self.complete`
```
EdwardJTL, line 25: since views read this value, its probably better to have this as published so SwiftUI doesn't miss any updates

EdwardJTL, line 25: this comment should also explain why we need this var, like you did in the PR comments

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
```diff
@@ -213,8 +213,12 @@ struct SignInOrUpFlowPlansView: View {
                         )
 
                         if model.currentPremiumPlan == nil {
```
EdwardJTL, line 215: I'd recommend making this a function in the model so that the view code is clean and the model owns the behaviour. the function name will also be a good hint to the reader as to what this button does

jedireza:
> I'm not sure what you mean here. Do you mean our SwiftUI Views should not be interacting with UserInfo directly?

That's one way that might work. I don't think it's the best route.

jedireza:
> one way to do this is do create mocked user info objects and inject those in the testing flows. Like how we use BrowserProfile most of the time but MockProfile in some tests, while both conform to Profile protocol.

That's helpful it gives me something to reference and investigate. Given that there are no existing references to testing around `NeevaUserInfo`, maybe we should land this PR sooner and follow up with test after.

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
```diff
@@ -308,7 +312,14 @@ struct SignInOrUpFlowPlansView: View {
                         model.complete()
```
EdwardJTL, line 312: I'm not sure why the purchase is triggered in an onAppear modifier. maybe a comment explaining how one may arrive in this code path?

it might also be helpful to declare a second `View` that uses the UI declarations in this View, but has this modifier attached. This can help provide some clarity on code flow.

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
```diff
@@ -308,7 +312,14 @@ struct SignInOrUpFlowPlansView: View {
...
                     }
                 } else {
-                    model.complete()
+                    if model.justSignedIn {
+                        model.complete()
```
EdwardJTL, line 316: I'm hoping this fixes it. But after this patch, I am of the opinion that we should not be blocking users from signing in even if NeevaUseInfo has bad data, this is also why I felt the need to understand why we have an onAppear that can dismiss itself.

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
EdwardJTL: I looked at the logic for populating the stores, I have some suggestions but alas I can't comment on them in the PR format. I can make a draftPR with code comments if you think that'd be helpful

EdwardJTL (commented):
left some comments on things

EdwardJTL:
Do we have documentation on the intended sign in or sign up flow?

Client/Frontend/SignInOrUpFlow/SignInOrUpFlowModel.swift
jedireza: This is only read in `onAppear`, so publishing isn't necessary imo.

jedireza:
> Do we have documentation on the intended sign in or sign up flow?

🔒 [Figma for iOS Premium](https://www.figma.com/file/NMz44MFSvyEDPQc0NtxIKC/Lilac---Mobile%3A-Day-0-Conversion?node-id=0%3A1&t=vAqW6b0V5503Eykh-0), [Figma for Android Premium](https://www.figma.com/file/dylvDPB69JG9U1rzZ3HQ52/Android-Premium?node-id=0%3A1&t=mBu1WfAlx5AJdTJk-0)

PR #4593, opened 2022-12-06
sign in/up; fix overlay auto closing (round 2)

jedireza:
Follow up to the previous PR #4589

This PR ensures there are no longer any lingering cookies when we present the sign in/up overlay. It also updates how the settings panel checks if we should show the "Sign in or Join Neeva" button item vs the account item.

### How was this tested?
- [x] I tested this manually

### Manual test cases

I tested in the simulator.

#### Steps to reproduce the bug on `main`
https://user-images.githubusercontent.com/979929/206000598-bac7f1b1-eb76-4cb6-9ff2-89875939fb52.mp4

#### Same steps using this branch
https://user-images.githubusercontent.com/979929/206000676-e9159133-79db-4e27-aaf3-8c7a18f6797b.mp4

## Issues addressed

🔒 https://app.clickup.com/t/32ceu9a


Client/Frontend/SignInOrUpFlow/SignInOrUpFlowPlansView.swift
```diff
@@ -309,7 +309,9 @@ struct SignInOrUpFlowPlansView: View {
                             onError: self.onPurchaseError,
                             onSuccess: self.onPurchaseSuccess)
                     } else {
-                        model.complete()
```
evancrow, line 312: Can simplify this to `else if  model.justSignedIn`, so no need for a nested if.

evancrow (approved):
LGTM with one nit

Also, does this fix the Neeva Scope bug I encountered, or is that separate from this?

jedireza:
> Also, does this fix the Neeva Scope bug I encountered, or is that separate from this?

I'm not familiar. I'll follow up on a DM.

Client/Frontend/Settings/Neeva/NeevaSettingsSection.swift
```diff
@@ -47,7 +47,7 @@ struct NeevaSettingsSection: View {
             }.onDisappear {
                 loaderOpacity = 0
             }
-        } else if userInfo.isUserLoggedIn && Defaults[.signedInOnce] && userInfo.email != nil
+        } else if userInfo.hasLoginCookie() && Defaults[.signedInOnce] && userInfo.email != nil
```
EdwardJTL, line 50: can you leave a todo here
`isUserLoggedIn` is published and this method is not. it will lead to worse performance and may miss state updates. we should fix that

EdwardJTL: i guess this todo can reference the todo you left on `isUserLoggedIn` lol

EdwardJTL:
> LGTM with one nit
> 
> Also, does this fix the Neeva Scope bug I encountered, or is that separate from this?

that's separate. This may fix it since it will allow you to sign in and get a new cookie, but it still remains unsolved why neeva user info has inconsistent states and is holding onto bad cookie

Client/Frontend/Settings/Neeva/NeevaSettingsSection.swift
jedireza: We have an issue to line this up throughout here: https://app.clickup.com/t/3xbbca4

PR #4594, opened 2022-12-06
sign in/up; better handling when user signs in through sign up

jedireza:
This PR adds better handling when an existing member "signs in" through the "sign up" screens.

This also fixes a case where we execute a purchase for a user who is already premium.

### How was this tested?
- [x] I tested this manually

### Manual test cases
I tested in the simulator.

N/A

## Issues addressed
🔒 https://app.clickup.com/t/3xbd6mz


evancrow (approved):
Tried it out on my devices, works well, code also LGTM!

PR #4598, opened 2022-12-09
Debug Tool - Export and Load Browser Profile

EdwardJTL:
This will allow us to export and re-import tab states for debugging and migration testing, and specify a real app delegate with different browser profiles in testing

- Add defaults variable and update BrowserProfile and TabManagerStore to use this variable
- Add debug views to export, browse, select, delete, and import browser profiles

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
See video. Browsing current profile dir, exporting it and saving it to User docs, browsing available profiles, importing and selecting profile.

https://user-images.githubusercontent.com/44557675/206801484-e4736e8f-01e4-4ca1-9cdd-dda392bf11ac.MP4


Providers/Profile.swift
```diff
@@ -134,6 +134,8 @@ open class BrowserProfile: Profile {
             }
         }
 
+        print(files.listRootPath())
```
chung-lai, line 137: change this to log instead of print?

Client/Frontend/Browser/Tabs/TabManagerStore.swift
```diff
@@ -25,19 +25,11 @@ class TabManagerStore {
...
-        let profilePath: String?
-
         if AppConstants.IsRunningTest || AppConstants.IsRunningPerfTest {
-            profilePath = (UIApplication.shared.delegate as? TestAppDelegate)?.dirForTestProfile
+            return (UIApplication.shared.delegate as? TestAppDelegate)?.dirForTestProfile
         } else {
-            profilePath =
-                fileManager.containerURL(
-                    forSecurityApplicationGroupIdentifier: AppInfo.sharedContainerIdentifier)?
-                .appendingPathComponent("profile.profile").path
```
chung-lai, line 36: are we changing the path name which contains `profile.profile`? Wonder how app upgrade will behave

Client/Frontend/Settings/Debug/TabStorageSettingsView.swift
```diff
@@ -0,0 +1,526 @@
...
+    }
+
+    var body: some View {
+        List {
+            Section("Warning") {
+                HStack(alignment: .top) {
+                    Image(systemSymbol: .exclamationmarkTriangle)
+                        .foregroundColor(.yellow)
+                    VStack(alignment: .leading) {
+                        Text("Intended for debug use only")
```
chung-lai, line 46: Let's use `Text(verbatim:)` for these strings as we don't need to localize the debug settings

Client/Frontend/Browser/Tabs/TabManagerStore.swift
EdwardJTL: AppDelegate.profile.files use the exact same routine to compute the path, so i think this will be fine. we can see how it goes on testflight

Providers/Profile.swift
EdwardJTL: this wasn't supposed to be included in the change my b

Client/Frontend/Settings/Debug/TabStorageSettingsView.swift
EdwardJTL: okay i think i got all of them. it picks up things from `Section`, `Button`, `LabeledContent` initializer and `.navigationTitle` as well. it was a good exercise on interpreting swift overload signatures

chung-lai (approved):
lGTM.

PR #4608, opened 2022-12-20
Fix Delete Single ArchivedTab Card

evancrow:
Needed to reload the data after deleting a single card, or else the view wouldn't refresh until the app reloaded. Fixed a similar issue with the debug method.

### How was this tested?
- [x] I tested this manually


jedireza:
Are there any low friction test cases we could add?

Client/Frontend/Browser/History+ArchivedTabs/ArchivedTabs/ArchivedTabsCardView.swift
```diff
@@ -59,12 +60,14 @@ struct ArchivedTabsCardView: View {
             if #available(iOS 15.0, *) {
                 Button(role: .destructive) {
                     tabManager.remove(archivedTabs: [tab])
+                    model.loadData()
```
EdwardJTL, line 63: would it be possible to make the model listen to tab manager when it makes those updates? I think it will be more robust this way and less likely for us to forget to refresh in `View`s in the future.

evancrow: We could, but then we'd have to make a subscription to `ArchivedTabs`, which was what we're trying to avoid, right?

Maybe instead, it's better to have one update method in the model that calls the `TabManager` function(s) and then calls `loadData`.

EdwardJTL (approved):

PR #4609, opened 2022-12-21
fix strings for unlimited subscription type

jedireza:
I learned today that we're not translating the names of the plans.
🔒 https://neevaco.slack.com/archives/C03C9RT13KJ/p1671646929498429?thread_ts=1671646755.897589&cid=C03C9RT13KJ


chung-lai (approved):
LGTM, again, these strings will get translated in next batch.

PR #4613, opened 2023-01-05
Add main thread guards to NeevaUserInfo login cookie methods

EdwardJTL:
add an assert to help triage race conditions.

### How was this tested?
- [ ] I tested this manually
- [ ] I added automated tests

### Manual test cases
sign in (apple and google), sign out, open zero query, open settings


Shared/API/NeevaUserInfo.swift
```diff
@@ -126,6 +126,7 @@ public class NeevaUserInfo: ObservableObject {
     }
 
     public func setLoginCookie(_ value: String) {
+        assert(Thread.isMainThread)
```
EdwardJTL, line 129: if we want this to fail in release, we can use `precondition`

EdwardJTL:
no crashes when i played around with it. if we want to catch non-synchronized access in release builds, we'll have to make the app crash

chung-lai (approved):
The assertions look good.

PR #4617, opened 2023-01-11
Prevent Archive Incognito Tabs

evancrow:
Noticed that incognito tabs started getting archived which is something we decided shouldn't happen. This fixes the bug by adding an `!incognito` check to `shouldBeArchived`. Also adds the same check to the debug code.

### How was this tested?
- [x] I tested this manually


evancrow:
Adding a ClientTest for this.

jedireza (approved):
Code looks good. Thanks for adding the test!

PR #4625, opened 2023-02-09
Archive Instead of Close Tabs

evancrow:
(Behind flag)

Instead of closing tabs, archives them (they can be fully deleted from the archive). This also suppresses the Restore Tab toast.

Planning on adding tests, might wait for the flag flip PR to do that.

### How was this tested?
- [x] I tested this manually

Related tickets 🔒:
https://app.clickup.com/t/3w3c3uq


evancrow:
Couple of things left to think through:
1. Rebrand "Recently Closed Tabs" to "Recently Archived Tabs"
2. Support unarchiving "Recently Archived Tabs"

Client/Frontend/Browser/Tabs/TabManagerStore.swift
```diff
@@ -166,7 +167,14 @@ class TabManagerStore {
         restoredTabs.reserveCapacity(savedTabs.count)  // Overestimating
 
         for savedTab in savedTabs {
-            if savedTab.shouldBeArchived {
+            // savedTab.shouldBeArchived
+            let shouldBeArchived =
```
EdwardJTL, line 171: it's a bit difficult to read `&&` followed by `||`. Perhaps we can have a (fileprivate) method that returns a bool, so that we can something more like
```
if (cond) {return true} 
else if (cond) { return true } 
return false
```
etc

Client/Frontend/Browser/Tabs/TabManager.swift
```diff
@@ -1146,8 +1150,17 @@ class TabManager: NSObject, TabEventHandler, WKNavigationDelegate {
...
+            )
+
+            if FeatureFlag[.archivedDontCloseTabs] {
+                archivedTabs.append(.init(savedTab: savedTab))
+            }
+
+            return savedTab
         }
 
         if recentlyClosedTabGroupTimer?.isValid ?? false, recentlyClosedTabs.count > 0 {
```
EdwardJTL, line 1166: does this still trigger the timer?

EdwardJTL (commented):
i will test drive this later, just some comments for now from just looking at the changes

Client/Frontend/Browser/Tabs/TabManager.swift
evancrow: Nope :)

evancrow:
Lot's of tests will need to be fixed, heads up

EdwardJTL (approved):
might be nice to have user education. In case people are wondering "how can i restore a tab i just closed" when there are no toasts.

evancrow:
Had an idea, use a toast to direct the user to archive:
_Toast will not auto-dismiss FYI, the user has to hit the button or swipe it away._



https://user-images.githubusercontent.com/37356984/221320565-7cdd5678-6d01-461f-9d71-70be789b2713.mp4

@Ria-yu Thoughts on this? I can also add it to the design sync for discussion.

Ria-yu:
One-time non dismissible toast sounds like a reasonable fix. I'd suggest including the full name of Archived Tabs in copy to be clearer, like "Closed tabs can be found in [icon] Archived Tabs" (skip icon if it can't show in there)

evancrow:
Done!

PR #4626, opened 2023-02-10
Sign-In Improvments

evancrow:
My changes:
- Only check the signed-in state through one variable `isUserLoggedIn`
- `entitledToPremiumFeatures() -> Bool?` returns nil if data is loading
- One logout function, `logOutUser`, which clears the login cookie and userInfo cache
- General code cleanup

Fixes:
- Premium promo showing when it shouldn't
- Inconsistent use of `getLoginCookie` and `isUserLoggedIn`
- Forced delays in `Fetch`

Tested all login types. The state is persistent, and offline works as expected. Logout clears all data and signing back in works as well.

### How was this tested?
- [x] I tested this manually
- [x] Existing automated tests are enough to cover my changes (will add these)

## Issues addressed
Clickup 🔒: https://app.clickup.com/t/3xbbca4


Shared/API/NeevaUserInfo.swift
```diff
@@ -258,5 +235,42 @@ public class NeevaUserInfo: ObservableObject {
         authProvider = nil
         subscriptionType = nil
         self.reachability.stopNotifier()
+
+        // TODO: Call `saveUserInfoToDefaults` here?
```
evancrow, line 239: Any thoughts on this?

evancrow:
Tests for Spaces are failing because we default login to false now... will likely have to create an exception for tests.

Shared/API/NeevaUserInfo.swift
```diff
@@ -20,53 +21,34 @@ public class NeevaUserInfo: ObservableObject {
...
-        self.displayName = displayName
-        self.email = email
-        self.pictureUrl = pictureUrl
-        self.authProvider = (authProvider?.rawValue).flatMap(SSOProvider.init(rawValue:))
-        isUserLoggedIn = true
-        fetchUserPicture()
-    }
-
     public static let previewLoggedOut = NeevaUserInfo(previewLoggedOut: ())
     public static let previewLoading = NeevaUserInfo(previewLoading: ())
```
EdwardJTL, line 33: nit: can you move this so all the `public static let` singletons are declared next to each other? unrelated to your change, but might as well clean it up. thanks!

Shared/API/NeevaUserInfo.swift
```diff
@@ -20,53 +21,34 @@ public class NeevaUserInfo: ObservableObject {
...
-    private init(previewLoggedOut: ()) {
-        isUserLoggedIn = false
-    }
-    private init(previewLoading: ()) {
-        isLoading = true
+
+    // MARK: - Update Login State
+    public func updateLoginState() {
```
EdwardJTL, line 36: i'd suggest renaming this to something like `updateLoginStateIfCookieExists` to make it clear that this operation only checks states on the device, and that it is not making a network call.

EdwardJTL, line 36: but in general, why does this method need to be public?

Shared/API/NeevaUserInfo.swift
```diff
@@ -76,27 +58,25 @@ public class NeevaUserInfo: ObservableObject {
 
         isLoading = true
```
EdwardJTL, line 59: should this function have a check for if loading is already true?

Shared/API/NeevaUserInfo.swift
```diff
@@ -76,27 +58,25 @@ public class NeevaUserInfo: ObservableObject {
...
+                // Once we've fetched UserInfo sucessfully, we don't need to keep monitoring connectivity anymore.
+                self.reachability.stopNotifier()
+            case .failureAuthenticationError:
+                self.clearUserInfoCache()
+                self.loadUserInfoFromDefaults()
+            case .failureTemporaryError:
+                self.loadUserInfoFromDefaults()
             }
+
+            self.updateLoginState()
```
EdwardJTL, line 79: why do we need to do this here? does the cookie get mutated in this routine?

Shared/API/NeevaUserInfo.swift
```diff
@@ -106,16 +86,22 @@ public class NeevaUserInfo: ObservableObject {
         }
     }
 
-    public func didLogOut() {
-        clearCache()
-        isUserLoggedIn = false
-        fetch()
+    private var isDeviceOnline: Bool {
+        if let connection = connection, connection != .unavailable {
```
EdwardJTL, line 90:
```
guard let connection = connection else { return false}
return connection == .available
```

Shared/API/NeevaUserInfo.swift
```diff
@@ -143,19 +128,16 @@ public class NeevaUserInfo: ObservableObject {
 
         // Remove preview cookie from keychain
         try? NeevaConstants.removePreviewCookie()
+        updateLoginState()
```
EdwardJTL, line 131: could you explain why we need to do this here? 
Reading just this function along, it appears to me that we would only need to do this if the login cookie is different than what's already in keychain

Shared/API/NeevaUserInfo.swift
EdwardJTL: yes let's remove the value from defaults

EdwardJTL: the function name is weird too. `Cache` here is not clear if it is referring to the in-memory values on the stack, or the values stored in the file system via defaults.
we can rename it `clearUserInfo` and just remove both

Shared/API/NeevaUserInfo.swift
```diff
@@ -258,5 +235,53 @@ public class NeevaUserInfo: ObservableObject {
         authProvider = nil
         subscriptionType = nil
         self.reachability.stopNotifier()
+
+        // TODO: Call `saveUserInfoToDefaults` here?
+    }
+
+    // MARK: - init
+    public init(
```
EdwardJTL, line 243: can we make one of these functions the default init and the rest as `convenience init`s? this way we have an idea of what the initial state should be in this object.

Shared/API/NeevaUserInfo.swift
evancrow: Not used outside this scope anymore, made it public

Shared/API/NeevaUserInfo.swift
evancrow: No, I'm going to remove that. I added it when I tested updating the login state based on the API fetch.

Shared/API/NeevaUserInfo.swift
evancrow: Good point! Will move it to where the cookie is actually set if it changed.

Shared/API/NeevaUserInfo.swift
evancrow: I wouldn't. There is only one public init for `SwiftUI Preview` methods, the rest are for creating specific states, two of which are for testing. None share a common `init`, so using `convenience init` isn't helpful.

EdwardJTL: okay. can you make them this way
```
// come comments
private init() { ... //the set up for test }
static func createForTesting() -> Self { ... }
```
just that it's obvious to the reader why we have different inits, even from reading the caller signature

evancrow: I'll add comments. No need to add a public method, because the testing methods are already surfaced through a public static instance.

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
```diff
@@ -125,7 +126,7 @@ class ZeroQueryModel: ObservableObject {
         }
 
         // impressions 4-10; show premium promo
-        if impressionCount < 10 {
+        if impressionCount < 10, canShowPremiumPromo {
```
EdwardJTL, line 129: `entitledToPremiumFeatures` could not be `nil` before. could you explain what the expected output is when it is `nil`? I think it will help readability if we can structure this in a way like
```
guard /*entitlement not nil*/ else { return }
guard /*not already entitled*/ else { return }
// set based on other states
```

Shared/API/NeevaUserInfo.swift
```diff
@@ -220,12 +194,36 @@ public class NeevaUserInfo: ObservableObject {
...
+        isVerified = userInfo.isVerified
+
+        NeevaFeatureFlags.update(featureFlags: userInfo.featureFlags)
+        UserFlagStore.shared.onUpdateUserFlags(userInfo.userFlags)
+    }
+
+    public func clearUserInfo() {
+        reachability.stopNotifier()
+
+        saveUserInfo(
```
EdwardJTL, line 213: nit: you can simplify this by either
1) allow the function to take in a `nil` value, which will instruct it to clear the defaults.
2) create a static let `UserInfo` like `nilUser` or `noUser` that has these specified.
This just helps reader understand what `clearUserInfo` intends to do. Looking at `saveUserInfo`, option 2 might be easier

EdwardJTL: or just add a comment `Set defaults to the default values of newly initialized fields` :)

Client/Frontend/ZeroQuery/ZeroQueryModel.swift
evancrow: The expected output when it is `nil` is the same as it was before, just with the premium promos excluded. I'll clean up that function to improve readability.

EdwardJTL (approved):
thanks! let's put this on testflight

PR #4627, opened 2023-02-13
Add deactivate and reload button to error page

chung-lai:
Sometimes content blocker may block content that user wants to open. Added a "Deactivate and Reload" button to the error page so it will add the url host to the content blocker allow list to disable Neeva Shield

screenshot

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


Client/Frontend/Strings.swift
```diff
@@ -56,6 +56,9 @@ extension Strings {
     public static let ErrorPagesGoBackButton = MZLocalizedString(
         "ErrorPages.GoBack.Button", value: "Go Back",
         comment: "Label for button to go back from the error page")
+    public static let ErrorPagesDeactivateAndReloadButton = MZLocalizedString(
+        "ErrorPages.DeactivateAndReload.Button", value: "Deactivate and Reload",
+        comment: "Label for button to deactivate content blocker and reload from the error page")
```
EdwardJTL, line 61: nit: `from the error page` is ambiguous. I'd suggest
"Label for a button on the error page, which deactivates content blockers and reloads the site"

Client/Frontend/Browser/Tabs/Tab.swift
```diff
@@ -876,6 +876,48 @@ extension Tab: ContentBlockerTab {
     }
 }
 
+class TabHelper {
```
EdwardJTL, line 879: I feel like this is better suited as a `TrackingPreventionConfig` util, or a `Tab` class extension.

chung-lai: was thinking about the same thing. Only reason I didn't add to `Tab` because we still want to add host to allow list when tab is not available. But the more i think about it, it doesn't make sense if tab does not exist. Will add it to `Tab` then :)

PR #4629, opened 2023-02-14
Fix Close All Tabs Also Closing Pinned+Archived Tabs

evancrow:
Fixes a bug where using the Close All Tabs option would also close Pinned Tabs and Archived Tabs.

### How was this tested?
- [x] I added automated tests

## Issues addressed
Clickup: 🔒 https://app.clickup.com/t/866966wk1


EdwardJTL:
is this the button from the overflow menu in the tab switcher?

evancrow:
That one, and the one via the Done button.

EdwardJTL:
thanks for doing this. I'm good with the code change. I ran it and noticed two things that I think we should get cleared with design
1. When there are only pinned tabs. with the option menu on the `Done` button, should we:
              1. dont allow force-touch option menu (how it behaves on this PR)
              2. allow the option menu, but show a disabled button
2. When there are only pinned tabs, via the overflow menu, the option to close all tabs is still there. When you press it, it shows "close 0 tabs". I find it a little confusing. I can think of these options, but we should just get design's input on what's suitable:
              1.  Show a disabled button in the overflow menu
              2. Allow this button to close pinned tabs as well. (so it wont show 0)
See screenshot here for point 2
IMG_0272

evancrow:
Adding to open hours!

evancrow:
Made the decision to suppress the menus from appearing.

EdwardJTL (approved):

PR #4641, opened 2023-03-08
Make Sure All UserInfo Cleared on SignOut

evancrow:
Fixes https://github.com/neevaco/ios-llm/pull/386/#issuecomment-1459059920

### How was this tested?
- [x] I tested this manually


EdwardJTL (approved):
TY!

PR #4646, opened 2023-03-17
Register push tokens for preview users

chung-lai:
* Register push tokens for preview users upon app upgrade from older version of apps
* Delete and re-register when user signs out


Client/Frontend/Notifications/NotificationPermissionHelper.swift
```diff
@@ -166,12 +178,28 @@ class NotificationPermissionHelper {
         UIApplication.shared.unregisterForRemoteNotifications()
     }
 
-    func registerDeviceTokenWithServer(deviceToken: String) {
-        guard NeevaUserInfo.shared.isUserLoggedIn else {
+    func registerDeviceTokenWithServer(deviceToken: String, fetchPreviewCookieIfNecessary: Bool) {
+        if !NeevaUserInfo.shared.isUserLoggedIn && (try? NeevaConstants.getPreviewCookie()) == nil {
+            if fetchPreviewCookieIfNecessary {
+                self.cookieFetcher.fetch(receiveOn: .main, pid: "neeva_app_ios") { [self] result in
```
evancrow, line 184: Should this key be a static instance somewhere?

Client/Frontend/Overlay/Content/Cheatsheet/CheatsheetService.swift
```diff
@@ -188,7 +188,7 @@ final class CheatsheetServiceProvider: CheatsheetDataService {
 
         // Fetch preview cookie before making the request
         if forceFetchCookie || (try? NeevaConstants.getPreviewCookie()) == nil {
-            self.cookieFetcher.fetch(receiveOn: .main) { [self] result in
+            self.cookieFetcher.fetch(receiveOn: .main, pid: "neeva_scope") { [self] result in
```
evancrow, line 191: Ditto

Client/Frontend/Notifications/NotificationPermissionHelper.swift
```diff
@@ -28,6 +28,18 @@ enum NotificationAuthorizationCallSite: String {
 class NotificationPermissionHelper {
     static let shared = NotificationPermissionHelper()
 
+    private lazy var cookieFetcher = PreviewCookieFetcher()
+
+    private lazy var previewClient: GraphQLAPI = {
+        let configuration = GraphQLAPI.Configuration(
```
evancrow, line 34: Can you move this configuration into `Network.swift` (inside `GraphQLAPI.Configuration`)

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -1015,8 +1015,16 @@ extension BrowserViewController: WKNavigationDelegate {
                         .ImplicitDeleteCookie,
                         attributes: EnvironmentHelper.shared.getFirstRunAttributes()
                     )
+
                     NotificationPermissionHelper.shared.deleteDeviceTokenFromServer()
+
                     userInfo.logOutUser()
+
+                    if let token = Defaults[.notificationToken] {
```
evancrow, line 1023: This function has been duplicated in a few places. Can we create one caller to replace them all?

evancrow (commented):
Few nits!

Client/Frontend/Notifications/NotificationPermissionHelper.swift
chung-lai: This is more an on-demand client so we probably don't need to create it most of the time when the app runs unless shared vs anonymous configuration.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
chung-lai: Right, soon this won't be required as it should be taken care of by backend. And don't want to pull out the `logOutUser` call into the same helper function. Will leave a TODO comment for me to remove it. Will likely be next release.

PR #4653, opened 2023-03-27
Shortcut To Copy Current Link

evancrow:
Adds a keyboard shortcut on iPad to copy the current link (cmd+shift+c). Also cleans up the keyCommand code up a bit.

### How was this tested?
- [x] I tested this manually


Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
```diff
@@ -164,59 +173,109 @@ extension BrowserViewController {
                 action: #selector(goForwardKeyCommand)),
         ]
         let tabNavigation = [
+            // Reload.
```
EdwardJTL, line 176: would it be helpful to make these `// MARK: - ` comments instead?

EdwardJTL (approved):
lgtm! ty for adding this

PR #4661, opened 2023-04-04
Improve How Refresh in Spaces Displays

evancrow:
Fixes:

- Spaces being pushed down because of the constant setting of the `RefreshControl` on the ScrollView
- Improved experience between switching from logged in & out & vice versa

Created a native SwiftUI `ProgressView` when we needed to manually show a refresh instead of trying to introspect the `ScrollView` to access the UIKit one.

### How was this tested?
- [x] I tested this manually


jedireza (approved):
I didn't test drive this. But we still have time to revert if there are any issues before our GA release.

PR #4662, opened 2023-04-05
Minor ZeroQuery Fixes

evancrow:
Some minor ZeroQuery fixes:

1. Fix a bug where the location bar would be hidden when swiping back to ZeroQuery from a tab
2. Fix the inability to scroll ZeroQuery past the keyboard

### How was this tested?
- [x] I tested this manually


jedireza:
Do you have steps to reproduce the bugs? I'd like to test drive and compare with `main`.

evancrow:
Bug 1) 
- Start a search, tap one of the page suggestions from FastTap
- Scroll the page
- Swipe back to ZeroQuery
- (Before) Location bar would be hidden (Now shows)

Bug 2)
- Open ZeroQuery
- (Before) Can't scroll through suggested sites/queries past the keyboard (Now you can)

jedireza:
Thanks! I confirmed the keyboard bug. Here is a video of trying to reproduce the search bar bug. Did I do it wrong?

https://user-images.githubusercontent.com/979929/230193386-04e9c3a4-915f-4005-9ee3-dd62f9cc5cf2.MP4

evancrow:
Kinda. After you perform the search, scroll down the search results page you navigated to, not the search suggestions.

jedireza:
Ah okay I see it. It took me a couple tries.

jedireza (approved):
Nice. I just confirmed the fixes. 🎉

PR #4667, opened 2023-04-10
Fix ArchivedTabRow Date

evancrow:
Been calculating the date used for sorting and creating headers incorrectly. This fixes the math used when calculating `numberOfDays` from today. Needed to inverse and fix the rounding; not sure how this worked before.

### How was this tested?
- [x] I tested this manually


evancrow:
Fixed:


Before:
_(Date doesn't exist)_

chung-lai (approved):
LGTM!

PR #4680, opened 2023-05-01
add functionality to export history

chung-lai:
add setting option export to a csv file

https://user-images.githubusercontent.com/87332578/235533607-eea38577-ce1f-451c-b1c9-cc364ed4bf6f.mov


Client/Frontend/Settings/SettingsView.swift
```diff
@@ -143,6 +153,52 @@ class SettingsViewController: UIHostingController<AnyView> {
                     bvc.presentSignInOrUpFlow()
                 }
             }
+            .environment(\.settingsExportBrowsingHistory) {
+                self.dismiss(animated: true) {
+                    let bvc = SceneDelegate.getBVC(for: nil)
+                    let profile = bvc.tabManager.profile
+                    profile.history.getAllSitesByLastVisit() >>== { sites in
+                        let formatter = DateFormatter()
```
evancrow, line 161: Can we make this `DateFormatter` static somewhere?

evancrow (approved):
Look alright to me. Do we need to do something similar for Archived Tabs?

chung-lai:
> Look alright to me. Do we need to do something similar for Archived Tabs?

History should be super set of all the archived tabs so i would consider archived tabs lower priority for export.

evancrow (approved):

